PR #25, opened 2023-08-31
Remove stripping of null values from Selection Set models

AnthonyMDev:
This PR removes the stripping of null values from the selection set models. 

This fixes apollographql/apollo-ios#3092. Because data returned from a network request was having null values stripped, when the returned models were then manually written to the cache, the null values were not written. This results in a `missingValue` error when you try to read those objects from the cache later.

This situation could occur if you fetched from the network with a cache policy of `.fetchIgnoringCacheCompletely` and then wrote some of the data to the cache manually. It also occurs in the common case of a fetched object be appended to another list of objects in a cache mutation.

## Null Value Representation

This bug fix revealed another related bug. We previously used `NSNull` to represent null values in a `SelectionSet`, then in order to work for Swift 5.3 and below, we needed to change this. We changed it to `Optional<AnyHashable>.none`, which actually was just `nil`. These `nil` values were being stripped by the executor as well. 

In order to handle null values in the selection set appropriately, we need to store them as a concrete value. This is now represented as `AnyHashable(Optional<AnyHashable>.none)`, which wraps a `nil` value inside of an `AnyHashable` struct. This is a concrete value instead of just `nil`, which provides the desired behavior.

To make the code clearer and more expressive, I've created a `DataDict.NullValue` static property that contains this value.

### DIFF
--- a/Tests/ApolloInternalTestHelpers/SelectionSet+TestHelpers.swift
+++ b/Tests/ApolloInternalTestHelpers/SelectionSet+TestHelpers.swift
@@ -1,0 +1,15 @@
+@testable import ApolloAPI
+@testable import Apollo
+
+public extension SelectionSet {
+
+  var _rawData: [String: AnyHashable] { self.__data._data }
+
+  func hasNullValue(forKey key: String) -> Bool {
+    guard let value = self.__data._data[key] else {
+      return false
+    }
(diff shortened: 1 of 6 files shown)


BobaFetters (approved):
This LGTM, nice work!

PR #28, opened 2023-08-31
feature: Require `label` argument on `@defer` directive

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3187

Adds validation to the frontend JS to ensure that all `@defer` directives specify a `label` argument.


BobaFetters (approved):
LGTM!

PR #31, opened 2023-09-11
feature: Generated models support the `@defer` directive

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3231

This is the new work to update the code generation engine to support @defer in generated models, which includes:
* JS frontend
* IR (lays the groundwork for the next phase of modifying the selection set templates)
* Operation definition templates (for multipart request header)


calvincestari:
Oops, this is meant to be a draft; still adding commits, tests to come next.

apollo-ios/Sources/ApolloAPI/Selection.swift
```diff
@@ -2,9 +2,11 @@ public enum Selection {
   /// A single field selection.
   case field(Field)
   /// A fragment spread of a named fragment definition.
-  case fragment(any Fragment.Type, deferred: Condition? = nil)
+  case fragment(any Fragment.Type)
   /// An inline fragment with a child selection set nested in a parent selection set.
-  case inlineFragment(any InlineFragment.Type, deferred: Condition? = nil)
+  case inlineFragment(any InlineFragment.Type)
+
+  case deferred(if: Condition? = nil, any Deferrable.Type, label: String)
```
AnthonyMDev, line 9: Can we add inline documentation here?

AnthonyMDev: Should `label` be optional since we don't use it for named fragments always now?

calvincestari: > Should label be optional since we don't use it for named fragments always now?

Yes with my merged PR from yesterday this will need to become optional - good catch.

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -1487,787 +1487,753 @@ class SelectionSetTemplateTests: XCTestCase {
...
 
     // when
     try buildSubjectAndOperation()
-    let allAnimals = try XCTUnwrap(
-      operation[field: "query"]?[field: "allAnimals"] as? IR.EntityField
+    let allAnimals_asDog = try XCTUnwrap(
+      operation[field: "query"]?[field: "allAnimals"]?[as: "Dog"]
+    )
+    let allAnimals_asDog_asCat = try XCTUnwrap(
+      allAnimals_asDog[as: "One"]?[field: "Friend"]?[as: "Cat"]
```
AnthonyMDev, line 1812: 1. The `as` subscript is not going to work here, because it will be looking for an actual type named "One" to match against the runtime type of the object. We'll need a new subscript for deferred fragments.

2. friend should be lowercase
```suggestion
      allAnimals_asDog[as: "One"]?[field: "friend"]?[as: "Cat"]
```

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -4751,235 +6121,271 @@ class SelectionSetTemplateTests: XCTestCase {
...
     """
 
     document = """
-    query TestOperation($a: Boolean!) {
+    query TestOperation {
       allAnimals {
-        ...FragmentA @include(if: $a)
+        __typename
+        id
+        ...DogFragment @defer(if: true)
```
AnthonyMDev, line 6383:
```suggestion
        ...DogFragment @defer(if: false)
```

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -5988,6 +7394,132 @@ class SelectionSetTemplateTests: XCTestCase {
...
+    let expected = """
+      public struct Root: TestSchema.InlineFragment, ApolloAPI.Deferrable {
+        public let __data: DataDict
+        public init(_dataDict: DataDict) { __data = _dataDict }
+
+        public typealias RootEntityType = TestOperation.Data.AllAnimal
+        public static var __parentType: ApolloAPI.ParentType { TestSchema.Objects.Dog }
+
+        public var species: String { __data["species"] }
+      }
```
AnthonyMDev, line 7441: the `id` field should also be merged in here.

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -5988,6 +7394,132 @@ class SelectionSetTemplateTests: XCTestCase {
...
+        ... on Dog @defer(label: "one") {
+          species
+        }
+        ... on Dog @defer(label: "two") {
+          genus
+        }
+      }
+    }
+    """
+
```
AnthonyMDev, line 7491: These should also both have the `id` field

AnthonyMDev: They also should both NOT have the merged field from the other fragment (One should have species and id, not genus; Two should have genus and id, not species)

AnthonyMDev (approved):

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
calvincestari: > The as subscript is not going to work here, because it will be looking for an actual type named "One" to match against the runtime type of the object. We'll need a new subscript for deferred fragments.

Yes - this will get done as part of making the tests pass.

apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/__tests__/referencedFragmentsTests.ts
```diff
@@ -25,8 +25,7 @@ describe("operation with referencedFragments", () => {
       ...FragmentA
       ...FragmentC
     }
-    `, "OperationA", { line: 1, column: 1 }),
-    false
+    `, "OperationA", { line: 1, column: 1 })
```
calvincestari, line 28: This is due to the move from graphql-js 16 to 17.

AnthonyMDev (approved):

PR #33, opened 2023-09-13
refactor: Don't require `label` argument for deferred named fragments

calvincestari:
The label value of the `@defer` directive is not used when we have a named fragment so we can drop the requirement for a `label` argument to only inline fragments.

### DIFF
--- a/apollo-ios-codegen/Sources/ApolloCodegenLib/Frontend/JavaScript/src/__tests__/deferDirectiveTests.ts
+++ b/apollo-ios-codegen/Sources/ApolloCodegenLib/Frontend/JavaScript/src/__tests__/deferDirectiveTests.ts
@@ -208,7 +208,7 @@ describe("given schema", () => {
 
       expect(validationErrors.length).toEqual(1)
       expect(validationErrors[0].message).toEqual(
-        "Apollo requires all @defer directives to use the 'label' argument. Please add a 'label' argument to the @defer directive on this inline fragment."
+        "Apollo does not support deferred inline fragments without a 'label' argument. Please add a 'label' argument to the @defer directive on this inline fragment."
       )
     })
   })
@@ -261,7 +261,7 @@ describe("given schema", () => {
 
       expect(validationErrors.length).toEqual(1)
       expect(validationErrors[0].message).toEqual(
-        "Apollo requires all @defer directives to use the 'label' argument. Please add a 'label' argument to the @defer directive on this inline fragment."
+        "Apollo does not support deferred inline fragments without a 'label' argument. Please add a 'label' argument to the @defer directive on this inline fragment."
       )
     })
   })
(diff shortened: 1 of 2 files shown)


apollo-ios-codegen/Sources/ApolloCodegenLib/Frontend/JavaScript/src/validationRules.ts
```diff
@@ -89,39 +88,23 @@ export function DeferredInlineFragmentNoTypeCondition(context: ValidationContext
...
-        )
-      )
+      if (node.directives) {
+        for (const directive of node.directives) {
+          if (directive.name.value == GraphQLDeferDirective.name && !(directive.arguments?.find((element) =>
+            element.name.value == 'label'
+          ))) {
+            context.reportError(
+              new GraphQLError(
+                "Apollo requires all @defer directives to use the 'label' argument. Please add a 'label' argument to the @defer directive on this inline fragment.",
```
AnthonyMDev, line 101: Should we also change the error message to reflect this new logic?

apollo-ios-codegen/Sources/ApolloCodegenLib/Frontend/JavaScript/src/__tests__/deferDirectiveTests.ts
```diff
@@ -230,13 +230,10 @@ describe("given schema", () => {
...
 
-    it("should fail validation", () => {
+    it("should pass validation", () => {
       const validationErrors: readonly GraphQLError[] = validateDocument(schema, document, emptyValidationOptions)
 
-      expect(validationErrors.length).toEqual(1)
-      expect(validationErrors[0].message).toEqual(
-        "Apollo requires all @defer directives to use the 'label' argument. Please add a 'label' argument to the @defer directive on this fragment spread."
-      )
+      expect(validationErrors.length).toEqual(0)
```
AnthonyMDev, line 236: Nit:
```suggestion
      expect(validationErrors).to(beEmpty())
```

AnthonyMDev (approved):

apollo-ios-codegen/Sources/ApolloCodegenLib/Frontend/JavaScript/src/validationRules.ts
calvincestari: I've updated it in https://github.com/apollographql/apollo-ios-dev/pull/33/commits/9f0dddd12a84c5033892a64a03bc62bf6618dce9.

apollo-ios-codegen/Sources/ApolloCodegenLib/Frontend/JavaScript/src/__tests__/deferDirectiveTests.ts
calvincestari: Turns out this isn't a valid matcher in Jest; had to use `.toHaveLength` instead.

PR #34, opened 2023-09-14
Include local SPM packages in project, instead of as additional files

AnthonyMDev:
This cleans this up a bit by including these directly in the Project. 

I was making some other changes to the structure of the codgen packages that was causing build errors, and doing this resolved them.

### DIFF
--- a/Project.swift
+++ b/Project.swift
@@ -7,7 +7,9 @@ let project = Project(
     name: "ApolloDev",
     organizationName: "apollographql",
     packages: [
-        .package(url: "https://github.com/Quick/Nimble.git", from: "10.0.0")
+        .package(url: "https://github.com/Quick/Nimble.git", from: "10.0.0"),
+        .package(path: "apollo-ios"),
+        .package(path: "apollo-ios-codegen"),
     ],
     settings: Settings.settings(configurations: [
         .debug(name: .debug, xcconfig: "Configuration/Apollo/Apollo-Project-Debug.xcconfig"),
--- a/Workspace.swift
+++ b/Workspace.swift
@@ -4,11 +4,7 @@ let workspace = Workspace(
     name: "ApolloDev",
     projects: [
         "./"
-    ],
-    additionalFiles: [
-        .folderReference(path: "apollo-ios"),
-        .folderReference(path: "apollo-ios-codegen")
-    ],
+    ],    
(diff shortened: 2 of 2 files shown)


BobaFetters:
Just making sure this is what we want, originally you mentioned not wanting the local packages included like this because of potential development weirdness which is why we went down the path we did, but if you aren't concerned about that anymore then this is fine with me.

If we are going this route I would see if we can add the `enableAutomaticXcodeSchemes` to the `Project.swift` and if so we can probably get rid of the workspace entirely, which if we do would probably have a few small docs updates to the READMEs etc that reference the workspace.

AnthonyMDev:
> you mentioned not wanting the local packages included like this because of potential development weirdness which is why we went down the path we did

Ahhhh, I see! I didn't realize that was the reason you did this. As I've been working with it, I think this is probably more appropriate though. Its the intended usage of both Tuist and Xcode, and a lot of the SPM weird bugs have been getting fixed in newer versions of Xcode.

Regardless, the other way of doing this was causing me a bunch of linker issues as I started to break out the targets in #35, so I think we should move forward with this one.

I'm definitely open to looking into `enableAutomaticXcodeSchemes` and the other changes you recommended. Can those be done in another PR after this is merged in?

BobaFetters:
I think before we merge this in we should verify whether `enableAutomaticXcodeSchemes` is available on a `Project` like it is on `Workspace` to ensure we will be able to do it, otherwise we need to make sure we have a plan for getting those schemes created, i guess worst case maybe everything is in the project and then we keep the workspace containing only the project but then it can still generate the schemes

Otherwise I am good with moving forward with this setup overall.

AnthonyMDev:
According to the [docs here](https://tuist.github.io/tuist/main/documentation/projectdescription/workspace/)

> By default, tuist generate generates an Xcode workspace that has the same name as the current project. It includes the project and all its dependencies.
>
> Tuist allows customizing this behaviour by defining a workspace manifest within a Workspace.swift file.

So it sounds like there is no value in getting rid of the workspace manifest since it always generates a workspace anyways.

BobaFetters:
Cool I am good with this then and we can just leave that option in the workspace and keeps the docs as is and work out of the workspace.

BobaFetters (approved):

PR #35, opened 2023-09-14
Break out GraphQLCompiler and TemplateString

AnthonyMDev:
The “Frontend” is now contained in it’s own target named `GraphQLCompiler` and TemplateString is also a new target. These targets are part of the `ApolloCodgenLib` package. But they are not standalone products and cannot be consumed externally.

This also changes the `auto_rollup` script used for compiling the JS Frontend. It changes the paths of the distribution to be inside of the “JavaScript” folder so it can more easily be excluded from the package files.


Tuist/ProjectDescriptionHelpers/Targets/Target+ApolloTests.swift
```diff
@@ -18,7 +18,7 @@ extension Target {
             dependencies: [
                 .target(name: ApolloTarget.apolloInternalTestHelpers.name),
                 .target(name: ApolloTarget.starWarsAPI.name),
-                .target(name: ApolloTarget.uploadAPI.name),
+                .target(name: ApolloTarget.uploadAPI.name),
```
BobaFetters, line 21: minor nit: just looks like some extra whitespace added

BobaFetters (approved):
Looks good! Love the refactoring and clean up here!

PR #37, opened 2023-09-18
Breakout Codegen IR to separate module

AnthonyMDev:
Breaking the IR into it's own module allows us to remove the namespacing via an extension and improves organization and separation of concerns.

Operation identifier computation is now handled via a new `OperationIdentifierFactory`.
In my next PR, I'll be working on the ability to do custom operation identifier computation.


apollo-ios-codegen/Package.swift
```diff
@@ -6,10 +6,7 @@ import PackageDescription
 let package = Package(
   name: "ApolloCodegen",
   platforms: [
-    .iOS(.v12),
-    .macOS(.v10_15),
-    .tvOS(.v12),
-    .watchOS(.v5)
+    .macOS(.v10_15)
```
calvincestari, line 9: Nice to be able to clean this up finally. That's a big win!

calvincestari (approved):
LGTM!

PR #38, opened 2023-09-19
fix: Subtree bug

BobaFetters:
-Adding the customized git subtree script to use for CI that included the subtree bugfix

### DIFF
--- a/git-sh-setup
+++ b/git-sh-setup
@@ -1,0 +1,357 @@
+# This shell scriplet is meant to be included by other shell scripts
+# to set up some variables pointing at the normal git directories and
+# a few helper shell functions.
+
+# Having this variable in your environment would break scripts because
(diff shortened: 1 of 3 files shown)


git-sh-setup
```diff
@@ -0,0 +1,357 @@
+# This shell scriplet is meant to be included by other shell scripts
```
AnthonyMDev, line 1: Is this a script you got from somewhere else? We just need to make sure we are checking the LICENSE and it's okay to reuse this.

AnthonyMDev (approved):
Amazing! :) Thank you Zach!!

git-sh-setup
BobaFetters: its part of git and required for the subtree script to run correctly

PR #39, opened 2023-09-19
Add referencedFragments to CompilationResult

AnthonyMDev:
Allow us to track the referenced fragments from the js front end. This means we have access to this information prior to running the `IRBuilder.build` functions.

We can now generate operation identifiers and manifests with this information without having to build the entire IR first.

This could also be used in the future to track dependencies between fragments to optimize the order of building (and releasing from memory) the fragments in the codegen.


Tests/ApolloCodegenTests/AnimalKingdomAPI/AnimalKingdomIRCreationTests.swift
```diff
@@ -1038,4 +1038,54 @@ final class AnimalKingdomIRCreationTests: XCTestCase {
...
+    // given
+    let operation = compilationResult.operations.first { $0.name == "AllAnimalsQuery" }
+
+    // when
+    let expected: [CompilationResult.FragmentDefinition] = [
+      "HeightInMeters",
+      "WarmBloodedDetails",
+      "PetDetails"
+    ].map { expectedName in
+      compilationResult.fragments.first(where:{ $0.name == expectedName })!
```
calvincestari, line 1054: This check looks like it's enforcing that the expected fragment name must be found but should the test also be validating that there are only the three referenced fragments?

apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/dist/ApolloCodegenFrontend.bundle.js
```diff
@@ -1 +1 @@
...
...]?e.reportError(new GraphQLError(`There can be only one directive named "@${i}".`,[t[i],r.name])):t[i]=r.name,!1;e.reportError(new GraphQLError(`Directive "@${i}" already exists in the schema. It cannot be redefined.`,r.name))}}},dn,pn,Un,function(e){const t=e.getSchema(),n=Object.create(null);for(const t of e.getDocument().definitions)cn(t)&&(n[t.name.value]=t);return{ScalarTypeExtension:i,ObjectTypeExtension:i,InterfaceTypeExtension:i,UnionTypeExtension:i,EnumTypeExtension:i,InputObjectTypeExtension:i};function i(i){const o=i.name.value,s=n[o],a=null==t?void 0:t.getType(o);let c;if(s?c=Sn[s.kind]:a&&(c=function(e){if(Ae(e))return m.SCALAR_TYPE_EXTENSION;if(De(e))return m.OBJECT_TYPE_EXTENSION;if(we(e))return m.INTERFACE_TYPE_EXTENSION;if(ke(e))return m.UNION_TYPE_EXTENSION;if(xe(e))return m.ENUM_TYPE_EXTENSION;if(Fe(e))return m.INPUT_OBJECT_TYPE_EXTENSION;r(!1,"Unexpected type: "+Y(e))}(a)),c){if(c!==i.kind){const t=function(e){switch(e){case m.SCALAR_TYPE_EXTENSION:return"scalar";case m.OBJECT_TYPE_EXTENSION:return"object";case m.INTERFACE_TYPE_EXTENSION:return"interface";case m.UNION_TYPE_EXTENSION:return"union";case m.ENUM_TYPE_EXTENSION:return"enum";case m.INPUT_OBJECT_TYPE_E
```
calvincestari, line 1: We should rather be removing this file instead of updating it. The content gets copied into `ApolloCodegenFrontendBundle.swift` anyways.

calvincestari (approved):
Looks good.

Tests/ApolloCodegenTests/AnimalKingdomAPI/AnimalKingdomIRCreationTests.swift
```diff
@@ -1038,4 +1038,54 @@ final class AnimalKingdomIRCreationTests: XCTestCase {
...
+      "HeightInMeters"
+    ].map { expectedName in
+      compilationResult.fragments.first(where:{ $0.name == expectedName })!
+    }
+
+    // then
+    expect(fragment?.referencedFragments).to(equal(expected))
+  }
+
+  func test__referencedFragments__PetDetails_isCorrect() throws {
```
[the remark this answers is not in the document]
AnthonyMDev: These are kind of odd tests because they test the example animal kingdom API. When that API, these tests need to be changed to accurately represent the correct information there. So changing the name also just doesn't add value. The only reason "shouldNotHaveReferencedFragments" is true is because the Pet Details fragment doesn't have any right now. It's not truly a test that "a fragment with no referenced fragments should have an empty referenced fragments array." The unit tests for this PR are the tests in the JS code. These are just some added integration tests.

apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/dist/ApolloCodegenFrontend.bundle.js
AnthonyMDev: Done

PR #40, opened 2023-09-21
Fixing hashable implementation of GraphQLResponse

BobaFetters:
-Fixing the implementation of GraphQLResponse that was causing xcframework build failures with Xcode 15 and Swift 5.9

Closes apollographql/apollo-ios#3154


calvincestari (approved):
Good find!

PR #41, opened 2023-09-22
Adding ci to auto close PRs

BobaFetters:
-Adding ci script to auto close and comment on PRs in the apollo-ios and apollo-ios-codegen repos

When this PR is merged we will need to manually go disable these workflows in their respective repos until we rollout the new project structure publicly and then we can re-enable them.

### DIFF
--- a/apollo-ios-codegen/.github/workflows/pr-close.yml
+++ b/apollo-ios-codegen/.github/workflows/pr-close.yml
@@ -1,0 +1,14 @@
+name: Close Pull Request
+
+on:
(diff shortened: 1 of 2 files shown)


BobaFetters:
Need to reassess this some because I'm not sure the action being used here will work correctly with forked repo PRs.

BobaFetters:
Reading into it a little more it looks like this should also work for forked repo PRs

calvincestari (approved):
Nice, these are going to helpful.

PR #46, opened 2023-09-26
Update test infra to use Nimble 12.3

AnthonyMDev:


calvincestari (approved):
Is there a particular feature that `12.3.0` unlocks for us?

AnthonyMDev:
Yes, it supports async/await for use with  test functions that test async code.

PR #61, opened 2023-10-05
Handle coercion of AnyHashable 

AnthonyMDev:
In iOS versions 14.4 and lower, `AnyHashable` coercion does not work. On these platforms, we need to do some additional unwrapping and casting of the values to avoid crashes and other run time bugs.

In addition to doing some more explicit handling of the null values, this PR changes the check for the coercion from 
`#if swift (>=5.4)` to `if #available(iOS 14.5, *)`. It turns out that this is not particularly related to the swift version, but to the iOS version being run.

This PR has been tested locally on iOS 14.4, 14.5, and 17.0.

### DIFF
--- a/apollo-ios/Sources/ApolloAPI/DataDict.swift
+++ b/apollo-ios/Sources/ApolloAPI/DataDict.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 /// A structure that wraps the underlying data for a ``SelectionSet``.
 public struct DataDict: Hashable {
   @usableFromInline var _storage: _Storage
@@ -51,11 +53,16 @@ public struct DataDict: Hashable {
 
   @inlinable public subscript<T: AnyScalarType & Hashable>(_ key: String) -> T {
     get {
-#if swift(>=5.4)
-        _data[key] as! T
-#else
-        _data[key]?.base as! T
-#endif
+      if DataDict._AnyHashableCanBeCoerced {
+        return _data[key] as! T
+      } else {        
+        let value = _data[key]
+        if value == DataDict._NullValue {
+          return (Optional<T>.none as Any) as! T
+        } else {
+          return (value?.base as? T) ?? (value._asAnyHashable as! T)
+        }
+      }
     }
     set {
       _data[key] = newValue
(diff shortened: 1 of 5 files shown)


apollo-ios/Sources/ApolloAPI/DataDict.swift
```diff
@@ -179,11 +216,11 @@ extension Array: SelectionSetEntityValue where Element: SelectionSetEntityValue
       fatalError("\(Self.self) expected list of data for entity.")
     }
     self = data.map {
-#if swift(>=5.4)
+      if DataDict._AnyHashableCanBeCoerced {
         Element.init(_fieldData:$0)
```
BobaFetters, line 220: Looks like these lines need return statements now since its no longer technically a single line of code because of the `#if` check

AnthonyMDev: Yeah, thanks. just fixed that. Oddly it was still compiling and running fine on my machine with out them. I have no idea how.

BobaFetters (approved):
LGTM! Nice work!

matijakregarGH:
@AnthonyMDev 
Thanks for the quick response, however an(other) issue still exists (Apollo v 1.6.0, iOS v 14.4.1).
I'm now getting a fatal error in `DataDict` file (line 182) in the `RootSelectionSet` extension guard statement.
`Fatal error: ... expected DataDict for entity, got Optional<AnyHashable>.`

Called from `extension Optional: SelectionSetEntityValue where Wrapped: SelectionSetEntityValue` (DataDict line 203)
Called from `@inlinable public subscript<T: SelectionSetEntityValue>(_ key: String) -> T` (DataDict line 78)

Please let me know how I can help to get this resolved ...

Best,
Matija

AnthonyMDev:
I had forgotten the `!` in front of one of the conditions for checking if we were on iOS 14.5+. Should be if NOT on iOS 14.5+. 
PR is up to fix this now. https://github.com/apollographql/apollo-ios-dev/pull/68

We'll get a patch out ASAP

PR #64, opened 2023-10-06
Moving docs publish ci job

BobaFetters:
Moving the docs publish ci job from the `apollo-ios` repo to the `apollo-ios-dev` repo

### DIFF
--- a/.github/workflows/docs-publish.yml
+++ b/.github/workflows/docs-publish.yml
@@ -4,7 +4,6 @@ on:
   push:
     branches:
       - main
-      - release/1.0
     paths:
       - docs/**
 


calvincestari (commented):
Do we need to remove there [branches from the `apollo-ios` repo too](https://github.com/apollographql/apollo-ios/blob/main/.github/workflows/docs-publish.yml#L5) then if we're moving publishing to this repo?

BobaFetters:
So I removed the `release/1.0` branch from the script when i moved it to this repo, but the main branch should be our target to publish docs from when changes are detected like we were in the other repo.

calvincestari (approved):

PR #70, opened 2023-10-10
Include tests into ApolloPaginationTests target

Iron-Ham:
Ideally this would be pointed at https://github.com/apollographql/apollo-ios-dev/pull/69, since this PR is based on that one. 

This PR:
- Includes the tests for the configuration introduced in https://github.com/apollographql/apollo-ios-dev/pull/69

Must be merged after: #69


Tests/ApolloPaginationTests/SubscribeTests.swift
```diff
@@ -0,0 +1,136 @@
...
+          "friendsConnection": friendsConnection
+        ]
+
+        let data: [String: AnyHashable] = [
+          "hero": hero
+        ]
+
+        return [
+          "data": data
+        ]
```
Iron-Ham, line 119: Would love to know if y'all have a better solution for this -- I noticed that intermingling `string` and `int` types leads to Swift being unhappy for literals like this, which had me declare chunks as a workaround -- any thoughts?

BobaFetters (approved):
I think this looks good, want to give @AnthonyMDev a chance to review as well since he is more familiar with this work overall than I am

Iron-Ham:
Closing in favor of #80, which re-writes all tests by necessity.

PR #74, opened 2023-10-11
[1/x] Refactor root field builder for deferred fragments

calvincestari:
This PR for https://github.com/apollographql/apollo-ios/issues/3141 refactors the root field builder to build deferred inline and nested fragments into the IR.
* Refactors the `IsDeferred` enum into a struct named `DeferCondition` that is an Optional when used; this allows us to use `nil` to signal "no defer condition" and `some` to signal "a defer condition" exists. It now also holds the label.
* Moved the defer condition from `ScopeDescriptor` to `ScopeCondition` which allows us to fully differentiate it in the scope path.
* The logic for building deferred inline/nested fragments into the IR is to wrap the deferred scope condition in the type case of it's parent, effectively nesting the selections of the deferred selection set.
* Merging of deferred direct selections is disabled.


Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -331,7 +331,7 @@ class SelectionSetTemplateTests: XCTestCase {
 
     let expected = """
       public static var __parentType: ApolloAPI.ParentType { TestSchema.Objects.Nested }
-    
+
```
calvincestari, line 334: I don't know why the diff for this file keeps looking so messed up. All I did here was to add `throw XCTSkip()` so that the PR can pass CI. The skips will be removed in the third PR which actually updates the selection set template.

apollo-ios-codegen/Sources/IR/IR+DeferCondition.swift
```diff
@@ -0,0 +1,25 @@
+import GraphQLCompiler
```
calvincestari, line 1: This was previously `IR+IsDeferred.swift` and was also an `enum`.

apollo-ios-codegen/Sources/IR/IR+RootFieldBuilder.swift
```diff
@@ -136,19 +136,14 @@ class RootFieldBuilder {
...
-        return true
-      }
-    }()
+    self.containsDeferredFragment = typeInfo.scope.scopePath.containsDeferredFragment
 
-    typeInfo.entity.selectionTree.mergeIn(
-      selections: target.readOnlyView,
-      with: typeInfo
-    )
+    if typeInfo.scope.scopePath.last.value.deferCondition == nil {
```
calvincestari, line 141: This is the check to disable all merging of deferred selections. It will be refactored in the work Anthony is doing.

apollo-ios-codegen/Sources/IR/IR+RootFieldBuilder.swift
```diff
@@ -353,6 +383,29 @@ class RootFieldBuilder {
     return InlineFragmentSpread(selectionSet: irSelectionSet)
   }
 
+  private func buildInlineFragmentSpread(
```
calvincestari, line 386: This is the new function that takes the deferred inline fragment and wraps it in the parent type case, effectively creating a nested type case fragment.

apollo-ios-codegen/Sources/IR/IR+SelectionSet.swift
```diff
@@ -160,3 +160,17 @@ public class SelectionSet: Hashable, CustomDebugStringConvertible {
   }
 
 }
+
+extension LinkedList where T == ScopeCondition {
+  var containsDeferredFragment: Bool {
```
calvincestari, line 165: This walks the linked list backwards because it would make sense that the tail of the scope descriptor being evaluated is the fragment that is deferred, which should provide an early exit.

BobaFetters (approved):
LGTM!

apollo-ios-codegen/Sources/IR/IR+RootFieldBuilder.swift
```diff
@@ -243,17 +216,69 @@ class RootFieldBuilder {
...
+      )
+      target.mergeIn(irTypeCase)
+    }
+  }
+
   private func scopeCondition(
     for conditionalSelectionSet: ConditionallyIncludable,
-    in parentTypePath: SelectionSet.TypeInfo
+    in parentTypePath: SelectionSet.TypeInfo,
+    deferCondition: CompilationResult.DeferCondition? = nil
```
calvincestari, line 269: This needs to be refactored because it's confusing why `deferCondition` is passed into the function but not used in the creation of the `ScopeCondition`; it's only used to determine the parent type. Maybe renaming the parameter name is enough?

calvincestari: Changed to a `bool` which implies it's more of a flag and not used to create the scope condition.

Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
```diff
@@ -4572,7 +4569,885 @@ class IRRootFieldBuilderTests: XCTestCase {
...
+
+    expect(allAnimals_AsDog).to(shallowlyMatch(
+      SelectionSetMatcher(
+        parentType: Object_Dog,
+        directSelections: [
+          .deferred(Fragment_DogFragment, label: "root"),
+        ],
+        mergedSelections: [
+          .field("id", type: .string()),
+          .field("species", type: .string()), // wrong
```
calvincestari, line 5224: @AnthonyMDev is there a quick way to disable merging of these fields like we did for direct selection merging? This will also likely be part of the work you're doing.

apollo-ios-codegen/Sources/GraphQLCompiler/CompilationResult.swift
```diff
@@ -408,16 +409,13 @@ public class CompilationResult: JavaScriptObject {
     }
   }
 
-  public enum IsDeferred: ExpressibleByBooleanLiteral {
-    case value(Bool)
-    case variable(String)
+  public struct DeferCondition {
```
AnthonyMDev, line 412: I think this is a lot cleaner now! Very nice.

AnthonyMDev (changes requested):

apollo-ios-codegen/Sources/IR/IR+SelectionSet.swift
```diff
@@ -160,3 +162,17 @@ public class SelectionSet: Hashable, CustomDebugStringConvertible {
...
+    var node: Node? = last
+    var deferCondition = node?.value.deferCondition
+
+    while node?.previous != nil && deferCondition == nil {
+      node = node?.previous
+      deferCondition = node?.value.deferCondition
+    }
+
+    return deferCondition != nil
```
AnthonyMDev, line 176: This is a small improvement in my opinion:

```suggestion
    do {
      guard node?.value.deferCondition == nil else {
        return true
      }

      node = node?.previous      
    } while node != nil

    return false
```

AnthonyMDev (changes requested):
This is looking really good! The implementation of it seems to work as intended and I'm happy with the naming changes and clean up.

I just see a lot of missing test cases still. It's crazy how many complex edge cases we have to account for with this stuff.

(I'm now realizing we haven't even gotten into tests `@defer` _with_ `@include/@skip`!)

AnthonyMDev (changes requested):
This is looking really good! The implementation of it seems to work as intended and I'm happy with the naming changes and clean up.

I just see a lot of missing test cases still. It's crazy how many complex edge cases we have to account for with this stuff.

(I'm now realizing we haven't even gotten into tests `@defer` _with_ `@include/@skip`!)

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -1488,787 +1488,771 @@ class SelectionSetTemplateTests: XCTestCase {
...
 
     // when
     try buildSubjectAndOperation()
-    let allAnimals = try XCTUnwrap(
-      operation[field: "query"]?[field: "allAnimals"] as? IR.EntityField
+    let allAnimals_asDog = try XCTUnwrap(
+      operation[field: "query"]?[field: "allAnimals"]?[as: "Dog"]
+    )
+    let allAnimals_asDog_asCat = try XCTUnwrap(
+      allAnimals_asDog[as: "One"]?[field: "friend"]?[as: "Cat"]
```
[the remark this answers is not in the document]
calvincestari: Ignore all the selection set template tests in this PR. They are there from a previous addition, and will be used in the subsequent PR which actually does the template work. But for now they're all being skipped too.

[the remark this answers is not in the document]
calvincestari: They're also not correct for what they should be testing, as you're finding with these comments. Rather the more valuable part of them is the operation definition - for now.

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -1488,787 +1488,771 @@ class SelectionSetTemplateTests: XCTestCase {
...
     document = """
-    query TestOperation($a: Boolean!) {
+    query TestOperation {
       allAnimals {
-        ... on Animal @include(if: $a) {
-          fieldA
-        }
+        __typename
+        id
+        ...DogFragment @defer(label: "root")
```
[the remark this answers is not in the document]
calvincestari: Will be handled in a subsequent PR that focuses on the selection set generation.

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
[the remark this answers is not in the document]
calvincestari: _Will be handled in a subsequent PR that focuses on the selection set generation._

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -4893,116 +5254,1205 @@ class SelectionSetTemplateTests: XCTestCase {
...
-
-    let actual = subject.render(inlineFragment: allAnimals_ifA)
-
+    
+    let actual = subject.render(inlineFragment: allAnimals_asDog)
+    
     // then
-    expect(actual).to(equalLineByLine(expected, atLine: 14, ignoringExtraLines: true))
+    expect(actual).to(equalLineByLine(expected, atLine: 15, ignoringExtraLines: true))
   }
```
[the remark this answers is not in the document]
calvincestari: Will be handled in a subsequent PR that focuses on the selection set generation.

apollo-ios-codegen/Sources/GraphQLCompiler/CompilationResult.swift
```diff
@@ -428,28 +426,43 @@ fileprivate protocol Deferrable {
...
+      return nil
     }
 
-    guard let argument = directive.arguments?.first(
+    guard
+      let labelArgument = directive.arguments?.first(
+        where: { $0.name == CompilationResult.Constants.ArgumentLabels.Label }),
+      case let .string(labelValue) = labelArgument.value
+    else {
+      preconditionFailure("Incorrect `label` argument. Either missing or value is not a String.")
```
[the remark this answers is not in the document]
calvincestari: Correct - validation is done in the JS layer for both [inline fragments](https://github.com/apollographql/apollo-ios-dev/blob/feature/defer/apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/__tests__/deferDirectiveTests.ts#L191) and [named fragments](https://github.com/apollographql/apollo-ios-dev/blob/feature/defer/apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/__tests__/deferDirectiveTests.ts#L216).

apollo-ios-codegen/Sources/GraphQLCompiler/CompilationResult.swift
```diff
@@ -428,28 +426,43 @@ fileprivate protocol Deferrable {
...
+        return nil
+      }
 
     case let .string(value), let .variable(value):
-      return .variable(value)
+      return .init(label: labelValue, variable: value)
 
     default:
-      preconditionFailure("Incompatible argument value. Expected Boolean or Variable, got \(argument.value).")
+      preconditionFailure("""
```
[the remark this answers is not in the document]
calvincestari: Same answer; JS has validated it for us already so if it gets here and they're missing it's an irrecoverable error.

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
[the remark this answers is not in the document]
calvincestari: > I can't remember if we came to the decision that we don't need the label argument for deferring named fragments? It looks like in the DeferCondition struct we have label as required. Was there a reason why we decided we still needed the label? Was it for execution or something?

I remember that conversation and yes I think we came to the conclusion that it may not be necessary because we already have something to name it in the named fragment. I would prefer to keep it around for now incase, as you mentioned, it's needed for execution. I created https://github.com/apollographql/apollo-ios/issues/3256 so we can make that change later if we find the label is not needed anywhere.

apollo-ios-codegen/Sources/IR/IR+DeferCondition.swift
```diff
@@ -0,0 +1,25 @@
...
+
+// TODO: Documentation for this to be completed in issue #3141
+public struct DeferCondition: Hashable, CustomDebugStringConvertible {
```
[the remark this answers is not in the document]
calvincestari: Yeah, it's a good callout; I'll remove the `IR` defined type since `GraphQLCompiler` is imported into many of the `IR` module files.

calvincestari:
@AnthonyMDev I believe this PR is ready for another review, it's now got:
* additional inline fragment tests on matching and different types cases
* additional named fragment tests on matching and different type cases
* fixes for merging behaviour across deferred and non-deferred named fragments when mixed
* removed `DeferCondition` from the IR and all code now uses the common one from `CompilationResult`
* additional tests for combining `@include` and `@skip` directives with `@defer` on inline and named fragments

What it does not have though is additional logic and tests for [merging deferred fragments and merging across deeply nested siblings](https://github.com/apollographql/apollo-ios-dev/pull/74#discussion_r1361230823); I've broken that requirement out into https://github.com/apollographql/apollo-ios/issues/3261 since it's convenience rather than a requirement for a preview release.

Thanks for the thorough review earlier, I think we've got much better code here as a result. 🎉

apollo-ios-codegen/Sources/GraphQLCompiler/CompilationResult.swift
```diff
@@ -164,7 +160,7 @@ public class CompilationResult: JavaScriptObject {
 
     lazy var directives: [Directive]? = self["directives"]
 
-    public lazy var isDeferred: IsDeferred = getIsDeferred()
+    public lazy var deferCondition: DeferCondition? = getDeferCondition()
```
AnthonyMDev, line 163: When I merge my changes for 1.7.0 into `main` this will need to change a little. For concurrency, we need to make all the JavaScript types immutable, so no `lazy` vars. I'll make the change to update this to calculate and store the value during initialization.

So this is fine for this PR.

apollo-ios-codegen/Sources/IR/IR+RootFieldBuilder.swift
```diff
@@ -170,90 +163,147 @@ class RootFieldBuilder {
...
     let inclusionResult = inclusionResult(for: conditionalSelectionSet.inclusionConditions)
     guard inclusionResult != .skipped else {
       return nil
     }
 
-    let type = parentTypePath.parentType == conditionalSelectionSet.parentType ?
-    nil : conditionalSelectionSet.parentType
+    let type = (
+      parentTypePath.parentType == conditionalSelectionSet.parentType
+      && !isDeferred
```
AnthonyMDev, line 303: Can we just add a comment here explaining _why_ we return `nil` when `isDeferred` is true? The logic is getting more and more complex and difficult to understand. These edge cases would be a lot easier to deal with in the future with inline explanations.

apollo-ios-codegen/Sources/IR/IR+RootFieldBuilder.swift
```diff
@@ -378,10 +464,12 @@ class RootFieldBuilder {
     let fragmentSpread = NamedFragmentSpread(
       fragment: fragment,
       typeInfo: typeInfo,
-      inclusionConditions: AnyOf(scopeCondition.conditions)
+      inclusionConditions: AnyOf(scope.conditions)
     )
 
-    entityStorage.mergeAllSelectionsIntoEntitySelectionTrees(from: fragmentSpread)
+    if fragmentSpread.typeInfo.deferCondition == nil {
```
AnthonyMDev, line 470: :+1:

AnthonyMDev (approved):
This is awesome! Great job @calvincestari! I know it was a difficult journey to get here. Excited to move forward!

Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
```diff
@@ -4572,7 +4569,2446 @@ class IRRootFieldBuilderTests: XCTestCase {
...
+    ))
+
+    expect(allAnimals_asDog_deferredAsRoot).to(shallowlyMatch(
+      SelectionSetMatcher(
+        parentType: Object_Dog,
+        directSelections: [
+          .field("species", type: .string()),
+        ],
+        mergedSelections: [
+          .field("id", type: .string()),
```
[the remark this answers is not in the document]
calvincestari: Correct, I think that will end up getting done in https://github.com/apollographql/apollo-ios/issues/3261.

apollo-ios-codegen/Sources/IR/IR+RootFieldBuilder.swift
calvincestari: Added in https://github.com/apollographql/apollo-ios-dev/pull/74/commits/ae4ce883f4e4819398e168d2008eb3ee1795dc66.

PR #77, opened 2023-10-11
Adding new validation job

BobaFetters:
- Adding new job to run nightly and with releases to validate tests on all platforms
- Updated PR Subtree CI to always run each subtree job regardless of if the others fails

### DIFF
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -1,0 +1,312 @@
+name: "Main Branch Health"
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 7 * * *"
+
+env:
+  XCODE_VERSION: "15.0.0"
+
+jobs:
+  run-swift-builds:
+    runs-on: macos-13
+    timeout-minutes: 15
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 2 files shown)


.github/workflows/release-check.yml
```diff
@@ -0,0 +1,312 @@
...
+          TestResults/ResultBundle.zip
+
+  legacy-unit-test:
+    runs-on: macos-11
+    timeout-minutes: 20
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          # iOS 14.4
```
calvincestari, line 137: I wonder if the legacy tests should test on the latest of each of the major iOS versions, e.g.: `13.7`, `14.8.1`, `15.7.9`, `16.7.1`?

I kinda regard `14.4` as a special case that we could have another job for; I'm not certain if the differing behavior would get caught by `13.7`?

BobaFetters: Guess it depends on how fine grained we want to be on testing older versions, originally i know we talked about doing 14.4 and newest, but I'm open to adding other versions as well

calvincestari (approved):

PR #83, opened 2023-10-12
Caching tuist output to share between jobs

BobaFetters:
Adding caching of the tuist generated files to share between jobs


.github/workflows/ci-tests.yml
```diff
@@ -25,16 +49,14 @@ jobs:
...
-      id: setup-repo
-      uses: ./.github/actions/setup-repo
     - name: Run Swift Build
       shell: bash
       run: |
         cd ${{ matrix.package }} && swift build
 
   build-and-unit-test:
     runs-on: macos-13
+    needs: tuist-generation
```
calvincestari, line 59: `needs` is a confusing property name for this; should be `required-first` or `after`.

calvincestari (approved):
👍🏻

.github/workflows/ci-tests.yml
BobaFetters: yea not sure how they landed on that name

PR #89, opened 2023-10-16
[2/x] Refactor root field builder for deferred named fragments

calvincestari:
This second PR for https://github.com/apollographql/apollo-ios/issues/3141 refactors the root field builder to build deferred named fragments into the IR. The next PR will generate selection sets from the refactored IR.

* Adds IR matchers for named fragment.
* Refactored logic for `containsDeferredFragment`; moved to where the scope condition is modified to include the defer condition.
* Refactored add selections logic for `.fragmentSpread`


calvincestari:
I'm gonna merge this into https://github.com/apollographql/apollo-ios-dev/pull/74 so the whole root field builder work can be evaluated as a whole. Especially since this one makes changes to code that's already in that PR.

PR #94, opened 2023-10-20
Adding name spacing to ApolloAPI.Object

BobaFetters:
-Name spacing our usage of ApolloAPI.Object in generated code in order to avoid conflicts with potential user defined custom scalar named 'Object'

Closes apollographql/apollo-ios#3258


calvincestari (approved):
This is great Zach, thanks for the quick fix!

Do you know what's up with the operation identifiers changing in StarWarsAPI? Is it just a module we didn't regenerate yet since the operation identifier work was done?

calvincestari:
> Do you know what's up with the operation identifiers changing in StarWarsAPI? Is it just a module we didn't regenerate yet since the operation identifier work was done?

We should make sure it's not some changing default behaviour showing up in that module because it uses a different setting, etc.

BobaFetters:
> > Do you know what's up with the operation identifiers changing in StarWarsAPI? Is it just a module we didn't regenerate yet since the operation identifier work was done?
> 
> We should make sure it's not some changing default behaviour showing up in that module because it uses a different setting, etc.

Yea was trying to figure that out now, guessing maybe they just hadn't been regenerated but need to find out for sure

PR #95, opened 2023-10-23
Remove import module for Test Mock when output schemeType is not in Module

Cookiezby:
Related to the issue reported here https://github.com/apollographql/apollo-ios/issues/3262

### DIFF
--- a/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
+++ b/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
@@ -295,7 +295,7 @@ struct ImportStatementTemplate {
     static func template(for config: ApolloCodegen.ConfigurationContext) -> TemplateString {
       return """
       import ApolloTestSupport
-      import \(config.schemaModuleName)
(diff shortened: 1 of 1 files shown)


apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
```diff
@@ -295,7 +295,7 @@ struct ImportStatementTemplate {
     static func template(for config: ApolloCodegen.ConfigurationContext) -> TemplateString {
       return """
       import ApolloTestSupport
-      import \(config.schemaModuleName)
+      \(if: config.output.schemaTypes.isInModule, "import \(config.schemaModuleName)")
```
calvincestari, line 298: This might work for your particular configuration but it will break other configurations that use the `none` or `swiftPackage` options. The reason the schema module is always imported here is because it is intended to always be a separate module; [see my comment on your issue](https://github.com/apollographql/apollo-ios/issues/3262#issuecomment-1775676144).

Cookiezby: Got it, thanks for the explanation, will try to put the mock data in the test target!

PR #98, opened 2023-10-24
Move file uploads doc to its own page

Meschreiber:
This PR moves the file uploads info to its own page for greater discoverability in the search bar.

It purposefully does _not_ add file uploads to the nav bar, since this isn't really a functionality we recommend. (Folks will be able to find it if they actively search for "file uploads" though.


calvincestari:
@Meschreiber you mention that adding it to the nav bar is not recommended but does that mean the only way users will be able to find this new page is now through a search?

Looking at the Kotlin docs they do have the [uploads page in the nav bar](https://www.apollographql.com/docs/kotlin/advanced/upload/) - see [config.json](https://github.com/apollographql/apollo-kotlin/blob/main/docs/source/config.json#L72) too. I think it's helpful to have it there where it's visible beyond just search.

Meschreiber:
@calvincestari , sure we can add it! I was actually referencing how React did theirs (with no [file-uploads.mdx](https://github.com/apollographql/apollo-client/blob/main/docs/source/data/file-uploads.md) in the [config.json](https://github.com/apollographql/apollo-client/blob/main/docs/source/config.json)), but am happy to align ios with Kotlin. (And of course, we should align all three clients at some point.)

Do you have an inclination for the best section to stick it in the ios nav?

calvincestari:
> (And of course, we should align all three clients at some point.)

Yes, I agree.

> @calvincestari , sure we can add it! I was actually referencing how React did theirs (with no [file-uploads.mdx]Do you have an inclination for the best section to stick it in the ios nav?

Meschreiber:
Hmm, I feel like Networking is an option, but maybe not exactly where folks would expect to see it.

As strange as it feels having a section with only one page in it, I think we can mirror the Kotlin docs nav and stick it in an "Advanced" section that we can continue to grow with other pages.

calvincestari:
> As strange as it feels having a section with only one page in it, I think we can mirror the Kotlin docs nav and stick it in an "Advanced" section that we can continue to grow with other pages.

Sure, I'm fine with a new "Advanced" section. We can probably move the "Advanced Network Configuration" page into that new section too and retitle it "Customizing the Request Chain"?

calvincestari (approved):
LGTM!

Meschreiber:
Yay, thanks for the collaboration @calvincestari ! Looks like I'll need you to do the merge since I don't have write access to this repo. :)

PR #102, opened 2023-10-25
Support SelectionSet to receive [String: Any] for initialization when create mock object

Cookiezby:
Relate to issue https://github.com/apollographql/apollo-ios/issues/3264

If the data object which has type `[String: AnyHashable]` includes nested dictionary object, the compile will throw error, but it is fine to define the object with `[String: Any]`, however, the DataDict only accept `[String: AnyHashable]` so there need a type conversion from `[String: Any]` to [`String: AnyHashable]`

I am not sure if it is better to add new convenience init method for DataDict type, so I only provide an extension as a help method

### DIFF
--- a/apollo-ios/Sources/ApolloTestSupport/SelectionSet+DictionaryIntializer.swift
+++ b/apollo-ios/Sources/ApolloTestSupport/SelectionSet+DictionaryIntializer.swift
@@ -16,21 +16,21 @@ extension RootSelectionSet {
     dict: [String: Any],
     variables: GraphQLOperation.Variables? = nil
   ) throws {
-    let jsonObject = Self.converDictToJSONObject(dict: dict)
+    let jsonObject = Self.convertDictToJSONObject(dict: dict)
     try self.init(data: jsonObject, variables: variables)
   }
 
   /// Convert dictionary type [String: Any] to JSONObject
   /// - Parameter dict: dictionary value
   /// - Returns: converted JSONObject
-  static func converDictToJSONObject(dict: [String: Any]) -> JSONObject {
+  static func convertDictToJSONObject(dict: [String: Any]) -> JSONObject {
     var result = JSONObject()
 
     for (key, value) in dict {
       if let hashableValue = value as? AnyHashable {
         result[key] = hashableValue
       } else if let dictValue = value as? [String: Any] {
-        result[key] = converDictToJSONObject(dict: dictValue)
+        result[key] = convertDictToJSONObject(dict: dictValue)
       }
     }
     return result
--- a/apollo-ios/Sources/ApolloTestSupport/SelectionSet+DictionaryIntializer.swift
+++ b/apollo-ios/Sources/ApolloTestSupport/SelectionSet+DictionaryIntializer.swift
@@ -1,0 +1,38 @@
+#if !COCOAPODS
+import ApolloAPI
+import Apollo
+#endif
+
+extension RootSelectionSet {
+  // Initializes a `SelectionSet` with a raw Dictionary.
+  ///
+  /// We could pass dictionary object to create a mock object for GraphQL object.
+  ///
+  /// - Parameters:
+  ///   - dict: A dictionary representing a dictionary response for a GraphQL object.
+  ///   - variables: [Optional] The operation variables that would be used to obtain
+  ///                the given JSON response data.
+  public init(
+    dict: [String: Any],
+    variables: GraphQLOperation.Variables? = nil
+  ) throws {
+    let jsonObject = Self.converDictToJSONObject(dict: dict)
+    try self.init(data: jsonObject, variables: variables)
+  }
+
+  /// Convert dictionary type [String: Any] to JSONObject
+  /// - Parameter dict: dictionary value
+  /// - Returns: converted JSONObject
+  static func converDictToJSONObject(dict: [String: Any]) -> JSONObject {
+    var result = JSONObject()
+
+    for (key, value) in dict {
+      if let hashableValue = value as? AnyHashable {
+        result[key] = hashableValue
+      } else if let dictValue = value as? [String: Any] {
+        result[key] = converDictToJSONObject(dict: dictValue)
+      }
+    }
+    return result
+  }
+}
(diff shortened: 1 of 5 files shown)


AnthonyMDev:
Thanks for much for the contribution @Cookiezby! 

I think that this issue might be a little more complicated than this PR addresses though!

For what you are doing, you should actually be using the special json initializer we have in `RootSelectionSet+JSONInitializer.swift`. Because nested selections sets and custom scalars aren’t going to be properly transformed if you just initialize the `DataDict` with your own JSON data.

The [example in our docs here](https://www.apollographql.com/docs/ios/testing/test-mocks/#mocking-graphql-responses-with-json) should probably be updated to use that as well. However the [issue #3264](https://github.com/apollographql/apollo-ios/issues/3264) won't be addressed by that alone, as it still takes in a `[String: AnyHashable]` dictionary.

We should probably add another method there that takes in a `[String: Any]` json dict and does a conversion of that internally. Rather than the proposed `.asAnyHashable` being `public` as this PR suggests.

We also should include the utility requested in [this issue](https://github.com/apollographql/apollo-ios/issues/3196) to convert the selection sets back into raw JSON.

If you would like to take a stab at making the suggested changes, please let me know, and feel free to update this PR.

AnthonyMDev:
It's worth noting that we already have `AnyHashableConvertible` as a protocol in `ApolloAPI` that could likely be used for this purpose as well (under the hood of the function implementation).

Cookiezby:
@calvincestari @AnthonyMDev Hi, thanks for your reviews, I am willing to update my PR for the suggested change, will check the detail about the RootSelectionSet+JSONInitializer.swift.

> We should probably add another method there that takes in a [String: Any] json dict and does a conversion of that internally. Rather than the proposed .asAnyHashable being public as this PR suggests

.

Cookiezby:
@AnthonyMDev Hi, can you help to confirm if I totally understand your suggestions

> For what you are doing, you should actually be using the special json initializer we have in RootSelectionSet+JSONInitializer.swift. Because nested selections sets and custom scalars aren’t going to be properly transformed if you just initialize the DataDict with your own JSON data.

Suppose I have a generated data structure
```swift
struct Document: SelectionSet {
   //.......
}
```
Do you mean we should init the `Document` by using the initialization method defined in SelectionSet+JSONInitializer.swift instead of padding a DataDict to initialize the data

```swift
// ✅ 
let jsonObject: JSONObject = [:]
let mockDocument = Document(data: jsonObject)

// ❌
let dict: [String: AnyHashable] = [:]
let mockDocument = Document(dataDict: DataDict(dict))
```

And because the JSONObject is still equal to [String: AnyHashable] so the next step I need to do is to add another initialization method for SelectionSet so that it can takes [String: Any] type of data for initialization, and in the method we would convert [String: Any] to JSONObject internally

AnthonyMDev:
Yes, that's right!

apollo-ios/Sources/ApolloTestSupport/SelectionSet+DictionaryIntializer.swift
```diff
@@ -0,0 +1,38 @@
+#if !COCOAPODS
```
Cookiezby, line 1: I put the new initialization method in ApolloTestSupport target to make sure this method is only used for mock purpose

Cookiezby:
@AnthonyMDev Thanks for your confirmation! I just update the code, could you help to check it?

AnthonyMDev:
> As I mentioned in [this comment](https://github.com/apollographql/apollo-ios/issues/3264#issuecomment-1788248483) in the `apollo-ios` repo, this is just a Swift compiler issue and can be solved by adding `as [String: AnyHashable]` to the nested dictionaries. I question whether it makes sense to add another initializer that allows for passing `[String: Any]` since internally Apollo needs the values to be `AnyHashable`.

Thanks for the feedback. While that is possible, I don't see any harm in adding the helper method here to make people's lives easier working with this.

AnthonyMDev (changes requested):
I actually think we should include this right alongside the other initializer in `RootSelectionSet+JSONInitializer.swift`. We should probably put `@_disfavoredOverload` on it, to prevent ambiguity issues also.

It should also have the parameter named `data` just like the other initializer.

Lastly, I'm wondering if we also need to do conversions of arrays (or arrays of dictionaries (`[[String: Any]]`) as well. 

@Cookiezby Can you test this with some data with arrays in it and add array conversion if its needed?

bdbergeron:
> > As I mentioned in [this comment](https://github.com/apollographql/apollo-ios/issues/3264#issuecomment-1788248483) in the `apollo-ios` repo, this is just a Swift compiler issue and can be solved by adding `as [String: AnyHashable]` to the nested dictionaries. I question whether it makes sense to add another initializer that allows for passing `[String: Any]` since internally Apollo needs the values to be `AnyHashable`.
> 
> Thanks for the feedback. While that is possible, I don't see any harm in adding the helper method here to make people's lives easier working with this.

My primary concern is that this initializer swallows any non-`Hashable` objects a developer may pass to it, without indicating it's doing so. At a minimum I think this new initializer should `throw` if an value in the dictionary cannot be converted to `AnyHashable`.

AnthonyMDev:
Oh yes! That's a great point, and one I missed. I 100% agree with that.

AnthonyMDev (changes requested):
Sorry if I wasn't clear about this! I think we should just move this method into the SelectionSet+JSONInitializer.swift` file that already exists.

Also, this implementation currently only will handle 1 dimensional arrays of objects. We need it to handle more arrays. As this is getting more complex, I'm thinking we might want to add some tests for this as well so we can verify we are handling all the different ways of doing this.

@Cookiezby Do you feel like you're still willing to work through these complications?

Cookiezby:
@AnthonyMDev Yes, I am willing to help with that, for the one dimensional object, I will add tests for that, and when I check the test case in the **SelectionSetTests** I found the nested array in some test cases, is this the `more arrays` that you mentions in here?

> Also, this implementation currently only will handle 1 dimensional arrays of objects. We need it to handle more arrays

```swift
 let expected = try! Hero(
      data: [
        "__typename": "Human",
        "nestedList": [[]]
      ],
      variables: nil
)
```

Cookiezby:
@AnthonyMDev Hi, I have added the implementation for handling the array (including nested array) when initialize the SelectionSet with dictionary object, and I also added some test cases for the new initializer, can you help to review it?

AnthonyMDev (approved):
This looks great to me! Thank you so much @Cookiezby for making this one happen!

PR #103, opened 2023-10-25
Selectively run Tuist Generation when there are changes to the Tuist configuration

Iron-Ham:
The purpose of this pull request is to speed up CI & to save y'all some actions runner minutes.

Running `Tuist generate` as a step testing the validity of the project shouldn't be necessary if there have been no changes to the configuration. To verify proper function: I will branch off of this PR with a change to Tuist. This pull request, having no changes to the configuration shouldn't run the action. The other, with changes to the configuration, should run the action.

cc: @BobaFetters, since you have the most context on the configuration: Is the filter I've setup sufficient? 

Edit: See #104 for an example of a pull request that triggers running `Run Tuist Generation`. 

EDIT2: I see that the broader matrix for testing this is reliant on the Tuist generate step. 🤔 Another thought would be to selectively filter & run tests based on the filter. I'll give it some more thought.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -8,8 +8,25 @@ env:
   XCODE_VERSION: "15.0.0"
 
 jobs:
+  changes:
+    runs-on: ubuntu-latest
+    outputs:
+      project: ${{ steps.filter.outputs.project }}
+    steps:
+      - uses: actions/checkout@v4
+      - uses: dorny/paths-filter@v2
+        id: filter
+        with:
+          filters: |
+            project:
+              - 'Workspace.swift'
+              - 'Tuist/**'
+              - 'Project.swift'
+
   tuist-generation:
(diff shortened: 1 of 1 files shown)


Iron-Ham:
Simple follow-up has been placed into #80 (I don't mind removing it, but unfortunately since I don't have access to the `apollographql/` repo, I cannot point a pull request to #80 as its base).

The follow up is basically: Only run the pagination tests when there are changes that could impact Pagination. Namely: If there are changes in Pagination, its tests, or `apollo-ios`.

BobaFetters:
@Iron-Ham Thanks for looking into/adding this! We had talked about maybe trying to filter selectively based on changes in `apollo-ios` or `apollo-ios-codegen` etc, looks like what you are doing with pagination would work great across the board.

Iron-Ham:
Assuming i can get it to play nice with the matrix, it'd be a decent jumping off point!.

PR #106, opened 2023-10-27
docs: Update subscriptions.mdx

zohairhadi:
Use the config parameter instead of connectingPayload which has been deprecated. Add the authPayload in the WebSocketTransport.Configuration instead.


calvincestari (approved):
Thank you for updating the docs @zohairhadi. I think this has been out-of-date for quite some time now. Apologies for the confusion.

PR #107, opened 2023-10-27
Fix CI by updating Xcode version to 15.0.1

Iron-Ham:
Seems like 15.0.0 is no longer available.

Screenshot 2023-10-27 at 11 35 02 AM

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -5,7 +5,7 @@ on:
     types: [opened, synchronize, reopened]
 
 env:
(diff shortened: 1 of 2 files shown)


calvincestari (approved):
Thanks Hesham. [It looks like `15.0.0` is still available](https://github.com/actions/runner-images/blob/main/images/macos/macos-13-Readme.md#xcode) but we may have to drop the patch version and target `15.0` specifically?

Either way 15.0.1 is probably what we should be using - thanks for updating this.

PR #108, opened 2023-10-30
Adding back xcbeautify

BobaFetters:
Adding xcbeautify back to ci since we aren't currently running older os tests where its not supported


calvincestari (approved):
Thanks!

PR #114, opened 2023-10-31
[ApolloPagination] Load All Pages

Iron-Ham:

> [!IMPORTANT]
> This pull request relies on #80, but I cannot point this branch to that branch _and_ attempt to open it up as pull request in this repository (as I'm an unaffiliated contributor to the repository). As such, the overwhelming majority of commits on this branch are not part of this pull request; only the final 4 commits are the contents of this pull request. 

I want consumers of the `GraphQLPager` to be able to load all pages at once, should they desire to do so. 

First, I modified the `fetch` function to follow `async`/`await` syntax. That's so that we can `await` its return prior to beginning to loop over `loadMore`. 

Then, I implemented the function using a while loop.

Finally, I made additive changes to the `MockGraphQLServer` to support subsequent loads of this nature. 

Yes: This doesn't work for ReversePagination -- and that's okay. ReversePagination is due for some re-working as part of https://github.com/apollographql/apollo-ios/issues/3265

______

### Summary
🧪🔄🔒

This pull request adds a new feature to the `GraphQLQueryPager` protocol and class, which allows loading all the pages of a paginated GraphQL query in one call. It also updates the tests and the mock GraphQL server to use the new feature and the async/await syntax. Additionally, it removes some unnecessary public modifiers from internal extensions.

> _Sing, O Muse, of the swift and skillful coder_
> _Who added to the `GraphQLQueryPager` a wondrous feature_
> _To load all pages of a query with one call, like a mighty river_
> _That flows from the mountains to the sea, unstoppable and eager._

### Walkthrough
*  Add `loadAll` methods to `AnyGraphQLPager` and `GraphQLQueryPager` protocols and classes to allow loading all pages of a paginated query in one call
*  Refactor `GraphQLQueryPager.Actor` class to use async/await syntax, `initialFetchTask` property, and `isLoadingAll` flag for pagination logic and cancellation handling
*  Add test cases for `AnyGraphQLQueryPager` and `ForwardPaginationTests` using mock GraphQL server and expectations
*  Add subscript operator to `MockGraphQLServer` class to access and set `RequestExpectation` objects for different operation types
*  Modify `FriendsQuery` extension methods to use specific query instances with expected variables instead of generic operation type
*  Remove unnecessary `public` modifiers from `GraphQLQueryPager.Actor` extensions in `AnyGraphQLPager.swift` and `Pager+Convenience.swift` files
*  Move deinitializer after initializer and remove `public` modifier from initializer in `GraphQLQueryPager` class
*  Add missing import statement to `FriendsQuery+TestHelpers.swift` file


Iron-Ham:
Closing in favor of #115

PR #115, opened 2023-10-31
[Apollo Pagination] Improve ReversePagination support, implement `loadAll` support, Bidirectional Pagination

Iron-Ham:

1. I want consumers of the `GraphQLPager` to be able to load all pages at once, should they desire to do so. 
2. In a quest to deliver bidirectional pagination, we must first solidify support for reverse pagination. 

closes https://github.com/apollographql/apollo-ios/issues/3265

**LoadAll Support**: 
First, I modified the `fetch` function to follow `async`/`await` syntax. That's so that we can `await` its return prior to beginning to loop over `loadMore`. 

Then, I implemented the function using a while loop.

Finally, I made additive changes to the `MockGraphQLServer` to support subsequent loads of this nature. 

**Reverse Pagination**:
Leaning hard into differentiating reverse and forward pagination from one another. 
If we are to support _both_ types of pagination at once, then they must be treated as being separate from one another: A page can be a previous page _or_ it can be a next page. It cannot be both. 

**BIdirectional Pagination**
Once reverse pagination was solidifed, this was a breeze to implement. It basically amounted to defining the pagination struct. 

Note that the output for the pagination controller is now a struct. This allows us to make changes to the output without breaking previous versions. I've removed logic around `Task` and `Continuation` management.

______

### Copilot Summary

This pull request primarily introduces improvements to the Apollo GraphQL testing suite and pagination functionality. The most significant changes include the addition of a new `BidirectionalPaginationTests` class for testing bidirectional pagination, modifications to the `MockGraphQLServer` class to enhance testing capabilities, and updates to the `AnyGraphQLQueryPagerTests` and `ForwardPaginationTests` classes to accommodate new pagination functionalities.

Enhancements to Apollo GraphQL testing:

* <a href="diffhunk://#diff-9182276fe42a90e862b08125a7a6e055e60bb3b510e9840c92efe9284634fe5cR79-R88">`Tests/ApolloInternalTestHelpers/MockGraphQLServer.swift`</a>: Added a private subscript to the `MockGraphQLServer` class to handle request expectations for a given operation type. Also, a new `expect` method was added to handle request expectations for a specific operation. <a href="diffhunk://#diff-9182276fe42a90e862b08125a7a6e055e60bb3b510e9840c92efe9284634fe5cR79-R88">[1]</a> <a href="diffhunk://#diff-9182276fe42a90e862b08125a7a6e055e60bb3b510e9840c92efe9284634fe5cR100-R114">[2]</a>

Improvements to Apollo GraphQL Pagination:

* <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L58-R58">`Tests/ApolloPaginationTests/AnyGraphQLQueryPagerTests.swift`</a>: Modified the `eraseToAnyPager` method to include an additional parameter. Also, added a new `test_loadAll` method to test the loading of all pages. Several other minor changes were made to accommodate updates to the pagination functionality. <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L58-R58">[1]</a> <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7R92-R106">[2]</a> <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L103-R125">[3]</a> <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L129-R145">[4]</a>
* <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L50-R50">`Tests/ApolloPaginationTests/ForwardPaginationTests.swift`</a>: Updated several methods to reflect changes in the pagination functionality. <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L50-R50">[1]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L60-R64">[2]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L75-R96">[3]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L111-R113">[4]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L125-R127">[5]</a>
* <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L28-R28">`Tests/ApolloPaginationTests/ConcurrencyTest.swift`</a>: Updated several methods to reflect changes in the pagination functionality. <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L28-R28">[1]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L51-R51">[2]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L78-R82">[3]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L94-R98">[4]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L111-R111">[5]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L127-R134">[6]</a>

Addition of new test class:

* <a href="diffhunk://#diff-eb8712982f5deb2896d926c2943479d019c26243bcdc5d05ef810e4b84ad499fR1-R197">`Tests/ApolloPaginationTests/BidirectionalPaginationTests.swift`</a>: Added a new `BidirectionalPaginationTests` class to test bidirectional pagination functionality. This class includes several methods to test fetching multiple pages, loading all pages, and more.


apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -37,22 +60,27 @@ public class GraphQLQueryPager<InitialQuery: GraphQLQuery, PaginatedQuery: Graph
     client: ApolloClientProtocol,
     initialQuery: InitialQuery,
     extractPageInfo: @escaping (PageExtractionData) -> P,
-    nextPageResolver: @escaping (P) -> PaginatedQuery
+    nextPageResolver: ((P) -> PaginatedQuery)?,
+    previousPageResolver: ((P) -> PaginatedQuery)?
```
AnthonyMDev, line 64: Would it be cleaner to have a single page resolver with a parameter for next/previous?

```swift
enum PaginationDirection {
    case next
    case previous
}

...
pageResolver: ((P, PaginationDirection) -> PaginatedQuery)?
```

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -37,22 +60,27 @@ public class GraphQLQueryPager<InitialQuery: GraphQLQuery, PaginatedQuery: Graph
...
+      previousPageVarMapPublisher.combineLatest(initialPublisher, nextPageVarMapPublisher).sink { [weak self] _ in
         guard let self else { return }
         Task {
-          let value = await self.pager.pageTransformation()
-          self.canLoadNextSubject.send(value?.canLoadMore ?? false)
+          let canLoadNext = await self.pager.canLoadNext
+          let canLoadPrevious = await self.pager.canLoadPrevious
+          self.canLoadNextSubject.send(canLoadNext)
+          self.canLoadPreviousSubject.send(canLoadPrevious)
         }
```
AnthonyMDev, line 84: Does this also need a to stored in `cancellables`?

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -37,22 +60,27 @@ public class GraphQLQueryPager<InitialQuery: GraphQLQuery, PaginatedQuery: Graph
...
+      let nextPageVarMapPublisher = await pager.$nextPageVarMap
       let initialPublisher = await pager.$initialPageResult
-      varMapPublisher.combineLatest(initialPublisher).sink { [weak self] _ in
+      previousPageVarMapPublisher.combineLatest(initialPublisher, nextPageVarMapPublisher).sink { [weak self] _ in
         guard let self else { return }
         Task {
-          let value = await self.pager.pageTransformation()
-          self.canLoadNextSubject.send(value?.canLoadMore ?? false)
+          let canLoadNext = await self.pager.canLoadNext
+          let canLoadPrevious = await self.pager.canLoadPrevious
```
AnthonyMDev, line 81: Same as above, [actor isolation](https://github.com/apple/swift-evolution/blob/main/proposals/0313-actor-isolation-control.md) could allow us to only have a single await here if the `Actor` has a property that returned both of these.

```swift
/// on Actor
var canLoadPages: (next: Bool, previous: Bool) {
    (canLoadNext, canLoadPrevious)
}
```

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -74,20 +98,50 @@ public class GraphQLQueryPager<InitialQuery: GraphQLQuery, PaginatedQuery: Graph
...
+      do {
+        try await pager.loadPrevious(cachePolicy: cachePolicy)
+        await completion?(nil)
+      } catch {
+        await completion?(error)
+      }
+    }
+  }
+
   public func loadMore(
```
AnthonyMDev, line 123: I like the function being called `loadMore()` for the single direction case. But since with bidirectional pagination everything else is labeled as `previous` and `next`, I wonder if we should rename this to `loadNext()`.

There is also the option of adding a `loadNext()` function that just calls `loadMore()`.

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -310,33 +495,110 @@ extension GraphQLQueryPager {
...
           publisher.send(completion: .finished)
         }
-        varMap[variables] = nextPageData
+      case .failure(let error):
+        currentValue = .failure(error)
+        publisher.send(completion: .finished)
+      }
+    }
+
+    private func onPreviousFetch(
```
AnthonyMDev, line 525: If we add a `PaginationDirection` enum (as suggested above), we could probably combine `onSubsequentFetch` and `onPreviousFetch` into a single function that takes in a `direction`.

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
```diff
@@ -5,43 +5,45 @@ import Combine
...
       }
 
-      self._subject?.send(returnValue)
+      _subject?.send(returnValue)
     }
   }
 
   /// Type-erases a given pager, transforming the initial page to an array of models, and the
-  /// subsequent pagination to an additional array of models, concatenating the results of each into one array.
+  /// subsequent pagination to an adition array of models, concatenating the results of each into one array.
```
AnthonyMDev, line 46:
```suggestion
  /// subsequent pagination to an additional array of models, concatenating the results of each into one array.
```

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
```diff
@@ -54,35 +56,53 @@ public class AnyGraphQLQueryPager<Model> {
...
-  /// Subscribe to new pagination `Output`s.
-  /// - Parameter completion: Receives a new `Output` for the consumer of the API.
-  /// - Returns: A `Combine` `AnyCancellable`, such that the caller can manage its own susbcription.
-  @discardableResult public func subscribe(completion: @escaping (Output) -> Void) -> AnyCancellable {
+  deinit {
+    pager.cancel()
+  }
+
+  @discardableResult public func subscribe(completion: @MainActor @escaping (Output) -> Void) -> AnyCancellable {
     guard let _subject else { return AnyCancellable({ }) }
```
AnthonyMDev, line 77: Does this `AnyCancellable` need to be stored in `cancellables`? Even though it doesn't do anything, it should still get a cancellation when the pager is cancelled, shouldn't it?

AnthonyMDev (commented):
Lots of great work here! I have a few suggestions, but please let me know if you disagree with any of them!

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
Iron-Ham: It's being stored on line 85, unless I'm misunderstanding?

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
Iron-Ham: I think this is a situation where we're going to have to decide what pattern we want to follow. We can either return the cancellable, or store it internally, but we probably shouldn't do both.

apollo-ios-pagination/Sources/ApolloPagination/CursorBasedPagination/BidirectionalPagination.swift
```diff
@@ -0,0 +1,23 @@
+extension CursorBasedPagination {
+  public struct BidirectionalPagination: PaginationInfo, Hashable {
```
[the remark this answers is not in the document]
Iron-Ham: I think that's super reasonable.

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
Iron-Ham: It's a tough one; I'm inclined to agree with you on this. Perhaps we can mark this as an area for future consideration; thinking outside of the FRP-world, I generally try to contain all of those concepts so that folks that aren't using an FRP paradigm don't suddenly have to juggle these concepts.

🤔

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
AnthonyMDev: Yupp, I don't know how I missed that. Disregard.

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
AnthonyMDev: >  I generally try to contain all of those concepts so that folks that aren't using an FRP paradigm don't suddenly have to juggle these concepts.

Yeah, I agree with that in general. But I'm thinking ahead to Apollo 2.0, and adopting modern swift concurrency is one of our biggest goals (async/await support is a highly requested features). But because of the multiple completion handler calls (from the cache, then server) we make, that is likely going to require us to expose `Combine` streams and other FRP concepts as return values for network requests.

As a singular feature library, I think that philosophy is usually correct, but Apollo is more of an entire SDK. Between model generation, caching, and networking request handling, it is almost always integrated into the heart of your project's infrastructure. Apple is steering the Swift language and `SwiftUI` development towards FRP with `Combine` and `async/await`. So I think it's appropriate for an SDK like Apollo to follow similar API design concepts as them.

---

This pagination module is an interesting case though, because it's being released prior to the 2.0 redesign. So it would be the first place we are exposing those concepts. On one hand, that gives us an opportunity to test the waters with this, and see how users respond; on the other, it may be inconsistent with the API design language/style of Apollo iOS.

So now, I'm leaning back towards exposing the `Cancellable`! 🙄 

CC: @bignimbus @apollographql/prometheus in case any of you have an opinion on this API design philosophy question.

Iron-Ham: @AnthonyMDev perhaps as a compromise:

```swift
  public func subscribe(completion: @MainActor @escaping (Output) -> Void) -> AnyCancellable {
    guard let _subject else { return AnyCancellable({ }) }
    return _subject.compactMap({ $0 }).sink { result in
      Task {
        await completion(result)
      }
    }
  }

  public func subscribe(completion: @MainActor @escaping (Output) -> Void) {
    subscribe(completion: completion).store(in: &cancellables)
  }
```

Tests/ApolloPaginationTests/ConcurrencyTest.swift
```diff
@@ -75,11 +75,11 @@ final class ConcurrencyTests: XCTestCase {
...
-      group.addTask { try? await pager.loadMore() }
-      group.addTask { try? await pager.loadMore() }
+      group.addTask { try? await pager.loadNext() }
+      group.addTask { try? await pager.loadNext() }
+      group.addTask { try? await pager.loadNext() }
+      group.addTask { try? await pager.loadNext() }
+      group.addTask { try? await pager.loadNext() }
 
       group.addTask { await self.fulfillment(of: [serverExpectation, expectation], timeout: 100) }
       await group.waitForAll()
```
AnthonyMDev, line 85: Nit: We usually put test helpers above the tests. And I think these three load data functions should be underneath the `// MARK: - Test Helpers` header as well.

Tests/ApolloPaginationTests/GraphQLQueryPagerTests.swift
```diff
@@ -0,0 +1,301 @@
...
+        pager.cancel()
+    }
+
+    wait(for: [secondPageExpectation], timeout: 2)
+    XCTAssertEqual(results.count, 1) // once for original fetch
+    XCTAssertEqual(errors.count, 2)
+    XCTAssertTrue(errors.contains(where: { $0 == .cancellation }))
+  }
+
+  @available(iOS 16.0, macOS 13.0, *)
```
AnthonyMDev, line 145: @BobaFetters, the Tuist configuration work you're doing now should make it possible for us to remove these, right? We should be able to set it up so that the pagination test target has a higher deployment target configured, since the pagination SPM package has a higher minimum platform version anyways.

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -1,342 +1,207 @@
...
+    execute(completion: completion) { [weak self] in
+      try await self?.pager.loadNext(cachePolicy: cachePolicy)
+    }
+  }
+
+  /// Loads all pages.
+  /// - Parameters:
+  ///   - fetchFromInitialPage: Pass true to begin loading from the initial page; otherwise pass false.  Defaults to `true`.  **NOTE**: Loading all pages with this value set to `false` requires that the initial page has already been loaded previously.
+  ///   - completion: An optional error closure that triggers in the event of an error. Defaults to `nil`.
+  public func loadAll(fetchFromInitialPage: Bool = true, completion: ((Error?) -> Void)? = nil) {
```
AnthonyMDev, line 128: I'm just now thinking... why are `loadNext()` and `loadAll()` using completion blocks instead of being `async throws` functions? If we're using `async/await` anyways, why aren't we using it for the `public` callsites?

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -1,342 +1,207 @@
...
+  var subscriptions: Set<AnyCancellable> = []
 
-    /// Whether or not we can load more information based on the current page.
-    public var canLoadNext: Bool {
-      currentPageInfo?.canLoadMore ?? false
-    }
+  func store(subscription: AnyCancellable) {
+    subscriptions.insert(subscription)
+  }
+}
```
AnthonyMDev, line 172: Is it better for us to have these `Subscriptions` as a seperate `actor` from the main `GraphQLQueryPager.Actor`? I imagine that it allows for subscriptions to be accessed in parallel to loading of pages being kicked off, so that might be better for performance. 

Is there any risk of this causing race conditions though? Could a subsscription possibly not be stored in time to get some results it would expect to get when the `Actor` gets results?

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
AnthonyMDev: Oh is the completion ony called if an error occurs? If so, that needs to be more clearly documented.

Is the idea here that you would normally just call `loadAll()` and then move on, not needing to `await` on it because once it gets data back, that will get pushed into the subscribers via the `Combine` subject?

I'm feeling like there is something off with this API design. It's kind of confusing that you get the results by subscribing, but the errors through a completion handler. 

It's unclear to me what the difference is between getting an `Error` from this completion block vs. getting a `Result.failure` with an error from the subscriber's output.

I'm not sure what the right solution to this is yet. Happy to chat about this with you some more though!

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -1,342 +1,207 @@
...
 
-    // MARK: - Private
+private class Completion {
```
AnthonyMDev, line 174: Can this be a `struct` instead?

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -1,342 +1,207 @@
...
-        guard let firstPageData = data.data else { return }
-        if let latest {
-          let (_, nextPage) = latest
-          currentValue = .success((firstPageData, nextPage, data.source == .cache ? .cache : .fetch))
-        }
-      case .failure(let error):
-        currentValue = .failure(error)
-      }
+  func execute(error: Error?) async {
+    await MainActor.run { [weak self] in
```
AnthonyMDev, line 182: If we feel the need to run these on the `MainActor`, instead of `MainActor.run` internally here, could we mark the `completion` parameters on the `loadNext()` and `loadAll()` with `@MainActor`?

Maybe I'm not 100% clear on how `MainActor` works, but I'm thinking that way it is explicitly shown to the consumer that the completion block they pass in to those methods _will_ be run on the main thread. Whereas now, it's a hidden implementation detail.

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -1,342 +1,207 @@
...
-      publisher: CurrentValueSubject<Void, Never>,
-      query: PaginatedQuery
-    ) {
-      switch result {
-      case .success(let data):
-        guard let nextPageData = data.data else {
-          publisher.send(completion: .finished)
-          return
-        }
+private actor CompletionManager {
```
AnthonyMDev, line 189: Same question I had about the `Subscriptions` actor. What is the value in having 3 different actors here, and does it open us the possiblity of race conditions?

We should probably have some inline documentation about the reasoning here on this for posterity.

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPagerOutput.swift
```diff
@@ -0,0 +1,32 @@
+import ApolloAPI
+import Foundation
+
+extension GraphQLQueryPager {
+  /// A struct which contains the outputs of pagination
+  public struct Output: Hashable {
+    /// An array of previous  pages, in pagination order
```
AnthonyMDev, line 7:
```suggestion
    /// An array of previous pages, in pagination order
```

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
```diff
@@ -5,38 +5,39 @@ import Combine
 /// Type-erases a query pager, transforming data from a generic type to a specific type, often a view model or array of view models.
 public class AnyGraphQLQueryPager<Model> {
   public typealias Output = Result<(Model, UpdateSource), Error>
-  public var canLoadNext: Bool { pager.canLoadNext }
+  public let _subject: CurrentValueSubject<Output?, Never> = .init(nil)
```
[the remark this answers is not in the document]
Iron-Ham: At GitHub, we have an extension on this class that I would think is entirely unneeded in the open source module. That extension remaps the `_subject` on `subscribe`: 

```swift
/// Maps results which include an update source to the legacy WatchResult type.
/// Note that this function is only implemented for `[Model]`, which was the most common consumer of `WatchResult`.
/// - Parameter result: Result containing an update source.
/// - Returns: Watch result mapped to the update source.
private func legacyMapper<Model, S: RangeReplaceableCollection>(result: Result<(S, UpdateSource), Error>) -> WatchResult<S> where S.Element == Model {
    switch result {
    case let .success(value):
        let (pages, update) = value
        return .success(pages, update)
    case .failure(let error):
        return .failure(error)
    }
}

extension AnyGraphQLQueryPager where Model: RangeReplaceableCollection {
    public func subscribe(completion: @MainActor @escaping (WatchResult<Model>) -> Void) {
        _subject.compactMap({ $0 }).sink { result in
            Task {
                await completion(legacyMapper(result: result))
            }
        }.store(in: &cancellables)
    }
}

```

The `WatchResult` exists to disambiguate between:

1. `success`
2. `partialSuccess`, where we got some data back but also got back `errors`
3. `failure`

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
```diff
@@ -1,342 +1,207 @@
...
+  var canLoadPrevious: Bool { get }
   func cancel()
-  func loadMore(
+  func loadPrevious(
     cachePolicy: CachePolicy,
-    completion: (@MainActor () -> Void)?
-  ) throws
+    completion: ((Error?) -> Void)?
+  )
+  func loadNext(
```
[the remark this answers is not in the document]
Iron-Ham: I really hesitated in even exposing a cache policy for `loadPrevious` and `loadNext`. I'm good with being super opinionated here and removing the cache policy arguments completely; if you `fetchIgnoringCacheCompletely` I question whether you'd need this API -- which uses watchers the whole way through. 

Initial fetches in practice I've only seen fit two buckets:

1. `returnCacheDataAndFetch`
2. `fetchIgnoringCacheData`

I would really hesitate to encourage the use of `returnCacheDataElseFetch`. If the data being requested is dynamic in any way, it's a great way to show users the wrong data.

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
Iron-Ham: I see where the confusion lies, and I think you're smelling something real.

The API structure is, from bottom to the top, something like this:

1. `GraphQLQueryPager.Actor`: An `actor` that can operate asynchronously and be easily used in `async` contexts. 
2. `GraphQLQueryPager`: A wrapper around `GraphQLQueryPager.Actor` which allows easier usage in non-async contexts. Think: Existing `UIKit` apps.
3. `AnyGraphQLQueryPager`: A wrapper around `GraphQLQueryPager` which erases its query, allowing the caller to map results to their own model. 

> Oh is the completion ony called if an error occurs? If so, that needs to be more clearly documented.

The completion is always triggered, but provides an `Error` if there was an error of some kind. I suppose it's an alternative to using `Result<Void, Error>`, where the success case just means there's no error.

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
Iron-Ham: > Is there any risk of this causing race conditions though? Could a subsscription possibly not be stored in time to get some results it would expect to get when the `Actor` gets results?


Risk seems minimal here. The subscriptions are represented via `CurrentValueSubject`, so it should receive a result as soon as it can. 


> Is it better for us to have these `Subscriptions` as a seperate `actor` from the main `GraphQLQueryPager.Actor`? I imagine that it allows for subscriptions to be accessed in parallel to loading of pages being kicked off, so that might be better for performance.

The `Subscriptions` actor only exists to manage the subscriptions array in `GraphQLQueryPager`. if we omit this `actor`, the alternative would be to store individual values for each `AnyCancellable` on the `GraphQLQueryPager`. Otherwise, we risk allowing concurrent access of the `subscriptions` array/set, which can lead to a crash.

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
Iron-Ham: Absolutely, if I mark its functions `mutating`.

Iron-Ham:
Special thank you to @AnthonyMDev for hopping on a call to chat about the Pager.

To summarize the call:

In this PR: 
- We will update the completion `Error` types to be `PaginationError`, to signify to the caller that the errors that you get here will only be `PaginationError`s thrown by the pager, not network errors. [f1b6c34](https://github.com/apollographql/apollo-ios-dev/pull/115/commits/f1b6c342ce65c2fa90f26552dee6941e303e895c)
- We will update the exposed `_subject` to instead be a `publisher`, which will prevent callers from sending new values into the Pager. [f92b550](https://github.com/apollographql/apollo-ios-dev/pull/115/commits/f92b550fc2b802f314866c2d9b4696850f243ff0)

In follow-up PRs:
- Examine the usage of `@MainActor`, the instantiation of `GraphQLQueryWatcher` to respond on the `main` `DispatchQueue` by default
- Rename `GraphQLQueryPager.Actor` to `GraphQLQueryPagerAsync` (or similar), and make it public
- Write an async type-erased `AnyGraphQLQueryPager` 
- Clean up the API within the existing `GraphQLQueryPager` type.

apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
```diff
@@ -0,0 +1,453 @@
...
+  /// - Parameters:
+  ///   - client: Apollo Client
+  ///   - initialQuery: The initial query that is being watched
+  ///   - extractPageInfo: The `PageInfo` derived from `PageExtractionData`
+  ///   - nextPageResolver: The resolver that can derive the query for loading more. This can be a different query than the `initialQuery`.
+  ///   - onError: The callback when there is an error.
+  public init<P: PaginationInfo>(
+    client: ApolloClientProtocol,
+    initialQuery: InitialQuery,
+    watcherDispatchQueue: DispatchQueue = .global(qos: .background),
```
Iron-Ham, line 75: cc: @AnthonyMDev I went back and forth on whether this should be `background` or `default`. Do you have any preferences?

AnthonyMDev: I think it should be `.default` personally.

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
```diff
@@ -5,38 +5,40 @@ import Combine
 /// Type-erases a query pager, transforming data from a generic type to a specific type, often a view model or array of view models.
 public class AnyGraphQLQueryPager<Model> {
   public typealias Output = Result<(Model, UpdateSource), Error>
-  public var canLoadNext: Bool { pager.canLoadNext }
+  private let _subject: CurrentValueSubject<Output?, Never> = .init(nil)
+  public var publisher: AnyPublisher<Output, Never> { _subject.compactMap { $0 }.eraseToAnyPublisher() }
```
AnthonyMDev, line 9: This is perfect :)

AnthonyMDev: Can we add inline documentation to the `publisher` property to indicate that this is the intended access point for using the pager as a `Combine` stream?

apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
Iron-Ham: Easy fix. Can update it shortly.

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
```diff
@@ -54,85 +56,84 @@ public class AnyGraphQLQueryPager<Model> {
...
+    pager.loadPrevious(cachePolicy: cachePolicy, completion: completion)
+  }
+
+  /// Loads all pages.
+  /// - Parameters:
+  ///   - fetchFromInitialPage: Pass true to begin loading from the initial page; otherwise pass false.  Defaults to `true`.  **NOTE**: Loading all pages with this value set to `false` requires that the initial page has already been loaded previously.
+  ///   - completion: A completion block that will always trigger after the execution of this operation. Passes an optional error, of type `PaginationError`, if there was an internal error related to pagination. Does not surface network errors. Defaults to `nil`.
+  public func loadAll(
+    fetchFromInitialPage: Bool = true,
+    completion: ((Error?) -> Void)? = nil
```
AnthonyMDev, line 112: Weren't we going to change the completion blocks to specifically return a `PaginationError`?

AnthonyMDev (approved):
Couple of minor things, but this is looking good!

AnthonyMDev (approved):
Couple of minor things, but this is looking good!

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
Iron-Ham: I'm surprised this didn't change, as I have it set that way in the internal GitHub version of this. The difficulties of maintaining this code in two places, I suppose 😅

apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
Iron-Ham: cc: @AnthonyMDev I've partially reverted this such that the default is `main`, but can be configured by the user. 

I noticed some unusual behaviors that can be triggered on iOS/iPadOS by:
- Having an ongoing request
- Backgrounding the application
- Using a non-`main` callback queue.

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
Pice12345: Ice

apollo-ios-pagination/Sources/ApolloPagination/AnyGraphQLPager.swift
Pice12345: > Weren't we going to change the completion blocks to specifically return a `PaginationError`?

Karliz24:
> ### What are you trying to accomplish?
> 
> 
> 
> 1. I want consumers of the `GraphQLPager` to be able to load all pages at once, should they desire to do so. 
> 
> 2. In a quest to deliver bidirectional pagination, we must first solidify support for reverse pagination. 
> 
> 
> 
> closes https://github.com/apollographql/apollo-ios/issues/3265
> 
> 
> 
> ### What approach did you choose and why?
> 
> 
> 
> **LoadAll Support**: 
> 
> First, I modified the `fetch` function to follow `async`/`await` syntax. That's so that we can `await` its return prior to beginning to loop over `loadMore`. 
> 
> 
> 
> Then, I implemented the function using a while loop.
> 
> 
> 
> Finally, I made additive changes to the `MockGraphQLServer` to support subsequent loads of this nature. 
> 
> 
> 
> **Reverse Pagination**:
> 
> Leaning hard into differentiating reverse and forward pagination from one another. 
> 
> If we are to support _both_ types of pagination at once, then they must be treated as being separate from one another: A page can be a previous page _or_ it can be a next page. It cannot be both. 
> 
> 
> 
> **BIdirectional Pagination**
> 
> Once reverse pagination was solidifed, this was a breeze to implement. It basically amounted to defining the pagination struct. 
> 
> 
> 
> ### Anything you want to highlight for special attention from reviewers?
> 
> 
> 
> Note that the output for the pagination controller is now a struct. This allows us to make changes to the output without breaking previous versions. I've removed logic around `Task` and `Continuation` management.
> 
> 
> 
> ### If something goes wrong, what are the mitigation strategies?
> 
> 
> 
> - **Rollback** - This change can only be disabled by reverting this pull request or commit.
> 
> 
> 
> ______
> 
> 
> 
> 
> 
> ### Copilot Summary
> 
> 
> 
> This pull request primarily introduces improvements to the Apollo GraphQL testing suite and pagination functionality. The most significant changes include the addition of a new `BidirectionalPaginationTests` class for testing bidirectional pagination, modifications to the `MockGraphQLServer` class to enhance testing capabilities, and updates to the `AnyGraphQLQueryPagerTests` and `ForwardPaginationTests` classes to accommodate new pagination functionalities.
> 
> 
> 
> Enhancements to Apollo GraphQL testing:
> 
> 
> 
> * <a href="diffhunk://#diff-9182276fe42a90e862b08125a7a6e055e60bb3b510e9840c92efe9284634fe5cR79-R88">`Tests/ApolloInternalTestHelpers/MockGraphQLServer.swift`</a>: Added a private subscript to the `MockGraphQLServer` class to handle request expectations for a given operation type. Also, a new `expect` method was added to handle request expectations for a specific operation. <a href="diffhunk://#diff-9182276fe42a90e862b08125a7a6e055e60bb3b510e9840c92efe9284634fe5cR79-R88">[1]</a> <a href="diffhunk://#diff-9182276fe42a90e862b08125a7a6e055e60bb3b510e9840c92efe9284634fe5cR100-R114">[2]</a>
> 
> 
> 
> Improvements to Apollo GraphQL Pagination:
> 
> 
> 
> * <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L58-R58">`Tests/ApolloPaginationTests/AnyGraphQLQueryPagerTests.swift`</a>: Modified the `eraseToAnyPager` method to include an additional parameter. Also, added a new `test_loadAll` method to test the loading of all pages. Several other minor changes were made to accommodate updates to the pagination functionality. <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L58-R58">[1]</a> <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7R92-R106">[2]</a> <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L103-R125">[3]</a> <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L129-R145">[4]</a>
> 
> * <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L50-R50">`Tests/ApolloPaginationTests/ForwardPaginationTests.swift`</a>: Updated several methods to reflect changes in the pagination functionality. <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L50-R50">[1]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L60-R64">[2]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L75-R96">[3]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L111-R113">[4]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L125-R127">[5]</a>
> 
> * <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L28-R28">`Tests/ApolloPaginationTests/ConcurrencyTest.swift`</a>: Updated several methods to reflect changes in the pagination functionality. <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L28-R28">[1]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L51-R51">[2]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L78-R82">[3]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L94-R98">[4]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L111-R111">[5]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L127-R134">[6]</a>
> 
> 
> 
> Addition of new test class:
> 
> 
> 
> * <a href="diffhunk://#diff-eb8712982f5deb2896d926c2943479d019c26243bcdc5d05ef810e4b84ad499fR1-R197">`Tests/ApolloPaginationTests/BidirectionalPaginationTests.swift`</a>: Added a new `BidirectionalPaginationTests` class to test bidirectional pagination functionality. This class includes several methods to test fetching multiple pages, loading all pages, and more.

Karliz24:
> > ### What are you trying to accomplish?
> > 
> > 
> > 
> > 1. I want consumers of the `GraphQLPager` to be able to load all pages at once, should they desire to do so. 
> > 
> > 2. In a quest to deliver bidirectional pagination, we must first solidify support for reverse pagination. 
> > 
> > 
> > 
> > closes https://github.com/apollographql/apollo-ios/issues/3265
> > 
> > 
> > 
> > ### What approach did you choose and why?
> > 
> > 
> > 
> > **LoadAll Support**: 
> > 
> > First, I modified the `fetch` function to follow `async`/`await` syntax. That's so that we can `await` its return prior to beginning to loop over `loadMore`. 
> > 
> > 
> > 
> > Then, I implemented the function using a while loop.
> > 
> > 
> > 
> > Finally, I made additive changes to the `MockGraphQLServer` to support subsequent loads of this nature. 
> > 
> > 
> > 
> > **Reverse Pagination**:
> > 
> > Leaning hard into differentiating reverse and forward pagination from one another. 
> > 
> > If we are to support _both_ types of pagination at once, then they must be treated as being separate from one another: A page can be a previous page _or_ it can be a next page. It cannot be both. 
> > 
> > 
> > 
> > **BIdirectional Pagination**
> > 
> > Once reverse pagination was solidifed, this was a breeze to implement. It basically amounted to defining the pagination struct. 
> > 
> > 
> > 
> > ### Anything you want to highlight for special attention from reviewers?
> > 
> > 
> > 
> > Note that the output for the pagination controller is now a struct. This allows us to make changes to the output without breaking previous versions. I've removed logic around `Task` and `Continuation` management.
> > 
> > 
> > 
> > ### If something goes wrong, what are the mitigation strategies?
> > 
> > 
> > 
> > - **Rollback** - This change can only be disabled by reverting this pull request or commit.
> > 
> > 
> > 
> > ______
> > 
> > 
> > 
> > 
> > 
> > ### Copilot Summary
> > 
> > 
> > 
> > This pull request primarily introduces improvements to the Apollo GraphQL testing suite and pagination functionality. The most significant changes include the addition of a new `BidirectionalPaginationTests` class for testing bidirectional pagination, modifications to the `MockGraphQLServer` class to enhance testing capabilities, and updates to the `AnyGraphQLQueryPagerTests` and `ForwardPaginationTests` classes to accommodate new pagination functionalities.
> > 
> > 
> > 
> > Enhancements to Apollo GraphQL testing:
> > 
> > 
> > 
> > * <a href="diffhunk://#diff-9182276fe42a90e862b08125a7a6e055e60bb3b510e9840c92efe9284634fe5cR79-R88">`Tests/ApolloInternalTestHelpers/MockGraphQLServer.swift`</a>: Added a private subscript to the `MockGraphQLServer` class to handle request expectations for a given operation type. Also, a new `expect` method was added to handle request expectations for a specific operation. <a href="diffhunk://#diff-9182276fe42a90e862b08125a7a6e055e60bb3b510e9840c92efe9284634fe5cR79-R88">[1]</a> <a href="diffhunk://#diff-9182276fe42a90e862b08125a7a6e055e60bb3b510e9840c92efe9284634fe5cR100-R114">[2]</a>
> > 
> > 
> > 
> > Improvements to Apollo GraphQL Pagination:
> > 
> > 
> > 
> > * <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L58-R58">`Tests/ApolloPaginationTests/AnyGraphQLQueryPagerTests.swift`</a>: Modified the `eraseToAnyPager` method to include an additional parameter. Also, added a new `test_loadAll` method to test the loading of all pages. Several other minor changes were made to accommodate updates to the pagination functionality. <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L58-R58">[1]</a> <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7R92-R106">[2]</a> <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L103-R125">[3]</a> <a href="diffhunk://#diff-93f14fc6d90ab25e52d223b3ef44de909097c352cabd917c877ec3258e3428e7L129-R145">[4]</a>
> > 
> > * <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L50-R50">`Tests/ApolloPaginationTests/ForwardPaginationTests.swift`</a>: Updated several methods to reflect changes in the pagination functionality. <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L50-R50">[1]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L60-R64">[2]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L75-R96">[3]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L111-R113">[4]</a> <a href="diffhunk://#diff-9ffaa1164673a2d19816d967837da49c0d7feb8b39482594294d7a267bb652c8L125-R127">[5]</a>
> > 
> > * <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L28-R28">`Tests/ApolloPaginationTests/ConcurrencyTest.swift`</a>: Updated several methods to reflect changes in the pagination functionality. <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L28-R28">[1]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L51-R51">[2]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L78-R82">[3]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L94-R98">[4]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L111-R111">[5]</a> <a href="diffhunk://#diff-10b2258b9cc67b10204e621bc4fa1078c3ea1e2e50679acdf016244bc0c22cd6L127-R134">[6]</a>
> > 
> > 
> > 
> > Addition of new test class:
> > 
> > 
> > 
> > * <a href="diffhunk://#diff-eb8712982f5deb2896d926c2943479d019c26243bcdc5d05ef810e4b84ad499fR1-R197">`Tests/ApolloPaginationTests/BidirectionalPaginationTests.swift`</a>: Added a new `BidirectionalPaginationTests` class to test bidirectional pagination functionality. This class includes several methods to test fetching multiple pages, loading all pages, and more.
> 
>

Heyokaskey:
Just be happy

4304417:
134479

PR #126, opened 2023-11-07
refactor: Multipart parsers

calvincestari:
This refactors the multipart response parsers so that each can be built to focus on just their parser data specification logic and common shared logic to process the general multipart response format.
* `MultipartResponseParsingInterceptor` now handles the mechanics of the multipart message chunk parsing.
* `MultipartResponseSpecificationParser` now just requires each specification parser to handle the logic for parsing each data line for a single specification.

_This only includes error based-tests for `MultipartResponseDeferParser` because the executor is not complete. Only once the executor is able to handle deferred fragments will the parsing based-tests be able to succeed._

### DIFF
--- a/apollo-ios/Sources/Apollo/MultipartResponseParsingInterceptor.swift
+++ b/apollo-ios/Sources/Apollo/MultipartResponseParsingInterceptor.swift
@@ -126,7 +126,13 @@ protocol MultipartResponseSpecificationParser {
   /// in an HTTP response.
   static var protocolSpec: String { get }
 
-  /// Function called to process each data line of the chunked response.
+  /// Called to process each chunk in a multipart response.
+  ///
+  /// The return value is a `Result` type that indicates whether the chunk was successfully parsed
+  /// or not. It is possible to return `.success` with a `nil` data value. This should only happen
+  /// when the chunk was successfully parsed but there is no action to take on the message, such as
+  /// a heartbeat message. Successful results with a `nil` data value will not be returned to the
+  /// user.
   static func parse(_ chunk: String) -> Result<Data?, Error>
 }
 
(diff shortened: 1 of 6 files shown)


apollo-ios/Sources/Apollo/MultipartResponseParsingInterceptor.swift
```diff
@@ -111,11 +126,14 @@ protocol MultipartResponseSpecificationParser {
...
 
-  /// Function that will be called to process the response data.
-  static func parse(
-    data: Data,
-    boundary: String,
-    dataHandler: ((Data) -> Void),
-    errorHandler: ((Error) -> Void)
-  )
+  /// Function called to process each data line of the chunked response.
+  static func parse(_ chunk: String) -> Result<Data?, Error>
```
calvincestari, line 130: It may seem odd for a successful result to have an optional `Data` value; this is because there are messages that can be parsed successfully, such as the HTTP subscription heartbeat message, that do not require moving to the next interceptor. In this case the multipart response parsing interceptor will swallow the success but nothing will be returned to the user.

I considered the following options, with the `Result<Data?, Error>` type seeming the best:
1. Passing in closures for success and failure to the `parse(chunk:)` protocol method but it's unnecessary to be creating closures for each message being processed.
2. Moving the logic into an extension on the `MultipartResponseSpecificationParser` protocol but that's not possible because the successful `proceedAsync` call needs to include the interceptor which the protocol won't have access to and the extension cannot store.
3. `Result` is fitting because the 'result' of parsing is either successful with potential data to pass on, or failure with an error.

AnthonyMDev: I think this approach is fine. We should just add documentation to this `parse(_:)` function that indicates what would be an appropriate time to return a `.success(nil)` and how that will be handled (ie. not throwing an error, but not proceeding down the chain).

In 2.0, we should probably find a way to allow the chain to continue in this case. In case the user has custom interceptors that still log something or react to a response with no data.

AnthonyMDev (approved):
Looks good!

AnthonyMDev (commented):
LGTM

PR #127, opened 2023-11-07
refactor: HTTP `Accept` header

calvincestari:
This change is to standardize on the HTTP `accept` header being sent by Apollo iOS and Kotlin clients.

The difference is that the `multipart/mixed` and `application/json` accept types will be included with every query and mutation operation to simplify the request generation and since the response parsing can handle both. This is related to https://github.com/apollographql/apollo-ios/issues/3277.


AnthonyMDev (approved):
LGTM

PR #128, opened 2023-11-08
fix Array JSONEncodable

arnauddorgans:
Avoid using fatal error on `JSONEncodable` instead of conforming protocol with where clause


AnthonyMDev:
Hmmm... I wanted to do this ages ago, and there was some reason why I couldn't. I think it was causing some build errors.

We're going to need to test this manually on iOS versions pre 14.4. I'm wondering if the issues I saw before had to do with the conversion of the nested `AnyHashable` types inside of the array (which doesn't work properly on iOS < 14.4).

Thanks for the PR. Give me some time to look into this a bit further before we pull it in.

AndersonMartins1:
Do you need help with something?

calvincestari (approved):

PR #131, opened 2023-11-08
Adding optimizations to ci tests

BobaFetters:
-Adding checks for changes to ios and codegen code
-Updating jobs to only run if changes were made in relevant areas of the code


AnthonyMDev (approved):
Very nice improvement! Thanks for this!

BobaFetters:
I think it's possible there may be even better ways to handle this where maybe the matrix based jobs can just not run at all instead of starting and skipping steps and then ending, may be something to look into for the future.

calvincestari (approved):
LGTM!

PR #132, opened 2023-11-09
InstallCLI Plugin Updates

BobaFetters:
- Updating the InstallCLI plugin to download the proper CLI version from GitHub based on the package version being used.
- Added a `download-cli.sh` script which handles most of the work for the plugin now

@AnthonyMDev @calvincestari I have tested this in an SPM based project and an Xcode Project and everything appears to be working but would appreciate a little extra scrutiny just to make sure I didn't overlook a possible use case. There is a `cli-plugin-update` branch on `apollo-ios` that can be targeted for the SPM dependency to test this.


apollo-ios/Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7
+// swift-tools-version:5.9
```
calvincestari, line 1: This will require users to be on Xcode 15; which is fine as long as we're comfortable with that requirement. 15.1 should be available soon so there has been some time for adoption since it was first released. I don't know if there is a reason someone would want to stay on 14.x though?

apollo-ios/Package.swift
```diff
@@ -64,6 +64,7 @@ let package = Package(
           description: "Installs the Apollo iOS Command line interface."),
         permissions: [
           .writeToPackageDirectory(reason: "Creates a symbolic link to the CLI executable in your project directory."),
```
calvincestari, line 66: Is this permission still relevant? It looks like we download the binary directly to the project folder now instead of creating a symbolic link.

On second though - we may still need the permission to write to the directory but the reason text should probably be updated then.

calvincestari (commented):
I think this looks good, just a few comments for us to mull over.

apollo-ios/Package.swift
BobaFetters: Yea I think its reasonable at a certain point to expect Xcode 15 especially since technically users could stay on a previous version.

apollo-ios/Package.swift
BobaFetters: Ah yea good call out, I can get that updated.

apollo-ios/Package.swift
BobaFetters: With this in mind, I imagine we would want to this update to be in a minor release that the cutoff for Xcode 15 is not in a patch release.

calvincestari (approved):

AnthonyMDev (approved):
This is awesome! Thanks @BobaFetters!

AnthonyMDev (approved):

BobaFetters:
Going to hold off on merging this in until we are ready for the 1.8 release since its going to require using Xcode 15.

PR #138, opened 2023-11-15
Update docs for input objects casing strategy

BobaFetters:


docs/source/code-generation/codegen-configuration.mdx
```diff
@@ -741,7 +743,8 @@ Below is an example that illustrates an `apollo-codegen-config.json` where every
     "cocoapodsCompatibleImportStatements" : true,
     "conversionStrategies" : {
       "enumCases" : "none",
-      "fieldAccessors" : "camelCase"
+      "fieldAccessors" : "camelCase",
+			"inputObjects": "camelCase"
```
calvincestari, line 747:
```suggestion
      "inputObjects": "camelCase"
```

Looks like the spacing here was tabs vs. spaces.

calvincestari (commented):
This is fine in that it demonstrates how to use it in the sample code but I wonder if we're reaching the point of needing long-form content for the output options? Right now we just link out to the API reference docs for each output option.

@AnthonyMDev is there anything else you wanted for this documentation?

BobaFetters:
Yea we could probably add more depth to the output options, can add an issue for that if everyone agrees we want to do that at some point.

calvincestari (approved):

PR #139, opened 2023-11-16
Adding GraphQLResult conversion extension

BobaFetters:
-Adding extension for `GraphQLResult` to convert response data into a json dictionary

Closes apollographql/apollo-ios#3196


apollo-ios/Sources/Apollo/GraphQLResult.swift
```diff
@@ -47,3 +47,28 @@ extension GraphQLResult: Equatable where Data: Equatable {
...
+  public func asDictionary() -> [String: Any] {
+    var dict: [String: Any] = [:]
+    if let data { dict["data"] = convert(value: data.__data) }
+    if let errors { dict["errors"] = errors.map { $0.asDictionary() } }
+    if let extensions { dict["extensions"] = extensions }
+    return dict
+  }
+  
+  private func convert(value: Any) -> Any {
+      var val: Any = value
```
BobaFetters, line 61: @AnthonyMDev @calvincestari Started with essentially what was in the issue with a few adjustments to fix errors. But I'm wondering if it would be better to have `val` be optional and then do a guard/throws before attempting to use it.

Similarly for the 2 `asDictionary` functions, similar for the `asDictionary` functions if there should maybe be some minimum requirements like if the result dictionary is empty, should we throw an error vs returning an empty dictionary to make it more clear to the user that its empty, or is it reasonable that they could have an empty result?

apollo-ios/Sources/Apollo/GraphQLError.swift
```diff
@@ -86,3 +86,14 @@ extension GraphQLError: LocalizedError {
     return self.description
   }
 }
+
+extension GraphQLError {
+  public func asDictionary() -> [String: Any] {
```
calvincestari, line 91: Could we make this an `internal` function instead? Unless there is value in letting users turn their `GraphQLError` instances into JSON, but they get that already through `GraphQLResult.asDictionary()`?

apollo-ios/Sources/Apollo/GraphQLResult.swift
calvincestari: IMO there is only value in optional if we're going to allow/return nulls. If we're simply going to `guard` and `throw` there is little benefit in a `private` function. Yes it's more verbose to have to check `data`, `errors`, `extensions` but it's clearer what you're going to get out of `convert`?

apollo-ios/Sources/Apollo/GraphQLResult.swift
```diff
@@ -47,3 +47,28 @@ extension GraphQLResult: Equatable where Data: Equatable {
 }
 
 extension GraphQLResult: Hashable where Data: Hashable {}
+
+extension GraphQLResult {
+  public func asDictionary() -> [String: Any] {
```
calvincestari, line 52: I wonder if `asDictionary` is clear enough. Everyone that's asked for this has done it in reference to wanting "JSON" output, would 'asJSON' be more intuitive and result in less follow-up questions?

apollo-ios/Sources/Apollo/GraphQLError.swift
BobaFetters: Yea good point, probably not a need to expose this publicly necessarily

apollo-ios/Sources/Apollo/GraphQLResult.swift
BobaFetters: Hmmm maybe even more verbose and say `asJSONDictionary`? I would think JSON would be clear enough but I guess you never know.

apollo-ios/Sources/Apollo/GraphQLResult.swift
AnthonyMDev: Yeah I agree with @calvincestari here. I also think it's okay to return an empty dictionary. If the result data were to bee empty, that's what we would want.

apollo-ios/Sources/Apollo/GraphQLResult.swift
AnthonyMDev: I agree. It should have `JSON` in the name. I'm wondering if we should actually just make `GraphQLResult` and `GraphQLError` conform to `JSONEncodable` and then make this a `var _jsonValue`. But that would return a `[String: AnyHashable]`. @jimisaacs would that be alright for you, or do you prefer it to return a `[String: Any]`?

jimisaacs: Just as long as it's simple to serialize to Foundation.Data, and marshal back into generated types, I'm good. My original public interface returned Foundation.Data.

jimisaacs: I personally think just leaving it as [String: Any] and adding JSON to the name is probably fine. Enforcement of AnyHashable feels unnecessary since encoders all have to check every value and throw anyway.

apollo-ios/Sources/Apollo/GraphQLResult.swift
```diff
@@ -47,3 +47,28 @@ extension GraphQLResult: Equatable where Data: Equatable {
 }
 
 extension GraphQLResult: Hashable where Data: Hashable {}
+
+extension GraphQLResult {
+  public func asJSONDictionary() -> [String: Any] {
```
calvincestari, line 52: Just missing documentation for this new function and then I think it's good to go.

calvincestari (approved):

AnthonyMDev (approved):

PR #152, opened 2023-11-20
Codegen EntitySelectionTree + MergedSelections Refactor + Perf Improvments

AnthonyMDev:
This PR does a number of things to refactor the Codegen engine in preparation for the [disabling of merged fragment fields](https://github.com/apollographql/apollo-ios/issues/2560) and improve the performance of the Codegen engine.

- `EntitySelectionTree` now computes the merged fields from merged fragments as needed when the merged fields are calculated, rather than when the fragment is initially merged in.
  - This reduces memory consumption by not storing duplicated selections.
  - This likely improves execution time as well
  - Thanks @BobaFetters for the idea to make this algorithmic improvement!
- `RootFieldEntityStorage` renamed to `DefinitionEntityStorage` and attached to definition.
  -  The new `EntitySelectionTree` algorithm requires the entity storage at the time of calculation merged fields (since we are no longer merging the fragments and creating new entities when merging the selections in)
- `IR.Definition` refactored from an `enum` to a `protocol`.
  - This improves usage flexibility.
- Refactor `IR.SelectionSet`
  - Improve initializers and clean up `RootFieldBuilder` logic.
  - Move `merged` selections to new `ComputedSelectionSet`
    - The merged selections require a lot of memory since they hold a lot of duplicate instances of the selections from their direct instance. Making them a `lazy var` on `SelectionSet` was convenient for use, but it meant they were retained for the entire time the `SelectionSet` was in memory. Since `IR.Operation` and `IR.NamedFragment` retain their `rootField` these were being retained for the entire execution of codegen.
    - `ComputedSelectionSet`, which now computes and stores the merged selections, is created as needed in the `SelectionSetTemplate` using the new `ComputedSelectionSet.Builder`. These computed selection sets can be used when needed to render the selection set templates and then be released. This should improve memory consumption greatly.
- Changes the generated model format for using referenced fragments (precursor to hoisted types).
  - Though we haven't implemented intelligent hoisted types yet, there are times when a field accessor for a merged field points to another generated `SelectionSet`, rather than generating its own child selection set with the same exact shape.
  - The determination of whether a selection set can use a referenced fragment depends on the merged selections, but we no longer want to compute these and retain them on the `IR.SelectionSet`, which stays in memory for the entirety of execution.
  - Since we don't compute the merged selections until the generation of that actual selection set, we don't know if it will be a referenced fragment at the time of the generation of field accessors, selection set initializer parameters, etc.
  - To work around this, we are generating those references using the name of the selection set as it would be if it were to be rendered as a child of the current selection set. When we compute the merged selections, if we recognize that we should use a reference to another selection set, we then create a `typealias` pointing to the correct fully qualified name of the referenced fragment instead of generating the selection set.
- Create new `IRTestWrapper` classes in `ApolloCodegenInternalTestHelpers`.
  - Because a lot of our tests are using the test subscripts to access and verify the built selections using both the direct and merged selections, they would have required a lot of changes and messy code to build the `ComputedSelectionSet` for each nested child entity.
  - These `IRTestWrapper` instances wrap the existing `IR` objects. The subscripts for testing are now provided by the `IRTestWrapper`, which handle the computation of merged fields internally _as they are accessed_.
  - This means the changes to the unit test code is minimal, while still maintaining the intentions and functionality of the existing tests.

TODO:

- [x] Finish fixing the call sites in unit tests as necessary
- [x] Fix type conflict validation (this has to be done in `SelectionSetTemplate` when the `ComputedSelectionSet` has been built.)
- [x] Improve `IRTestWrapper` to store the `ComputedSelectionSet` for reuse during a single test run.
- [x] Finish documentation updates
- [ ] Run performance tests for metrics

After this PR is completed, the work to create the [configuration options for disabling merged fragment fields](https://github.com/apollographql/apollo-ios/issues/2560) will be unlocked.


pm-dev:
Looking forward to this one

AnthonyMDev:
@pm-dev. This is the foundational refactoring needed to enable disabling of fragment field merging. That will finally get us to a point where the codegen will work on your project!

calvincestari (approved):
Great improvements here, thank you!

PR #153, opened 2023-11-22
tests: Add merged selection deferred test case

calvincestari:


apollo-ios/Sources/Apollo/GraphQLResult.swift
```diff
@@ -63,7 +63,7 @@ extension GraphQLResult {
   
   private func convert(value: Any) -> Any {
       var val: Any = value
-      if let value = value as? ApolloAPI.DataDict {
+      if let value = value as? DataDict {
```
calvincestari, line 66: @BobaFetters I had to make this change to get the cocoapods codegen test configuration project to pass. The change makes sense but how have other PRs been building, including the one that included this change originally?

calvincestari: It's because [the Codegen Test Configurations job was skipped](https://github.com/apollographql/apollo-ios-dev/actions/runs/6948892168/job/18905863839). We'll need to revise the logic that turns that on/off.

AnthonyMDev (approved):
This looks right to me. The deferred selection should be a `mergedSelection` there. It's being merged in from the parent selection set.

PR #154, opened 2023-11-28
Adjusting codegen config test usage

BobaFetters:
Changing codegen config tests to run with every PR.

Closes apollographql/apollo-ios#3295


calvincestari (approved):
Thanks @BobaFetters!

PR #155, opened 2023-11-29
defer: Incremental response parsing + field collection + execution

calvincestari:
~This PR implements the infrastructure for the multipart parser to identify and collect the deferred fragments that are returned in incremental responses, and pass them forward to the GraphQL executor for field collection. Field collection then operates on the logic that if the deferred selection label (provided by the user) is in the list of fulfilled deferred fragments then all fields for that deferred selection must be collected and executed on.~

~Notes:~
* ~This PR **does not** include the actual collection of the deferred fragment identifiers in the multipart parser; that will come in a subsequent PR. The logic for the executor to use the fulfilled fragments is included here and fully tested.~

I'm updating this solution to one that executes on the incremental responses and merges them together for the result that gets passed to the user.


Tests/ApolloInternalTestHelpers/MockOperation.swift
```diff
@@ -84,7 +84,7 @@ open class MockFragment: MockSelectionSet, Fragment {
   open class var fragmentDefinition: StaticString { "" }
 }
 
-open class MockTypeCase: MockSelectionSet, InlineFragment {
+open class MockTypeCase: MockSelectionSet, InlineFragment, Deferrable {
```
calvincestari, line 87: I might change the composition of `InlineFragment` to include `Deferrable` in a later PR. `Fragment` conforms to `Deferrable` so it makes sense that `InlineFragment` should too. It would clean up the codegen and generated models a bit too.

apollo-ios/Sources/Apollo/ApolloStore.swift
```diff
@@ -232,6 +232,7 @@ public class ApolloStore {
       return try executor.execute(
         selectionSet: type,
         on: object,
+        expecting: .all,
```
calvincestari, line 235: `all` seems correct here based on what we spoke about yesterday @AnthonyMDev, but please double-check. It'll be reviewed when the cache work is done too.

apollo-ios/Sources/Apollo/SelectionSet+JSONInitializer.swift
```diff
@@ -16,6 +16,7 @@ extension RootSelectionSet {
   ///                the given JSON response data.
   public init(
     data: JSONObject,
+    fulfilledFragments: FulfilledFragments = .labels([]),
```
calvincestari, line 19: We may want to leave this parameter out of here depending on whether the defer preview release ships with support for selection set initializers or not. If it doesn't then it might be confusing to users of what this actually does.

apollo-ios/Sources/ApolloAPI/Selection.swift
```diff
@@ -6,7 +6,7 @@ public enum Selection {
   /// An inline fragment with a child selection set nested in a parent selection set.
   case inlineFragment(any InlineFragment.Type)
   /// A fragment spread or inline fragment marked with the `@defer` directive.
-  case deferred(if: Condition? = nil, any Deferrable.Type, label: String?)
+  case deferred(if: Condition? = nil, any Deferrable.Type, label: String)
```
calvincestari, line 9: A label on the defer directive is required by the JS frontend validation, this enforces that in the generated models. It's also required for fulfilled fragments to work.

calvincestari:
This was closed because of a change to parse incremental selection sets instead of parsing on the root selection set with the fulfilled fragments specified to the executor.

PR #156, opened 2023-11-29
[codegen] add additionalImportedModulesNames to apollo codegen config

hemel:
The purpose of this change is to give codegen the capability to add custom import statements to generated operations.
This change would enable the consumer to share fragments across generated packages.

Addresses this thread:

https://github.com/apollographql/apollo-ios/issues/3284


calvincestari:
Thanks for the PR @hemel but I don't think this correctly solves the issue in https://github.com/apollographql/apollo-ios/issues/3284.

The problem description is three modules (A, B, C) and only two (A, B) need to share the fragments (D); _C should not import D_. Your solution essentially replicates putting the fragments into the shared schema module which all operation files already import.

I'm going to close this PR but if you want to keep iterating on it please go ahead. Manipulating the templates is the easy part but finding an elegant configuration model is the difficult part.

jimisaacs:
@hemel I'm confused, these don't look like all the changes we landed on? Can you share some context?

jimisaacs:
@hemel nevermind, was confused found https://github.com/apollographql/apollo-ios-dev/pull/158

PR #158, opened 2023-12-01
[codegen] add additionalImportedModulesNames to apollo codegen config

hemel:
The purpose of this change is to give codegen the capability to add custom import statements to generated operations.
This change would enable the consumer to share fragments across generated packages.

Addresses this thread:

https://github.com/apollographql/apollo-ios/issues/3284

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -191,8 +191,6 @@ jobs:
 
   run-codegen-test-configurations:
     runs-on: macos-13
-    needs: [changes]
-    if: ${{ needs.changes.outputs.codegen  == 'true' }}
     timeout-minutes: 20
     name: Codegen Test Configurations - macOS
     steps:
--- a/Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/FragmentFileGeneratorTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/FragmentFileGeneratorTests.swift
@@ -50,6 +50,7 @@ class FragmentFileGeneratorTests: XCTestCase {
     
     subject = FragmentFileGenerator(
       irFragment: irFragment,
+      patternMatchedOutputOptions: nil,
       config: ApolloCodegen.ConfigurationContext(config: ApolloCodegenConfiguration.mock())
     )
   }
--- a/Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/OperationFileGeneratorTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGeneration/FileGenerators/OperationFileGeneratorTests.swift
@@ -46,7 +46,7 @@ class OperationFileGeneratorTests: XCTestCase {
 
     let config = ApolloCodegen.ConfigurationContext(config: ApolloCodegenConfiguration.mock())
     
-    subject = OperationFileGenerator(irOperation: irOperation, operationIdentifier: nil, config: config)
+    subject = OperationFileGenerator(irOperation: irOperation, operationIdentifier: nil, patternMatchedOutputOptions: nil, config: config)
   }
 
   // MARK: Property Tests
--- a/Tests/ApolloCodegenTests/CodeGeneration/Templates/FragmentTemplateTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGeneration/Templates/FragmentTemplateTests.swift
@@ -53,6 +53,7 @@ class FragmentTemplateTests: XCTestCase {
     fragment = await ir.build(fragment: fragmentDefinition)
     subject = FragmentTemplate(
       fragment: fragment,
+      patternMatchedOutputOptions: nil,
       config: ApolloCodegen.ConfigurationContext(config: config)
     )
   }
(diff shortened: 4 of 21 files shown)


AnthonyMDev:
Thanks for creating another PR here! This is an interesting approach to this. I think there are other options that we may want to use pattern matching for in the future as well. I'd like to do a design review with the team on this and decide if we want to make any tweaks to this before shipping it. I'll add this to the agenda for our next team meeting.

AnthonyMDev:
@hemel. How would you feel about us adding an `@import("MyModuleName")` directive that you could add to your operation definitions in the individual GraphQL files instead?

We want to try to keep configuration that affects individual files out of the global configuration if possible. We've also considered allowing you to add a `.apollo-config` file to individual directories in your project that would override the `OutputOptions` for any files in that directory. Similar to how `.swift-format` files work, for example.

jimisaacs:
I realize this is @hemel's PR, but my 2¢ is that I'm not a fan of `.apollo-config`, as it reminds of the old days of `.htaccess` confusion, and the errors prone to figuring out config hierarchies via files systems. The only way I could see mitigating these errors is using a combination of the 2 approaches you proposes, i.e. instead of `@import("MyModuleName")`, it could be something like `@apollo-config("./.apollo-config")`

jimisaacs:
> We want to try to keep configuration that affects individual files out of the global configuration if possible.
I'd also be interesting in the reasoning of this.

AnthonyMDev:
> my 2¢ is that I'm not a fan of .apollo-config, as it reminds of the old days of .htaccess confusion, and the errors prone to figuring out config hierarchies via files systems.

That's fair. I think the `@import` directive is the best way to go anyways. That was just another suggestion to consider.

> We want to try to keep configuration that affects individual files out of the global configuration if possible.
> > I'd also be interested in the reasoning of this.

There are a few reasons for this. But complexity of the configuration is the primary one. Different types of configuration options are going to need to work differently, and this pattern matching approach won't work the same for all of them.

We already have the `selectionSetInitializers` option, which can selectively turn the feature on for different operations by name. But that wouldn't work with pattern matching as cleanly. What if two patterns matched the same file with differing values? This additional imports config is additive, so that's not really an issue (ie. it would add all imports that are pattern matched), but for a boolean config option, there is conflicts.

It also prevents local reasoning, forcing you to edit a single global conflict file if you want to fix imports for a single generated file in a single module. Which is why we feel `@import` is a better approach.

jimisaacs:
Thanks for the response @AnthonyMDev. I see you added this as an item in the ROADMAP. I'll leave to @hemel to response on the priority of the support.  imagine that if said priority is or becomes date driven we'll need to align on that.

hemel:
Hello, @AnthonyMDev. Happy New Year! Thank you for reviewing the PR. In terms of priority, we would need the @import capability on our end by the end of the month (January 2024). Is it feasible to accomplish this on your side?

AnthonyMDev:
It's not too difficult of an addition. The first step is adding the directive to our `graphql-js` wrapper. Then is will be recognized and exposed in the Swift code. If I get that part done, would you be interested in creating the PR to consume the directive and generate the import statements in the `ApolloCodegenLib`?

AnthonyMDev:
Going to close this PR, as we are moving forward with the other approach.

hemel:
That works @AnthonyMDev. Please let me know when the directive is ready and i can get started on creating the PR. Thank you again!

AnthonyMDev:
I've got #228 that adds the import directive merging into `main` in the next few minutes. You should be able to branch off of there and consume that directive to add the import statements!

hemel:
Hi @AnthonyMDev, I ran the codegen based on the changes you made. I am receiving a JS error - Unknown directive \"@import\"."] - please let me know if this is expected with the changes above. thank you.

AnthonyMDev:
You should not need to add the directive to your schema at all. And you shouldn't have to handle the new directive in JS.

The codegen engine should be including a schema extension that has the `@import(module:)` directive during codegen execution. But it should be stripped out of the GraphQL definitions included in the generated models, so your schema should not need to include it.

I'm not entirely clear on when you are seeing this issue yet. Is this happening during codegen, or when fetching your operations?

hemel:
I was running the build ApolloCodegen.Build function when encountering the issue. Once i redownloaded the schema, the directive was added, which resolved my issue.

AnthonyMDev:
Did you need to manually add the `@import(module:)` directive to your schema though? You should not have needed to do so.

It sounds like you are able to move forward on creating this PR, but if this part isn't working correctly, I want to make sure I resolve it.

hemel:
Hi @AnthonyMDev . I had a question about how the import directive should work for nested fragments.

When adding the directive to a fragment C in the case of 
`Query A { ...B }  Fragment B { ...C } Fragment C @import(module: "MyModule") { ...FragmentFromMyModule}`

The enclosing fragments and queries (A,B) will reference fragments from inside C, which would be included in the imported module. Do you expect the codegen to add the import statement automatically to the Fragment B and Query A, or should we explicitly add the import directive in A and B?

hemel:
> Did you need to manually add the `@import(module:)` directive to your schema though? You should not have needed to do so.
> 
> It sounds like you are able to move forward on creating this PR, but if this part isn't working correctly, I want to make sure I resolve it.

I did not have to manually add the directive. So think that part is working as expected.

PR #159, opened 2023-12-01
Make ApolloInterceptor id requirement read-only

SebastianOsinski:
As far as I understand, there is no real need for interceptors' ids to be mutable, at least not by the library itself. I decided to leave ids as `var` in all default interceptors though as they're randomly generated and maybe someone has a need to change them to something more descriptive.

### DIFF
--- a/apollo-ios/Sources/Apollo/ApolloInterceptor.swift
+++ b/apollo-ios/Sources/Apollo/ApolloInterceptor.swift
@@ -10,7 +10,7 @@ public protocol ApolloInterceptor {
   /// Each operation request has it's own interceptor request chain so the interceptors do not
   /// need to be uniquely identifiable between each and every request, only unique between the
   /// list of interceptors in a single request.
-  var id: String { get set }
+  var id: String { get }
   
   /// Called when this interceptor should do its work.
   ///


AnthonyMDev (approved):
@calvincestari Is there any reason why we need this to be writable in the protocol? I'm in agreement with @SebastianOsinski on this one.

calvincestari:
This change is correct, the ID does not need to be writable. I think the set was from an approach where I had the request chain setting the ID on initialization, and we simply didn't catch it in review. Changing the ID after initialization, as is possible now, would break the interceptor request chain because it wouldn't be able to find it.

I was going to get to this PR later in my day to ensure it isn't a breaking change in any other way.

calvincestari:
Thanks for this contribution @SebastianOsinski. I was hoping we could change the interceptor properties to be constants too but you're correct, there may be some projects relying on them being mutable. As long as the interceptor IDs are set before creating the request chain then everything should continue to work as expected.

I've just updated this PR with the latest changes from `main` and once all the tests pass again I'll get it merged.

calvincestari (approved):

calvincestari:
The gitleaks check seems to be holding this up; I'm going to go ahead and merge it anyways.

PR #180, opened 2023-12-04
chore(deps): update dependency rollup to v4

svc-secops:
This PR contains the following updates:

| Package | Change | Age | Adoption | Passing | Confidence |
|---|---|---|---|---|---|
| [rollup](https://rollupjs.org/) ([source](https://togithub.com/rollup/rollup)) | [`2.79.1` -> `4.6.1`](https://renovatebot.com/diffs/npm/rollup/2.79.1/4.6.1) | [age](https://docs.renovatebot.com/merge-confidence/) | [adoption](https://docs.renovatebot.com/merge-confidence/) | [passing](https://docs.renovatebot.com/merge-confidence/) | [confidence](https://docs.renovatebot.com/merge-confidence/) |

---

### Release Notes

<details>
<summary>rollup/rollup (rollup)</summary>

### [`v4.6.1`](https://togithub.com/rollup/rollup/blob/HEAD/CHANGELOG.md#461)

[Compare Source](https://togithub.com/rollup/rollup/compare/v4.6.0...v4.6.1)

*2023-11-30*

##### Bug Fixes

-   Resolve a situation where declaring the same `var` several times was considered a conflict (#&#8203;[https://github.com/rollup/rollup/issues/5276](https://togithub.com/rollup/rollup/issues/5276)s/5276))

##### Pull Requests

-   #&#8203;[https://github.com/rollup/rollup/pull/5275](https://togithub.com/rollup/rollup/pull/5275)l/5275): Add TNG as special sponsor ([@&#8203;lukastaegert](https://togithub.com/lukastaegert))
-   #&#8203;[https://github.com/rollup/rollup/pull/5276](https://togithub.com/rollup/rollup/pull/5276)l/5276): Allow to redeclare parameters multiple times in nested scopes ([@&#8203;lukastaegert](https://togithub.com/lukastaegert))

### [`v4.6.0`](https://togithub.com/rollup/rollup/blob/HEAD/CHANGELOG.md#460)

[Compare Source](https://togithub.com/rollup/rollup/compare/v4.5.2...v4.6.0)

*2023-11-26*

##### Features

-   Allow `this.addWatchFile` in all plugin hooks (#&#8203;[https://github.com/rollup/rollup/issues/5270](https://togithub.com/rollup/rollup/issues/5270)s/5270))

##### Bug Fixes

[...]

PR #177, opened 2023-12-03
chore(deps): update dependency typescript to v5

svc-secops:
This PR contains the following updates:

| Package | Change | Age | Adoption | Passing | Confidence |
|---|---|---|---|---|---|
| [typescript](https://www.typescriptlang.org/) ([source](https://togithub.com/Microsoft/TypeScript)) | [`4.9.5` -> `5.3.2`](https://renovatebot.com/diffs/npm/typescript/4.9.5/5.3.2) | [age](https://docs.renovatebot.com/merge-confidence/) | [adoption](https://docs.renovatebot.com/merge-confidence/) | [passing](https://docs.renovatebot.com/merge-confidence/) | [confidence](https://docs.renovatebot.com/merge-confidence/) |

---

### Release Notes

<details>
<summary>Microsoft/TypeScript (typescript)</summary>

### [`v5.3.2`](https://togithub.com/microsoft/TypeScript/releases/tag/v5.3.2): TypeScript 5.3

[Compare Source](https://togithub.com/Microsoft/TypeScript/compare/v5.2.2...v5.3.2)

For release notes, check out the [release announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-5-3/).

For the complete list of fixed issues, check out the

-   [fixed issues query for Typescript 5.3.0 (Beta)](https://togithub.com/Microsoft/TypeScript/issues?utf8=%E2%9C%93\&q=milestone%3A%22TypeScript+5.3.0%22+is%3Aclosed+).
-   [fixed issues query for Typescript 5.3.1 (RC)](https://togithub.com/Microsoft/TypeScript/issues?utf8=%E2%9C%93\&q=milestone%3A%22TypeScript+5.3.1%22+is%3Aclosed+).
-   [fixed issues query for Typescript 5.3.2 (Stable)](https://togithub.com/Microsoft/TypeScript/issues?utf8=%E2%9C%93\&q=milestone%3A%22TypeScript+5.3.2%22+is%3Aclosed+).

Downloads are available on:

-   [NuGet package](https://www.nuget.org/packages/Microsoft.TypeScript.MSBuild)

### [`v5.2.2`](https://togithub.com/microsoft/TypeScript/releases/tag/v5.2.2): TypeScript 5.2

[Compare Source](https://togithub.com/Microsoft/TypeScript/compare/v5.1.6...v5.2.2)

For release notes, check out the [release announcement](https://devblogs.microsoft.com/typescript/announcing-typescript-5-2/).

[...]

PR #181, opened 2023-12-04
chose(deps): update GraphQLCompiler dependencies

calvincestari:
This PR is expected to fail CI; I'm making it a full PR to verify the new GitHub Action is valid and works as expected. I will change this PR to a Draft once that is done.

**This PR should remain a Draft until there is a resolution to https://github.com/graphql/graphql-js/pull/3927. Until that is resolved the rollup command will fail with the following error:**
```
[!] RollupError: Expected ident (Note that you need plugins to import files that are not JavaScript)
node_modules/graphql/jsutils/instanceOf.mjs (11:43)
```

Combines the changes from PRs:
* https://github.com/apollographql/apollo-ios-dev/pull/174
* https://github.com/apollographql/apollo-ios-dev/pull/177
* https://github.com/apollographql/apollo-ios-dev/pull/180

Notes:
* There are also some manual dependency changes required to get compatibility with some of the other dependencies.
* The auto rollup script has been changed to fail on error (`set -e`) which will prevent partial updates to the Swift file which might be perceived as valid changes.
* Includes a new CI job to validate that the bundle can be built, rollup will succeed and there are no missed changes that should have been committed to the Swift file.


apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/auto_rollup.sh
```diff
@@ -1,8 +1,9 @@
 #!/usr/bin/env bash
+set -e
 
 SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
 output_file="$SCRIPT_DIR/../ApolloCodegenFrontendBundle.swift"
-$( cd "$SCRIPT_DIR" && rollup -c )
+cd "$SCRIPT_DIR" && npm run build
```
calvincestari, line 6: Use the package version of `rollup` instead of needing a global installation of it.

apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/package.json
```diff
@@ -6,27 +6,27 @@
...
-    "build": "rollup --config",
+    "build": "rollup --config --bundleConfigAsCjs",
     "test": "jest"
   },
   "engines": {
     "npm": ">=7"
   },
   "dependencies": {
-    "graphql": "17.0.0-alpha.3",
+    "graphql": "17.0.0-alpha.2",
```
calvincestari, line 16: This is the solution to fixing the Node error in the description. I do not believe we need anything in `alpha.3` - all our codegen tests still pass.

.github/workflows/ci-tests.yml
```diff
@@ -207,6 +207,24 @@ jobs:
       shell: bash
       run: |
         ./scripts/run-test-codegen-configurations.sh -t
+  
+  verify-frontend-bundle-latest:
```
AnthonyMDev, line 211: Would it be more efficient to include this as a step in the codegen unit tests CI job instead of spinning up another job? Not sure if it matters since the CI jobs on these runners are free minutes?

AnthonyMDev (approved):

PR #193, opened 2023-12-07
ci: Adds frontend bundle validation job

calvincestari:
With dependency updates being automated and having a few to deal with lately it's easy to miss something in the JS bundle, aka.  GraphQLCompiler; especially since it is a manual process to build and compile the changes into Swift.

This PR adds a bundle validation step that will install all the requisite node packages, build the bundle and then fail if there is a git diff which would indicate that there are changes we haven't included that should be.


calvincestari:
Closing this in favour of #181

PR #196, opened 2023-12-11
Update enums.mdx

trevorblades:


calvincestari:
Thanks for fixing the Netlify config @trevorblades!

PR #198, opened 2023-12-13
tests: TestCodegenConfigurations projects should fail on warnings

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3301.

I don't know when this stopped working in our test configuration projects but this behaviour is essential to avoid exposing inconvenience warnings to users projects.

I can't find any mention of why the `unsafeFlags` configuration no longer works but I do prefer that the flag is now set on the test script command which is more explicit to the workflow we want for the test - [reference](https://forums.swift.org/t/warnings-as-errors-for-libraries-frameworks/58393).

_I haven't managed to get the behaviour back for Xcode-based projects yet though._

### DIFF
--- a/Tests/TestCodeGenConfigurations/EmbeddedInTarget-InSchemaModule/Package.swift
+++ b/Tests/TestCodeGenConfigurations/EmbeddedInTarget-InSchemaModule/Package.swift
@@ -23,9 +23,6 @@ let package = Package(
       name: "TestApp",
       dependencies: [
         .product(name: "ApolloAPI", package: "apollo-ios")
-      ],
-      swiftSettings: [
-        .unsafeFlags(["-warnings-as-errors"])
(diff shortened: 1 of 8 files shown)


calvincestari:
As expected the [SPM-based test codegen configurations are now failing](https://github.com/apollographql/apollo-ios-dev/actions/runs/7190133181/job/19582767330?pr=198#step:5:308) with the warnings being treated as errors. The next commit will remove the warning and they should pass CI.

calvincestari:
..and the test codegen configurations job is [now passing](https://github.com/apollographql/apollo-ios-dev/actions/runs/7190250906/job/19583080798?pr=198), as expected.

BobaFetters (approved):
LGTM! Thanks for doing this!

PR #201, opened 2023-12-15
Tuist updates

BobaFetters:
Updating to newest tuist version to get multiplatofmr target support, and making a few tweaks to get rid of some warnings in the generated project

Closes apollographql/apollo-ios#3303


calvincestari:
It looks like the `Run Tuist Generation` test was skipped here. We should definitely have that CI test run right?

BobaFetters:
> It looks like the `Run Tuist Generation` test was skipped here. We should definitely have that CI test run right?

Ah yea didn't think about that since no changes are happening in the Swift code it is skipping the project generation, and I need to bump the version in the CI script

BobaFetters:
Going to add a sample change to a swift file to trigger the jobs and make sure they pass

AnthonyMDev (approved):
Not sure what's causing those test failures. Where is `NSImage` being used???

But otherwise LGTM

BobaFetters:
Looks like that fixed, going to remove that test change and merge. Also NSImage is used in `URLSessionClientTests`

PR #202, opened 2023-12-16
Codegen Vaildation - Handle Non Fatal Errors

AnthonyMDev:
This PR is based on top of #152. It should not be merged until after that PR is merged.

Adds `NonFatalErrors` error type and `NonFatalError` struct to collect and report non fatal errors that occur during the code generation.

Instead of being thrown when encountered, these errors are returned from the `TemplateRenderer` and `FileGenerator` objects to allow code generation execution to continue. After codegen is finished, any errors returned here are thrown from `ApolloCodegen.build()` before exiting.

This also includes changes to `SelectionSetTemplate` to implement the checking for type conflict errors incrementally as `ComputedSelectionSet`s are built, rather than requiring all `ComputedSelectionSet`s to be pre-built and retained for the entirety of execution.


calvincestari (approved):
This was a long one; lots of changes from merges to filter through. The gist of this looks good though. 👍🏻

PR #216, opened 2023-12-20
[docs] small tweaks to apollo-ios-cli doc

martinbonnin:
* `--schema-name` => `--schema-namespace`
* remove square brackets for required arguments. See https://stackoverflow.com/a/47026846/752630


calvincestari (approved):
Thanks Martin!

PR #217, opened 2023-12-21
feature(defer): Partial and Incremental Parsing

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3277
Closes https://github.com/apollographql/apollo-ios/issues/3145
Closes https://github.com/apollographql/apollo-ios/issues/3147

To recap the progress so far:
* `main` has some defer changes merged for codegen but they are disabled
* `feature/defer-execution-networking` is where I'm collecting the rest of the changes that will enable preview releases.

This PR brings parsing of the initial response (partial) and subsequent responses (incremental) to the defer implementation. The changeset includes changes to the templates and rendering but mostly are focused on the execution and networking layers. Details below:
* Removed `hasDeferredFragments` from the Operation definition. This was only used for the request HTTP headers but that behaviour has changed too and operations now get a `deferredFragments` property to help with incremental execution.
* `Deferrable` conformance was moved to the `InlineFragment` declaration to match the definition of `Fragment`. This allows us to clean up some of the inline fragment code generation.
* `IncrementalJSONResponseParsingInterceptor` is a new interceptor built to handle incremental payloads and combining them with previously parsed results.
* There is no collection of deferred fields when a `.deferred` selection set is encountered because those fields will not be present in the response data, but the deferred fragment is marked for future collection. Incremental execution will then pick up the deferred fragments fields for collection when the relevant response is received.
* `GraphQLExecutor` now has two entry points for execution; one used for the initial partial response and the other used during incremental execution parsing. They both converge in subsequent execution calls but the difference is in the generic constraints of the function and which selection set they operate on - `RootSelectionSet` or `SelectionSet`.
* `GraphQLResponse` has been broken up into a base struct named `AnyGraphQLResponse` which provides common logic between `GraphQLResponse`, used for partial responses, and `IncrementalGraphQLResponse` which is used for incremental response parsing.
* `DataDictMapper` was broken out of `GraphQLSelectionSetMapper` and handles the conversion of JSON data into the `DataDict` format.
* `GraphQLResult` now has the ability to merge in data from an `IncrementalGraphQLResult`. This is used during incremental response parsing and is required because of the way the interceptors require a single 'complete' result to be passed into the call to `proceedAsync`. The partial and incremental results are built upon each other during each incremental response storing the 'completed' response in the incremental interceptor which is unique to each operation request.
* `PathEntry` from `GraphQLError` was broken out into it's own type for reuse in the incremental execution.
* Caching is disabled for now so the cache write interceptor simply ignores caching behaviour for any operation using `@defer`. This will come in a later PR.

~There are no tests included with this PR yet, they're still coming.~ Last of the tests outstanding is for incremental execution.


apollo-ios/Sources/Apollo/AnyGraphQLResponse.swift
```diff
@@ -0,0 +1,100 @@
...
+      withRootCacheReference: rootKey,
+      variables: variables,
+      accumulator: accumulator
+    )
+  }
+
+  func execute<
+    Accumulator: GraphQLResultAccumulator,
+    Data: RootSelectionSet
+  >(
```
AnthonyMDev, line 48: We should try to narrow the scope of when this can be called so it's not inadvertently called on the root of an operation. Maybe make the `Data` have a `Deferrable` requirement?

If not possible, we should at least document the use cases and warn that this should not be used when executing an entire operation.

apollo-ios/Sources/Apollo/DefaultInterceptorProvider.swift
```diff
@@ -38,12 +38,21 @@ open class DefaultInterceptorProvider: InterceptorProvider {
...
         MultipartResponseParsingInterceptor(),
-        JSONResponseParsingInterceptor(),
+        jsonParsingInterceptor(for: operation),
         AutomaticPersistedQueryInterceptor(),
         CacheWriteInterceptor(store: self.store),
     ]
   }
 
+  private func jsonParsingInterceptor<Operation: GraphQLOperation>(for operation: Operation) -> any ApolloInterceptor {
+    if (Operation.deferredFragments?.isEmpty ?? true) {
```
AnthonyMDev, line 48: We are calling `deferredFragments.isEmpty` in mutliple places. Maybe create a computed property that is more clear on the intention here.

apollo-ios/Sources/Apollo/GraphQLExecutor.swift
```diff
@@ -191,10 +197,51 @@ final class GraphQLExecutor<Source: GraphQLExecutionSource> {
...
+      accumulator: accumulator
+    )
+  }
+
+  private func execute<
+    Accumulator: GraphQLResultAccumulator
+  >(
+    selectionSet: any SelectionSet.Type,
+    on data: Source.RawObjectData,
+    variables: GraphQLOperation.Variables? = nil,
```
AnthonyMDev, line 236: variables should come after withRootCacheReference to mirror above methods.

apollo-ios/Sources/Apollo/FieldSelectionCollector.swift
```diff
@@ -27,6 +29,10 @@ struct FieldSelectionGrouping: Sequence {
     fulfilledFragments.insert(ObjectIdentifier(type))
```
AnthonyMDev, line 29: Add a precondition check that a selection set is not added to both `fulfilledFragments` and `deferredFragments`.

apollo-ios/Sources/Apollo/GraphQLResult.swift
```diff
@@ -76,3 +128,48 @@ extension GraphQLResult {
...
+    case invalidPathDataType(String)
+    case cannotOverwriteData(AnyHashable, AnyHashable)
+
+    public var errorDescription: String? {
+      switch self {
+      case let .invalidPathDataType(got):
+        return "Invalid data type for merge - \(got)"
+
+      case let .cannotOverwriteData(current, new):
+        return "Incremental merging cannot overwrite existing data \(current) with \(new)"
```
AnthonyMDev, line 143: Lets make this error message more clear. Indicated that increment response had mismatching value for an existing field.

apollo-ios/Sources/Apollo/GraphQLResult.swift
```diff
@@ -1,38 +1,90 @@
...
+  ) {
     self.data = data
     self.extensions = extensions
     self.errors = errors
     self.source = source
     self.dependentKeys = dependentKeys
   }
+
+  func merging(_ incrementalResult: IncrementalGraphQLResult) throws -> GraphQLResult<Data> {
+    var mergedData = self.data
```
AnthonyMDev, line 41: We can consider using a protocol or function that you pass a closure to in order to clean up the repetitive part of this merging.

apollo-ios/Sources/Apollo/MultipartResponseDeferParser.swift
```diff
@@ -64,12 +64,14 @@ struct MultipartResponseDeferParser: MultipartResponseSpecificationParser {
         }
 
       case let .json(object):
-        if let hasNext = object.hasNext {
-          preconditionFailure("This will be done in #3147")
-        }
+        if let _ = object.incremental {
```
AnthonyMDev, line 67: We don't need to check `.incremental` and `.object`. We do the same serialization either way. So let's just do it.

apollo-ios/Sources/Apollo/SubscriptableByPathComponent.swift
```diff
@@ -0,0 +1,121 @@
+#if !COCOAPODS
+import ApolloAPI
+#endif
+
+/// Enables subscript syntax to be used where the subscript is an array of `PathComponent` values
+/// typically returned in a GraphQL response.
+protocol SubscriptableByPathComponent {
```
AnthonyMDev, line 7: This definitely needs a lot of unit test.

I think we should change the name to `PathComponentSubscriptAccessible`

apollo-ios/Sources/Apollo/SubscriptableByPathComponent.swift
```diff
@@ -0,0 +1,121 @@
...
+  subscript(path: [PathComponent]) -> AnyHashable? { get set }
```
AnthonyMDev, line 8: This function should just be in the extension and not declared on the protocol.

apollo-ios/Sources/Apollo/SubscriptableByPathComponent.swift
AnthonyMDev: We should make the subscript throw and pull the getting and setting out of the extensions on `DataDict` and `[AnyHashable]` to all be in the protocol extension. Checking for correct type for the next path component should be done there.

apollo-ios/Sources/ApolloAPI/GraphQLOperation.swift
```diff
@@ -53,13 +53,26 @@ public struct OperationDefinition {
   }
 }
 
+/// A unique identifier used as a key to map a deferred selection set type to an incremental
+/// response label and path.
+public struct DeferredFragmentIdentifier: Hashable {
+  let label: String
+  let path: [String]
```
AnthonyMDev, line 60: rename to fieldPath

apollo-ios/Sources/ApolloAPI/GraphQLOperation.swift
```diff
@@ -71,10 +84,17 @@ public extension GraphQLOperation {
...
-  /// fragment marked with the `@defer` directive.
-  static var hasDeferredFragments: Bool {
-    false
+  static var deferredFragments: [DeferredFragmentIdentifier: any SelectionSet.Type]? {
+    return nil
+  }
+
+  static func deferredSelectionSetType(
+    withLabel label: String,
+    atPath path: [JSONValue]
```
AnthonyMDev, line 93: This should take in `[PathComponent]`

apollo-ios/Sources/Apollo/IncrementalGraphQLResponse.swift
```diff
@@ -0,0 +1,90 @@
...
+      body: body,
+      rootKey: rootKey,
+      variables: operation.__variables
+    )
+  }
+
+  func parseIncrementalResult() throws -> IncrementalGraphQLResult {
+    guard 
+      let label = base.body["label"] as? String,
+      let path = base.body["path"] as? [JSONValue],
```
AnthonyMDev, line 42:
```suggestion
      let jsonPath = base.body["path"] as? [JSONValue],
      let path = jsonPath.compactMap(PathComponent.init)
```

apollo-ios/Sources/Apollo/IncrementalGraphQLResponse.swift
```diff
@@ -0,0 +1,90 @@
...
+    let selectionSet: (any SelectionSet)?
+    if let data = executionResult?.0 {
+      selectionSet = selectionSetType.init(_dataDict: data)
+    } else {
+      selectionSet = nil
+    }
+
+    return IncrementalGraphQLResult(
+      label: label,
+      path: path.compactMap(PathComponent.init),
```
AnthonyMDev, line 69:
```suggestion
      path: path,
```

AnthonyMDev (changes requested):

apollo-ios/Sources/Apollo/DefaultInterceptorProvider.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/217/commits/12ca32c502448d358d9d4a8a1022996e28e7e989.

apollo-ios/Sources/Apollo/FieldSelectionCollector.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/217/commits/a92e0e6f57e8c02b74a105ba9d0fb13e2d16c70e.

apollo-ios/Sources/Apollo/GraphQLResult.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/217/commits/6d737393cc3b85b54140171e3828574ce1832268.

apollo-ios/Sources/Apollo/GraphQLResult.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/217/commits/34c171ed6be6c04bfe0b7081b4548f6d9b8c8566.

apollo-ios/Sources/Apollo/MultipartResponseDeferParser.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/217/commits/92c094355eda77f85969f28d92132b501586672a.

apollo-ios/Sources/ApolloAPI/GraphQLOperation.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/217/commits/6907fd1420facf2e97af380d500e60edd2b299d2.

apollo-ios/Sources/Apollo/GraphQLExecutor.swift
calvincestari: Done in Done in https://github.com/apollographql/apollo-ios-dev/pull/217/commits/a92e0e6f57e8c02b74a105ba9d0fb13e2d16c70e..

apollo-ios/Sources/ApolloAPI/GraphQLOperation.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/217/commits/292cc50d4af5bbe997995d5cfa376051e3e65cef.

I changed this to take in `[String]` instead of `[PathComponent]` because `PathComponent` is defined in `Apollo` and `ApolloAPI` has none, and shouldn't have, knowledge of the `Apollo` module. Moving `PathComponent` into `ApolloAPI` didn't seem correct just for this so the logic of filtering out indexes is done by the caller now instead and this function is given a path array of only field names.

apollo-ios/Sources/Apollo/SubscriptableByPathComponent.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/217/commits/dd6b435640e9a376100e3835e16367638b2b0e34.

The refactoring for this includes:
* I got rid of the subscript syntax because the merge process should throw if any errors are found and Swift subscripts cannot throw. There is also no reason it had to be a subscript.
* I pulled all the merging logic into a single file named `DataDict+Merging`. This is the `internal` entry point for getting and setting a value by `[PathComponent]`. The rest of the functions in that file are `fileprivate` because they're internal implementation details. This also negated the need for subscript syntax.
* I kept the protocol but it is renamed to `PathComponentAccessible` and it's needed to escape the type-erased nature of the the protocol for the type-specific mutable setters. We can discuss this in the next review but I think what we've ended up with here is pretty good and concise.

apollo-ios/Sources/Apollo/AnyGraphQLResponse.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/217/commits/4ee07f0a97e89fe044d90a5570acc5f7a9c6c18e and https://github.com/apollographql/apollo-ios-dev/pull/217/commits/98e61d10ba84783ccc37de1ddf1d68b7cebc55ff.

calvincestari:
@AnthonyMDev I think this PR is finally ready for review.

apollo-ios/Sources/Apollo/AnyGraphQLResponse.swift
```diff
@@ -0,0 +1,104 @@
+#if !COCOAPODS
+import ApolloAPI
+#endif
+
+/// An abstract GraphQL response used for full and incremental responses.
+struct AnyGraphQLResponse {
```
AnthonyMDev, line 6: I see that the properties in this struct are `public` but the struct itself is `internal`. That makes those `public` properties not actually `public`. Is this supposed to be `public`, or can we leave it `internal`? If it doesn't need to be public, I'd prefer to leave it internal, but then we should remove the `public` modifiers on the properties (and the functions in the extensions), just for clarity.

apollo-ios/Sources/Apollo/GraphQLExecutor.swift
```diff
@@ -191,10 +197,51 @@ final class GraphQLExecutor<Source: GraphQLExecutionSource> {
...
+      selectionSet: selectionSet,
+      on: data,
+      withRootCacheReference: root,
       variables: variables,
       schema: SelectionSet.Schema.self,
+      accumulator: accumulator
+    )
+  }
+
+  func execute<
```
AnthonyMDev, line 210: It feels like we could really still just use the one execution function above if we just change it from needed `RootSelectionSet` to just `SelectionSet`. It doesn't look like there is any other actual deviation in functionality between these two functions. Am I missing something?

apollo-ios/Sources/Apollo/GraphQLResult.swift
```diff
@@ -5,34 +5,95 @@ import ApolloAPI
...
+      errors: mergedErrors,
+      source: source,
+      dependentKeys: mergedDependentKeys
+    )
+  }
+
+  fileprivate func merge<T>(
+    _ newValue: T?,
+    into currentValue: T?,
+    onCurrent: (_ currentValue: T, _ newValue: T) throws -> T
```
AnthonyMDev, line 84: What does the parameter name `onCurrent` mean? I'm not able to understand what the closure I should be passing into this parameter is supposed to be if I were calling this function.

apollo-ios/Sources/ApolloAPI/GraphQLOperation.swift
```diff
@@ -88,22 +111,36 @@ public extension GraphQLOperation {
...
 }
 
+// MARK: - GraphQLMutation
+
 public protocol GraphQLMutation: GraphQLOperation {}
 public extension GraphQLMutation {
   @inlinable static var operationType: GraphQLOperationType { return .mutation }
 }
 
+// MARK: - GraphQLSbuscription
```
AnthonyMDev, line 142:
```suggestion
// MARK: - GraphQLSubscription
```

AnthonyMDev (changes requested):
I've just reviewed the production code, and had a few very minor comments. I'm going to review the tests shortly as well!

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -7407,7 +7407,7 @@ class SelectionSetTemplateTests: XCTestCase {
     expect(rendered_allAnimals_deferredAsRoot).to(equalLineByLine(
       """
       /// AllAnimal.Root
-      public struct Root: TestSchema.InlineFragment, ApolloAPI.Deferrable {
+      public struct Root: TestSchema.InlineFragment {
```
AnthonyMDev, line 7410: Do these tests even serve a purpose anymore? If we aren't adding the `Deferrable` protocol conformance, then what behavior are we testing?

AnthonyMDev (approved):
The tests LGTM! Just a couple minor questions.

apollo-ios/Sources/Apollo/AnyGraphQLResponse.swift
calvincestari: I'm not sure why they're public but they don't need to be. I'll reduce their permissions scope, that's a good catch!

apollo-ios/Sources/Apollo/GraphQLResult.swift
calvincestari: This is just poor follow-up on my part after refactoring. `onMerge` might be more appropriate since the closure is what gets called when there is both a new and current value.

Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
```diff
@@ -4384,9 +4396,11 @@ class IRRootFieldBuilderTests: XCTestCase {
     expect(Array(self.computedReferencedFragments)).to(equal(expected))
   }
 
-  // MARK: - Deferred Fragments - hasDeferredFragments property
+  // MARK: - Deferred Fragments - containsDeferredFragment property
+
+  #warning("Need tests for deferredFragments property")
```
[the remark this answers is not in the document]
calvincestari: This will be removed in the next PR (https://github.com/apollographql/apollo-ios-dev/pull/219) which takes care of the deferred fragment metadata definitions.

Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
```diff
@@ -52,6 +54,16 @@ class IRRootFieldBuilderTests: XCTestCase {
     subject = result.rootField
   }
 
+  fileprivate func addDeferDirective() {
```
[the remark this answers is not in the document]
calvincestari: I added a comment about this in 93bd99d.

Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
calvincestari: Yes I see your point; I'll handle this in the next PR (https://github.com/apollographql/apollo-ios-dev/pull/219) which deals with codegen.

apollo-ios/Sources/Apollo/FieldSelectionCollector.swift
```diff
@@ -27,6 +29,10 @@ struct FieldSelectionGrouping: Sequence {
...
   }
 
+  mutating func addDeferredFragment<T: SelectionSet>(_ type: T.Type) {
```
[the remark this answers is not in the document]
calvincestari: I don't believe this is possible because the directive definition doesn't allow `@defer` to be used on fields. It's only valid on inline and named fragments - `directive @defer(label: String, if: Boolean! = true) on FRAGMENT_SPREAD | INLINE_FRAGMENT`.

apollo-ios/Sources/Apollo/GraphQLExecutor.swift
calvincestari: OK. I've made a note that we should chat about it this week and if there are changes I'll make a separate PR for that.

PR #218, opened 2024-01-03
Fixing fragment definition generation when not generating operation definitions

BobaFetters:
Only generate definitions in fragment files if the 'operationDocumentFormat' config contains the 'definition' value.

Closes apollographql/apollo-ios#3282


AnthonyMDev (commented):
I'm confused how the tests are passing here. `fragmentDefinition` is a requirement of the `Fragment` protocol, so generating a fragment without this property should make it not conform to the `Fragment` protocol.

BobaFetters:
> I'm confused how the tests are passing here. `fragmentDefinition` is a requirement of the `Fragment` protocol, so generating a fragment without this property should make it not conform to the `Fragment` protocol.

@AnthonyMDev Going to assume it has to do with it just doing string comparison of what gets generated, more tests would still need to be added to cover actual compilation before this patch would be ready to go.

Mainly got this up so we could have a discussion about how/if we wanted to handle this

AnthonyMDev:
Ah I see. Yes, I do think this is the right approach. We don't need to separate configuration option for disabling fragment definitions, it should use the `operationDocumentFormat` as you've done here.

We just need to figure out the right way to make this compile. We could make the `Fragment` protocol have a default implementation of `fragmentDefinition` that is just an empty `String` so we don't break backwards compatibility I guess?

The Codegen Test Configurations CI job actually generates the code, compiles a project using it and runs some tests. So maybe we need to add a configuration there that has this option turned off.

Tests/TestCodeGenConfigurations/EmbeddedInTarget-InSchemaModule/apollo-codegen-config.json
```diff
@@ -13,7 +13,8 @@
     },
     "pruneGeneratedFiles" : true,
     "queryStringLiteralFormat" : "multiline",
-    "warningsOnDeprecatedUsage" : "include"
+    "warningsOnDeprecatedUsage" : "include",
+    "operationDocumentFormat" : ["operationId"]
```
AnthonyMDev, line 17: We don't need to address this in this current PR. But for integration testing individual configuration options, maybe we could build something better here.

We need to have these multiple different project setups in `TestCodeGenConfigurations` for testing the actual project configurations (Cocoapods, SPM, Xcode) and codegen methods (Relative vs absolute, schema module embedded, etc).

These are things that affect how/where the files are generated and if they include the correct namespacing and import statements to link and compile properly.

But for testing other individual configuration properties which would be the same regardless of these configurations, we should probably have a single project that we re-generate with different configurations so that we can isolate and test _just_ the configuration options we want to test. These could be done using the Swift Scripting codegen to make it easier to run multiple of them rapidly. 

This will take some time to set up properly, so it's not something I think we need to do right now. But this isn't the first time I've noticed that we are kind of overloading these integration tests, so it might be worth thinking about.

apollo-ios/Sources/ApolloAPI/FragmentProtocols.swift
```diff
@@ -9,6 +9,14 @@ public protocol Fragment: SelectionSet, Deferrable {
   static var fragmentDefinition: StaticString { get }
 }
 
+/// Extension providing default implementation for the ``Fragment`` protocol.
+extension Fragment {
+  // Default implementation for the `fragmentDefinition` variable
+  static var fragmentDefinition: StaticString {
```
AnthonyMDev, line 15: Does this need to be `public`?

AnthonyMDev (approved):

apollo-ios/Sources/ApolloAPI/FragmentProtocols.swift
AnthonyMDev: I think the failing CI test is because of this actually.

BobaFetters: It does indeed, forgot to mark it public

PR #219, opened 2024-01-03
feature(defer): Code generation

calvincestari:
This PR adds the missing pieces to codegen for the `@defer` implementation:
1. **Deferred fragment metadata** - this metadata is needed to support incremental parsing which needs to be able to link a deferred selection set to a defer label received in each incremental response.
2. **Operations that use `@defer` on a named fragment** - the template changes in this draft PR at the moment only solve for simple named fragment use cases in order to ready for a preview release. As I add the test cases I'll resolve for all use cases of deferred named fragments.

**There are no tests included with this PR yet, they're still coming**


apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
```diff
@@ -538,13 +568,22 @@ struct SelectionSetTemplate {
     let isOptional =
       fragment.inclusionConditions != nil
       && !scope.matches(fragment.inclusionConditions.unsafelyUnwrapped)
+    let isDeferred = fragment.typeInfo.deferCondition != nil
 
     return """
```
calvincestari, line 573: I think we might be at the point where it makes sense to break this function into smaller focused functions.

apollo-ios-codegen/Sources/IR/IR+NamedFragmentSpread.swift
```diff
@@ -18,7 +18,7 @@ public class NamedFragmentSpread: Hashable, CustomDebugStringConvertible {
   /// Indicates the location where the fragment has been "spread into" its enclosing
   /// operation/fragment. It's `scopePath` and `entity` reference are scoped to the operation it
   /// belongs to.
-  let typeInfo: SelectionSet.TypeInfo
+  public let typeInfo: SelectionSet.TypeInfo
```
calvincestari, line 21: Is there any reason this shouldn't be a public constant? Doing so matches what we make available on `InlineFragment` and means we don't need to have a public method just to get at `deferCondition`. The type info is also used in the deferred fragment metadata template, but that may change.

apollo-ios/Sources/Apollo/GraphQLResult.swift
```diff
@@ -151,7 +151,12 @@ fileprivate extension DataDict {
     }
 
     let mergedData = try pathDataDict._data.merging(newDataDict._data) { current, new in
-      throw Error.cannotOverwriteData(current, new)
+      // TODO: This is a quick fix for __typename being returned with fragments, needs a rethink!
```
calvincestari, line 154: To throw here seems the correct behaviour because there should never be a case of overwriting response data; incremental data is only ever additive. The exception to this rule is `__typename` which is being returned for fragments on the same type case and already exists in the base data set.

apollo-ios-codegen/Sources/IR/IR+NamedFragmentSpread.swift
Karliz24: defer/oustanding-defer-codegen

Karliz24 (commented):
********

calvincestari:
Be sure to handle these two items in this PR too:
* https://github.com/apollographql/apollo-ios-dev/pull/217#discussion_r1531105715
* https://github.com/apollographql/apollo-ios-dev/pull/217#discussion_r1531109692

calvincestari:
Closing this as the work has been split into PRs:
1. https://github.com/apollographql/apollo-ios-dev/pull/327
2. https://github.com/apollographql/apollo-ios-dev/pull/328

PR #227, opened 2024-01-08
[chore] Add user survey to READMEs

calvincestari:
1. Adds a link to the user survey in the README of `apollo-ios`, `apollo-ios-codegen` and `apollo-ios-dev`; _when `apollo-ios-pagination` is at 1.0 we can add it there too_. The link leads to a Google Form with pre-filled content to help us determine the target and source of the survey.

2. Adds a GitHub Action that adds a comment to an issue when it is closed, for **both** `completed` and `not planned` status closures. The link leads to the same Google Form but with different pre-filled content.


AnthonyMDev (approved):
LGTM

bignimbus (approved):

calvincestari:
@AnthonyMDev, @bignimbus - I've cleaned up this PR and added the GH action for closed issue comments if you wanted to take another look at the diff before I merge it.

@alessbell, @martinbonnin - this PR can serve as a reference for getting the same change into your project repos but we'll create different pre-filled links to use.

calvincestari:
If you wanted to see an example of what this looks like, I've got it working in https://github.com/calvincestari/actions-test/issues/1.

martinbonnin:
Could we have the bot skip the comment if the original poster is part of the CODEOWNERS?

AnthonyMDev (approved):

calvincestari:
> Could we have the bot skip the comment if the original poster is part of the CODEOWNERS?

It doesn't look like there is any way to determine that through the default [GitHub Actions Contexts](https://docs.github.com/en/actions/learn-github-actions/contexts) which means we'll need a custom action in JS to determine that.

calvincestari:
> Could we have the bot skip the comment if the original poster is part of the CODEOWNERS?

I thought about this some more and it may still be valuable to have the survey comment even if the issue was created by a maintainer/codeowner because there may have been other participants in the conversation that we could gather feedback from.

So for now I'm going to merge this as-is and see how it goes and be ready to adapt if we need to.

alessbell:
> Could we have the bot skip the comment if the original poster is part of the CODEOWNERS?

I've done this elsewhere by skipping all members of the apollographql org, you can use the gh CLI to call the REST endpoint:

```bash
APOLLO_GRAPHQL_ORG_MEMBERS=$(gh api -H "Accept: application/vnd.github+json" -H "X-GitHub-Api-Version: 2022-11-28" /orgs/apollographql/members --paginate --jq '[.[] | .login]')
```

PR #228, opened 2024-01-08
Add @import directive

AnthonyMDev:
This adds a new `@import(module:)` directive that can be used when defining operations or fragments to be used as input to the Apollo iOS Codegen Engine. 

This PR only adds the new directive to the `graphql-js` frontend wrapper.

In following PRs:

- [x] The `@import` statement directives will be stripped from the definition string that is sent on network requests (and used in persisted query lists). (done in #229)
- [ ] Each `@import` directive instance will add an import statement to the generated Swift definition file that imports a module with the value of the `module` argument.


calvincestari (approved):
LGTM!

PR #232, opened 2024-01-10
chore: Removes boundary from `accept` header for Subscription operations

calvincestari:
The `boundary` portion of the `accept` header has no effect on the `boundary` in the received multipart response so we can remove it and send less data in the request. The multipart response handler is already built to handle any value sent back as the `boundary` and use that to parse the chunks.

More details in https://github.com/apollographql/router/pull/4395 and https://github.com/apollographql/router/pull/4447.


calvincestari:
@martinbonnin is this change applicable only to subscriptions or should we be making the change for the `accept` header in defer requests too? I believe the same logic would apply there but are the changes being made to the router consistent across both features?

martinbonnin:
Yup, changes are being made to both `@defer` and subs: See https://github.com/apollographql/router/pull/4447/files#diff-50ebf56cea38746e9a011ee0e8dc7a210f8c8abfb740ae8ec6f67adb75b599eb

AnthonyMDev (approved):

PR #234, opened 2024-01-10
chore: Removes boundary from accept header

calvincestari:
This is the same as https://github.com/apollographql/apollo-ios-dev/pull/232 but limited to non-subscription operations and to be merged into the `feature/defer-execution-networking` branch to only be applicable once all the defer work is merged into `main`.

> The `boundary` portion of the `accept` header has no effect on the `boundary` in the received multipart response so we can remove it and send less data in the request. The multipart response handler is already built to handle any value sent back as the `boundary` and use that to parse the chunks.

> More details in https://github.com/apollographql/router/pull/4395 and https://github.com/apollographql/router/pull/4447.

### DIFF
--- a/Tests/ApolloTests/RequestChainTests.swift
+++ b/Tests/ApolloTests/RequestChainTests.swift
@@ -380,7 +380,7 @@ class RequestChainTests: XCTestCase {
         return
       }
 
-      XCTAssertEqual(header, "multipart/mixed;boundary=\"graphql\";\(MultipartResponseDeferParser.protocolSpec),application/json")
+      XCTAssertEqual(header, "multipart/mixed;\(MultipartResponseDeferParser.protocolSpec),application/json")
       expectation.fulfill()
     }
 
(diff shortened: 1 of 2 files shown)


AnthonyMDev (approved):

calvincestari:
I'm going to go ahead and merge this with the test failure because the test failure is caused by a warning in the test codegen configuration job. The warning is about unused property assignments that will be handled in PR #217 as part of [review feedback](https://github.com/apollographql/apollo-ios-dev/pull/217/files#r1446558563).

> /Users/runner/work/apollo-ios-dev/apollo-ios-dev/apollo-ios/Sources/Apollo/MultipartResponseDeferParser.swift:67:16: error: value 'hasNext' was defined but never used; consider replacing with boolean test
>         if let hasNext = object.hasNext {
>            ~~~~^~~~~~~~~~
>                                         != nil
> /Users/runner/work/apollo-ios-dev/apollo-ios-dev/apollo-ios/Sources/Apollo/MultipartResponseDeferParser.swift:71:16: error: value 'incremental' was defined but never used; consider replacing with boolean test
>         if let incremental = object.incremental {
>            ~~~~^~~~~~~~~~~~~~
>                                                 != nil

PR #235, opened 2024-01-10
fix: `@defer` directive is duplicated when schema already has one

calvincestari:
This is a quick-fix for users on the router with a schema that supports defer.

The defer code merged into `main` manually adds the `@defer` directive to the schema because it is still an experimental spec directive. This is causing a duplicate directive error and is blocking codegen.

All the Swift codegen tests for defer as being skipped now. This is OK-ish because defer is disabled in the JS frontend anyways so nobody can actually generate deferred operation models from `main`. This will be reverted in the `feature/defer-execution-networking` branch.


AnthonyMDev (approved):
I would think that using `throw XCTSkip` would cause the tests to show an error, but I guess the skip error is handled properly by `XCTest`? So if all the tests are passing, this LGTM.

PR #236, opened 2024-01-11
import directive

hemel:
This PR handles the import directive.
If a graphql operation adds this directive, an import statement to the corresponding generated operation file will be added.

Open question:

How should the import directive work for nested fragments?
 
When adding the directive to a fragment C in the case of
`Query A { ...B }  Fragment B { ...C } Fragment C @import(module: "MyModule") { ...FragmentFromMyModule}`

The enclosing fragments and queries (A,B) will reference fragments from inside C, which would be included in the imported module. Should the codegen add the import statement automatically to the Fragment B and Query A, or should the dev explicitly add the import directive in A and B?

This PR implements the case of explicitly adding the import directives to any operation which includes any fragment from the imported module.

### DIFF
--- a/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
+++ b/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
@@ -7,7 +7,7 @@ enum TemplateTarget: Equatable {
   /// Used in schema types files; enum, input object, union, etc.
   case schemaFile(type: SchemaFileType)
   /// Used in operation files; query, mutation, fragment, etc.
-  case operationFile
+  case operationFile(importModules: [String])
   /// Used in files that define a module; Swift Package Manager, etc.
   case moduleFile
   /// Used in test mock files; schema object `Mockable` extensions
@@ -23,7 +23,7 @@ enum TemplateTarget: Equatable {
     case customScalar
     case inputObject
 
-    var namespaceComponent: String? {      
+    var namespaceComponent: String? {
       switch self {
       case .schemaMetadata, .enum, .customScalar, .inputObject, .schemaConfiguration:
         return nil
@@ -103,7 +103,7 @@ extension TemplateRenderer {
     let body = {
       switch target {
       case let .schemaFile(type): return renderSchemaFile(type, errorRecorder)
-      case .operationFile: return renderOperationFile(errorRecorder)
+      case let .operationFile(importModules): return renderOperationFile(importModules, errorRecorder)
       case .moduleFile: return renderModuleFile(errorRecorder)
       case .testMockFile: return renderTestMockFile(errorRecorder)
       }
@@ -153,12 +153,14 @@ extension TemplateRenderer {
   }
 
   private func renderOperationFile(
+    _ importModules: [String],
     _ errorRecorder: ApolloCodegen.NonFatalError.Recorder
   ) -> String {
     TemplateString(
     """
     \(ifLet: renderHeaderTemplate(nonFatalErrorRecorder: errorRecorder), { "\($0)\n" })
     \(ImportStatementTemplate.Operation.template(for: config))
+    \(AdditionalImportStatementTemplate.template(importModules: importModules))
 
     \(if: config.output.operations.isInModule && !config.output.schemaTypes.isInModule,
       renderBodyTemplate(nonFatalErrorRecorder: errorRecorder)
@@ -348,6 +350,23 @@ struct ImportStatementTemplate {
       """
     }
   }
+}
+
+/// Provides the format to import additional Swift modules required by the template type.
+struct AdditionalImportStatementTemplate {
+
+  static func template(
+    importModules: [String]
+  ) -> TemplateString {
+    return """
+    \(forEachIn: importModules, {
+      return """
+      import \($0)
+      """
+    })
+    """
+  }
+    
 }
 
 fileprivate extension ApolloCodegenConfiguration {
--- a/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
+++ b/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
@@ -7,7 +7,7 @@ enum TemplateTarget: Equatable {
   /// Used in schema types files; enum, input object, union, etc.
   case schemaFile(type: SchemaFileType)
   /// Used in operation files; query, mutation, fragment, etc.
-  case operationFile(importModules: [String]?)
+  case operationFile(moduleImports: [String]?)
   /// Used in files that define a module; Swift Package Manager, etc.
   case moduleFile
   /// Used in test mock files; schema object `Mockable` extensions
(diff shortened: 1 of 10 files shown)


apollo-ios-codegen/Sources/GraphQLCompiler/CompilationResult.swift
```diff
@@ -139,12 +142,19 @@ public final class CompilationResult: JavaScriptObjectDecodable {
...
       self.rootType = rootType
       self.selectionSet = selectionSet
-      self.directives = directives
       self.referencedFragments = referencedFragments
       self.source = source
       self.filePath = filePath
       self.isLocalCacheMutation = directives?
         .contains { $0.name == Constants.DirectiveNames.LocalCacheMutation } ?? false
+        
+      let referencedImports: [Directive] = self.referencedFragments
```
AnthonyMDev, line 151: I see what you're doing here, but I don't think we should be adding the referenced fragment's `@import` directives to the parent's `directives`. We should add them to the `moduleImports` but not the actual `directives` property.

apollo-ios-codegen/Sources/GraphQLCompiler/CompilationResult.swift
```diff
@@ -139,12 +142,19 @@ public final class CompilationResult: JavaScriptObjectDecodable {
...
+        .compactMap { $0.directives?.filter { $0.name == Constants.DirectiveNames.Import } }
+        .flatMap { $0 }
+      self.directives = Array(Set(directives ?? [] + referencedImports))
+
+      self.moduleImports = self.directives?
```
AnthonyMDev, line 156: I'm also wondering if we should alphabetize these so that changing the order of Fragment Spreads doesn't change the order of import statements. That could be confusing when you change something in one Fragment definition and then a bunch of operations have their `import` statements change.

I see the use of `Set` when setting up the `directives` here. But if we are going to not add the referenced fragment imports to the `directives`, we should probably just make `moduleImports` an `OrderedSet` to ensure we don't get duplicate import statements.

apollo-ios-codegen/Sources/GraphQLCompiler/CompilationResult.swift
AnthonyMDev: This means `moduleImports` probably just becomes a stored property instead of being computed.

apollo-ios-codegen/Sources/GraphQLCompiler/CompilationResult.swift
```diff
@@ -214,15 +224,24 @@ public final class CompilationResult: JavaScriptObjectDecodable {
...
     init(_ jsValue: JSValue, bridge: isolated JavaScriptBridge) {
       self.name = jsValue["name"]
       self.type = .fromJSValue(jsValue["typeCondition"], bridge: bridge)
       self.selectionSet = .fromJSValue(jsValue["selectionSet"], bridge: bridge)
-      self.directives = .fromJSValue(jsValue["directives"], bridge: bridge)
       self.referencedFragments = .fromJSValue(jsValue["referencedFragments"], bridge: bridge)
       self.source = jsValue["source"]
       self.filePath = jsValue["filePath"]
+        
+      let directives: [Directive]? = .fromJSValue(jsValue["directives"], bridge: bridge)
```
AnthonyMDev, line 240: Same as above. We should also try to break this into a static function so that we can re-use the logic instead of repeating it in both places.

apollo-ios-codegen/Sources/GraphQLCompiler/CompilationResult.swift
```diff
@@ -650,7 +669,34 @@ public final class CompilationResult: JavaScriptObjectDecodable {
...
+            where: { $0.name == Constants.ArgumentNames.Module }),
+          case let .string(moduleValue) = moduleArgument.value
+        else {
+          return nil
+        }
+        moduleName = moduleValue
+      }
+    
+      var debugDescription: String {
+        return "Import \"\(moduleName)\""
```
AnthonyMDev, line 697:
```suggestion
        return "@import(module: \"\(moduleName)\")"
```

apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/TemplateRenderer.swift
```diff
@@ -7,7 +7,7 @@ enum TemplateTarget: Equatable {
   /// Used in schema types files; enum, input object, union, etc.
   case schemaFile(type: SchemaFileType)
   /// Used in operation files; query, mutation, fragment, etc.
-  case operationFile
+  case operationFile(moduleImports: [String]?)
```
AnthonyMDev, line 10: Can we make this have a default value of `nil` so we don't need to provide `moduleImports: nil` in all of the unit tests?

I can't remember if you can do default values on enum case associated values? 🤔

AnthonyMDev (changes requested):
Just some minor improvements to be made here!

I also don't see any new unit tests for this. If you are unsure how to write the tests, let me know and I'll pull this down and write a few tests.

hemel:
> Just some minor improvements to be made here!
> 
> I also don't see any new unit tests for this. If you are unsure how to write the tests, let me know and I'll pull this down and write a few tests.

Hello @AnthonyMDev thank you for reviewing the PR. I applied the changes based on your review. In terms of the tests i was not sure where to add them unfortunately. I would greatly appreciate it if you can add some. Thank you in advance.

apollo-ios-codegen/Package.resolved
```diff
@@ -23,8 +23,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-collections",
       "state" : {
-        "revision" : "937e904258d22af6e447a0b72c0bc67583ef64a2",
-        "version" : "1.0.4"
+        "revision" : "d029d9d39c87bed85b1c50adee7c41795261a192",
+        "version" : "1.0.6"
```
hemel, line 27: OrderedSet was not Sendable (resulting in a warning) in version 1.0.4 so i updated collections to 1.0.6 as it was fixed there.

apollo-ios-codegen/Sources/GraphQLCompiler/CompilationResult.swift
```diff
@@ -255,6 +279,18 @@ public final class CompilationResult: JavaScriptObjectDecodable {
     public static func ==(lhs: FragmentDefinition, rhs: FragmentDefinition) -> Bool {
       return lhs.name == rhs.name
     }
+    
+    private static func getImportModuleNames(directives: [Directive]?,
```
hemel, line 283: I can add this as a global function in order to avoid any code duplication. For now i left this helper inside the corresponding class.

apollo-ios-codegen/Package.resolved
AnthonyMDev: That's great. If we need that patch version in order for this to compile correctly, then we should probably make it the minimum version number in the `Package.swift` file too.

AnthonyMDev (commented):
This all looks great! I'm going to leave this PR pending until I can take some time to write the unit tests for it. If you get the opportunity to use this branch with your project, it would be great to get some real-world testing too.

I'll get the tests done and have this shipped for the next release we make. And I'll be sure that's before the end of the month for you!

hemel:
Ill double check to see if we can use the fork on our side @AnthonyMDev. Thank you again!

PR #243, opened 2024-01-22
fix not used implementation

matsudamper:
fix: https://github.com/apollographql/apollo-ios/issues/3323

fix Custom Scalar should not be edited. output

### DIFF
--- a/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/CustomScalarTemplate.swift
+++ b/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/CustomScalarTemplate.swift
@@ -14,7 +14,7 @@ struct CustomScalarTemplate: TemplateRenderer {
 
   func renderHeaderTemplate(
     nonFatalErrorRecorder: ApolloCodegen.NonFatalError.Recorder
-  ) -> TemplateString {
+  ) -> TemplateString? {
     HeaderCommentTemplate.editableFileHeader(
       fileCanBeEditedTo: "implement advanced custom scalar functionality."
     )    


AnthonyMDev (approved):
Thanks so much for this bug fix! This test was missing and made this bug difficult to catch during my refactor.

calvincestari:
Do we need to be worried that this was more than just a comment-related bug? I'd be far more concerned if we were actually overwriting pre-existing custom scalar files; do we need an additional test for that?

AnthonyMDev:
No, this wouldn't have affected that. It was just a change in the function signature for the functions:

```swift
func renderHeaderTemplate(
     nonFatalErrorRecorder: ApolloCodegen.NonFatalError.Recorder
   ) -> TemplateString? {

func renderDetachedTemplate(
    nonFatalErrorRecorder: ApolloCodegen.NonFatalError.Recorder
  ) -> TemplateString? {
    nil
  }

and

func renderHeaderTemplate(
     nonFatalErrorRecorder: ApolloCodegen.NonFatalError.Recorder
   ) -> TemplateString? {

calvincestari:
> Do we need to be worried that this was more than just a comment-related bug? I'd be far more concerned if we were actually overwriting pre-existing custom scalar files; do we need an additional test for that?

Executing the `run-codegen.sh` script on the `1.8.0` tagged release doesn't change any of the generated files and we have generated custom scalars in both [AnimalKingdomAPI](https://github.com/apollographql/apollo-ios-dev/tree/main/Sources/AnimalKingdomAPI/AnimalKingdomAPI/Sources/Schema/CustomScalars) and [GitHubAPI](https://github.com/apollographql/apollo-ios-dev/tree/main/Sources/GitHubAPI/GitHubAPI/Sources/Schema/CustomScalars).

I've had a look through `FileGeneratorTests` and `ApolloCodegenTests` but I don't think we have a test that actually checks written file output to ensure nothing is written to pre-existing files that should not be overwritten.

calvincestari:
> No, this wouldn't have affected that. It was just a change in the function signature for the functions:

Sorry. I had my reply in draft for a while and didn't see your reply before posting it.

PR #250, opened 2024-01-26
[📚docs] fix typo in file-uploads.mdx

martinbonnin:
I don't think there's such a thing as `multipart-form`? ([Matching apollo-kotlin PR](https://github.com/apollographql/apollo-kotlin/pull/5566))

Also heads up that the router is prepping for file uploads (see https://github.com/apollographql/router/issues/4496). So we should probably remove this sentence at some point in the future.


calvincestari (approved):

calvincestari:
Thanks @martinbonnin and yes we can remove it when the router begins preview support for it.

PR #251, opened 2024-01-26
Add privacy manifest for Apollo targets

BobaFetters:
Closes apollographql/apollo-ios#3328

### DIFF
--- a/apollo-ios/Apollo.podspec
+++ b/apollo-ios/Apollo.podspec
@@ -22,6 +22,7 @@ Pod::Spec.new do |s|
 
   s.subspec 'Core' do |ss|
     ss.source_files = 'Sources/Apollo/**/*.swift','Sources/ApolloAPI/**/*.swift'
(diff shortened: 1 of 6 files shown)


calvincestari:
I think we may need to add that file extension to the podspec too - [resource_bundles](https://guides.cocoapods.org/syntax/podspec.html#resource_bundles)

calvincestari (commented):
This looks good @BobaFetters. I don't think there is a way for us to test this except for installing with each dependency manager and then manually inspecting that the privacy manifest file exists? That seems a bit much for something like this that isn't likely to change - thoughts?

Also, the pod spec lint error we were getting earlier is still present but I believe it's related to versioning and tags. Not sure if we need to worry about that.

BobaFetters:
@calvincestari Yea we would need to have projects for each dependency manager, archive them, and then inspect the archive I think to ensure the privacy manifest was in the archive. But no I'm not sure that's necessary for this honestly.

Maybe something we can look at fixing separately if its not directly related to these changes?

PR #252, opened 2024-01-30
[ApolloPagination] Rename classes, make `GraphQLQueryPager` a publisher

Iron-Ham:
This pull request aims to do the following:

1. Rename `GraphQLQueryPager` and `AsyncGraphQLQueryPager` to `GraphQLQueryPagerCoordinator` and `AsyncGraphQLQueryPagerCoordinator`. These classes have been marked `internal`. 
2. Rename `AnyGraphQLQueryPager` to `GraphQLQueryPager`. Similarly, `AnyAsyncGraphQLQueryPager` has been renamed to `AsyncGraphQLQueryPager`. 
3. Both `GraphQLQueryPager` and `AsyncGraphQLQueryPager` now conform to `Combine.Publisher`. They do so by forwarding their existing publishers. 
4. Convenience initializers have been routed to `GraphQLQueryPager` and `AsyncGraphQLQueryPager`. They're set up such that if no `transform` is provided, the output is `Result<(PaginationOutput<InitialQuery, PaginatedQuery>, UpdateSource), Error>`. 
5. `PaginationOutput`'s `updateSource` variable has been marked internal-only in the hopes that it reduces confusion, as the `transform`-less case would then include the `updateSource` both at the top-level and within the `PaginationOutput`. 
6. The `eraseToAnyPager(...)` functions have been moved to the test suite.


Iron-Ham:
cc: @calvincestari – I've not delved too deeply into creating custom publishers. At present, i just have them forwarding their internal publisher. Would love to grab some early eyes on that as a gut check. If that's sufficient – great!

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPagerOutput.swift
```diff
@@ -14,7 +14,7 @@ public struct PaginationOutput<InitialQuery: GraphQLQuery, PaginatedQuery: Graph
   public let nextPages: [PaginatedQuery.Data]
 
   /// The source of the most recent `Output`: either from the cache or server.
-  public let updateSource: UpdateSource
```
AnthonyMDev, line 17: Since it's become a bit awkward having this in two places, I'm wondering if we can just remove it from `PaginationOutput` completely? The only place it looks like it's being used right now is a few unit tests, and to pass the `updateSource` through to the `Output` result.

Could we just add the `updateSource` to the result value of the `GraphQLQueryPagerCoordinator` and then pass it through that way instead of having it as a stored property at all?

AnthonyMDev (approved):
One comment on the `updateSource`, otherwise, this looks great! What a huge improvement! Thanks @Iron-Ham!

AnthonyMDev (approved):
One comment on the `updateSource`, otherwise, this looks great! What a huge improvement! Thanks @Iron-Ham!

AnthonyMDev (approved):

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPagerOutput.swift
Iron-Ham: Totally possible.

Iron-Ham:
@AnthonyMDev not within scope of this PR, but what do you think about potentially performing a `fetch` on behalf of users who attempt to `loadPrevious`/`loadMore` without having previously loaded a first page? Currently, we throw an error – but perhaps we can detect that misuse, have an assertionFailure, but otherwise carry on?

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPagerOutput.swift
Iron-Ham: @AnthonyMDev done in https://github.com/apollographql/apollo-ios-dev/pull/252/commits/523255f01c7f00f4485dde6b8768297dfccd9378

AnthonyMDev (approved):

AnthonyMDev:
> @AnthonyMDev not within scope of this PR, but what do you think about potentially performing a `fetch` on behalf of users who attempt to `loadPrevious`/`loadMore` without having previously loaded a first page? Currently, we throw an error – but perhaps we can detect that misuse, have an assertionFailure, but otherwise carry on?

I'm not opposed to that. Agree that it sounds like something for a next PR though.

PR #253, opened 2024-02-01
Remove debug print statements for websocket status

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3325.

This removes the two `debugPrint` statements from websocket disconnection status changes. There really is no point to them since `debugPrint` only writes to stdout and in both cases the disconnection is sent to the delegate's handler or all websocket watchers.

There is another `debugPrint` statement being called in the [`DispatchQueue` extension](https://github.com/apollographql/apollo-ios-dev/blob/aa3b6fb370a5bb9c7cf1a05e853f71357cd8a842/apollo-ios/Sources/Apollo/DispatchQueue%2BOptional.swift#L25) but I'm not really sure how we should handle that one. My first thought was to change it to a `preconditionFailure` which would alert users that there is an error flow not being handled but the standard Release build (`-0`) will crash their app which is a breaking change. That function also doesn't throw so there really is nothing to be done other than swallowing the error; maybe we should just leave that one alone for now since at least there will be log message being printed somewhere?


BobaFetters (approved):
Looks good to me thanks!

Also yea I think the other one in the DispatchQueue extension we should probably leave alone for now.

PR #254, opened 2024-02-01
[ci]: Tooling update

calvincestari:
The release of [Tuist `4.0.0-beta-1`](https://github.com/tuist/tuist/releases) appears to have broken our workflow. It looks like [they're trying to fix the environment](https://github.com/tuist/tuist/pull/5858) but we should stay on 3.x until 4 is stable.

The aim of this PR was to fix Tuist which it does by:
* brings the installation into this repo in a script which we can control and with the installed version controlled by our pinned version of Tuist.
* bumps the Tuist pinned version to the latest 3.x release which is `3.42.2`.
* updates the README to use the local installation script for Tuist.

Other changes in this PR:
* bumps the [macos runner image to `macos-14`](https://github.blog/changelog/2024-01-30-github-actions-macos-14-sonoma-is-now-available/).
* bumps the xcode version to `15.2`.


.github/actions/run-tuist-generation/action.yml
```diff
@@ -6,5 +6,5 @@ runs:
     - name: Run Tuist
       shell: bash
       run: |
-        curl -Ls https://install.tuist.io | bash
-        tuist generate
\ No newline at end of file
+        ./scripts/install-tuist.sh
```
calvincestari, line 9: The local Tuist installation script now under our control.

.github/workflows/ci-tests.yml
```diff
@@ -36,11 +37,13 @@ jobs:
...
               - 'apollo-ios/**'
+            tuist:
+              - '.tuist-version'
 
   tuist-generation:
-    runs-on: macos-13
+    runs-on: macos-14
     needs: [changes]
-    if: ${{ needs.changes.outputs.ios == 'true' || needs.changes.outputs.codegen  == 'true' || needs.changes.outputs.pagination  == 'true' }}
+    if: ${{ needs.changes.outputs.ios == 'true' || needs.changes.outputs.codegen  == 'true' || needs.changes.outputs.pagination  == 'true' || needs.changes.outputs.tuist == 'true' }}
```
calvincestari, line 46: The tuist-generation job should run whenever we change the pinned Tuist version.

.github/workflows/ci-tests.yml
```diff
@@ -50,10 +53,7 @@ jobs:
     - name: Checkout Repo
       uses: actions/checkout@v3
     - name: Run Tuist Generation
-      uses: tuist/tuist-action@0.13.0
-      with:
-          command: 'generate'
-          arguments: ''
+      uses: ./.github/actions/run-tuist-generation
```
calvincestari, line 56: We already had a local action but it wasn't being used. This drops the broken tuist action from the marketplace and uses our own one that we can control.

scripts/install-tuist.sh
```diff
@@ -0,0 +1,59 @@
...
+ohai() {
+  printf "${tty_blue}==>${tty_bold} %s${tty_reset}\n" "$(shell_join "$@")"
+}
+
+warn() {
+  printf "${tty_red}Warning${tty_reset}: %s\n" "$(chomp "$1")"
+}
+
+# The line below extracts the version from the local .tuist-version file and installs that version tag from the github releases.
+LATEST_VERSION=$(cat .tuist-version)
```
calvincestari, line 24: The installed Tuist version will always match what we've pinned it to.

calvincestari: It's worth noting that this installation never actually installs Tuist itself but rather tuistenv which was their prior version manager, and what they're moving away from in favour of mise. If you look at the terminal output the actual Tuist version is installed when you issue the first command, such as `generate`.

BobaFetters (approved):
This looks great! Nice work!

PR #257, opened 2024-02-05
fix: Selection set initializer reserved parameter name

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3330.

Adds an underscored alias name to initializer parameters that would conflict with any Swift reserved names. This has only been an issue with `self` specifically within initializers but this enforces the same behaviour for any of the Swift reserved keywords.


AnthonyMDev (approved):
This is a great solution!

PR #258, opened 2024-02-05
chore: Update user survey conditions

calvincestari:
This skips the user survey comment when an issue is closed if the issue was opened by a maintainer and there were zero comments on the issue at the time. The parameters available through the `issue` event are detailed [here](https://docs.github.com/en/webhooks/webhook-events-and-payloads#issues).

Done to avoid issue survey requests like [this](https://github.com/apollographql/apollo-ios/issues/3328#issuecomment-1919687978).


calvincestari:
Turns out that the `author_association` property of an issue is affected by your organization members visibility setting.
* If you set it to `private` then you are simply seen as a `contributor` to the repo which [breaks this change](https://github.com/apollographql/apollo-ios/actions/runs/7850525011/job/21426066407).
* If you set it to `public` then you are seen as a `member` [which is what we want](https://github.com/apollographql/apollo-ios/actions/runs/7850968540/job/21427285863) for the updated condition to work.

[This was the tip-off](https://github.com/orgs/community/discussions/18690) to understanding the problem.

PR #260, opened 2024-02-09
[ApolloPagination] Update variable mapping logic to use a Set instead of an Array

Iron-Ham:
A special thank you to @ralph-101-dev for finding this issue in a pre-release build of `ApolloPagination`. 

Closes https://github.com/apollographql/apollo-ios/issues/3331

_____

This pull request primarily focuses on modifying how variables are handled in `AsyncGraphQLQueryPagerCoordinator.swift`. The changes involve altering the data types of variables from arrays to sets, which should eliminate the possibility of a mismatch between mis-ordered array outputs. This is especially needed, because the underlying values being pulled from are an unordered dictionary, so multiple invocations of the same function can result in a key-miss. 

The change to production code is a one-liner: We've updated the value of the `variables` variable – which houses the input variables of a given `GraphQLQuery` – from an `Array<JSONValue` to a `Set<JSONValue`. 

The tests were updated to unwrap the `AnyHashable` key-value of the `nextPageVarMap` property as a `Set` instead of as an `Array`. Notably, the tests were masking this error, as they were previously taking the `Array` outputs, wrapping them in a `Set`, and then comparing the sets!

### DIFF
--- a/Tests/ApolloPaginationTests/ForwardPaginationTests.swift
+++ b/Tests/ApolloPaginationTests/ForwardPaginationTests.swift
@@ -124,7 +124,7 @@ final class ForwardPaginationTests: XCTestCase, CacheDependentTesting {
     ]
 
     let expectedVariables = Set(nextQuery.__variables?.values.compactMap { $0._jsonEncodableValue?._jsonValue } ?? [])
-    let actualVariables = try await XCTUnwrapping(await pager.nextPageVarMap.keys.first as? Set<JSONValue>)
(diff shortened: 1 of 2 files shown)


AnthonyMDev:
> Notably, the tests were masking this error, as they were previously taking the Array outputs, wrapping them in a Set, and then comparing the sets!

🤣 Amazing

Iron-Ham:
@AnthonyMDev That sounds right – Updating this PR.

PR #261, opened 2024-02-09
fix: Type case with only reserved fields is not composite inline fragment

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3326.

* Adds a property to `TypeInfo` of `SelectionSet` that the root field builder uses to declaratively state whether a selection is composite or not.
* No longer generates empty `__selections` array into generated models. There is a [default implementation of an empty `__selections` property in an extension](https://github.com/apollographql/apollo-ios-dev/blob/3b8e5701cbf7a74953b0e63b1991b70cf3d64a90/apollo-ios/Sources/ApolloAPI/SelectionSet.swift#L87) that negates the need for us to do this in generated models.


Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
```diff
@@ -6952,6 +7003,53 @@ class SelectionSetTemplateTests: XCTestCase {
       .to(equalLineByLine(predator_expected, atLine: 21, ignoringExtraLines: true))
   }
 
+  func test__render_nestedSelectionSet__givenInlineFragmentWithOnlyReservedField_doesNotRenderCompositeInlineFragment() async throws {
```
AnthonyMDev, line 7006: Can we add the link to the issue here as well?
```suggestion
// Related to https://github.com/apollographql/apollo-ios/issues/3326
  func test__render_nestedSelectionSet__givenInlineFragmentWithOnlyReservedField_doesNotRenderAsCompositeInlineFragment() async throws {
```

Tests/CodegenIntegrationTests/Tests/3326-typeCaseWithOnlyReservedFieldSelectionIsNotCompositeInlineFragment/README.MD
```diff
@@ -0,0 +1,11 @@
+# Overview
+
+When a type case selection consisted of only GraphQL reserved fields, such as `__typename` the selection was not being recognized and a composite inline fragment was generated. This resulted in empty `__selections` and `__mergedSources` lists. This is turn caused the `_asInlineFragment` accessor to behave incorrectly because it checks whether the definitions contained within `__mergedSources` are a match for the returned type. The empty list always matches and `_asInlineFragment` incorrectly matches the returned type.
```
AnthonyMDev, line 3:
```suggestion
When a type case selection consisted of only GraphQL reserved fields, such as `__typename` the selection was not being recognized and a composite inline fragment was generated. This resulted in empty `__selections` and `__mergedSources` lists. This in turn caused the `_asInlineFragment` accessor to behave incorrectly because it checks whether the definitions contained within `__mergedSources` are a match for the returned type. The empty list always matches and `_asInlineFragment` incorrectly matches the returned type.
```

apollo-ios-codegen/Sources/IR/IR+SelectionSet.swift
```diff
@@ -15,6 +15,10 @@ public class SelectionSet: Hashable, CustomDebugStringConvertible {
     /// The selection set's `scope` is the last element in the list.
     public let scopePath: LinkedList<ScopeDescriptor>
 
+    public var isUserDefined: Bool
```
AnthonyMDev, line 18:
```suggestion
    /// Indicates if the `SelectionSet` was created directly due to a selection set in the user defined `.graphql` definition file.
    ///
    /// If `false`, the selection set was artificially created by the IR. Currently, the only reason for this is a `CompositeInlineFragment` created during calculation of merged selections for field merging.
    public var isUserDefined: Bool
```

AnthonyMDev (approved):
Looks good! Thanks for adding that integration test case!

Just a few minor suggestions.

PR #262, opened 2024-02-09
Add ApolloPagination to the DocC generation project

Iron-Ham:
This PR added `ApolloPagination` to the list of valid targets for the `DocumentationGenerator` command.  I added the `ApolloPagination` package as a dependency of the project.

**NOTE**: I am receiving the following warning –

```
warning: 'apollo-ios-pagination': 'apollo-ios-pagination' dependency on 'https://github.com/apollographql/apollo-ios.git' conflicts with dependency on '/Users/zer0/Developer/OSS/apollo-ios-dev/apollo-ios' which has the same identity 'apollo-ios'. this will be escalated to an error in future versions of SwiftPM.
```

For now, I'm not going to concern myself with the warning, and point additional changes (adding `.mdx` documentation) against this PR.


docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,127 @@
...
+---
+
+Apollo Pagination provides a convenient and easy way to interact with and watch paginated APIs. It provides a flexible and powerful way to interact with paginated data, and works with both cursor-based and offset-based pagination. Its key features include:
+
+- Watching paginated data
+- Forward, reverse, and bidirectional pagination support
+- Multi-query pagination support
+- Support for custom model types
+
+Apollo Pagination provides two classes to interact with paginated endpoints: `GraphQLQueryPager` and `AsyncGraphQLQueryPager`. They have very similar APIs, but the latter is designed for use in asynchronous contexts.
```
AnthonyMDev, line 12:
```suggestion
Apollo Pagination provides two classes to interact with paginated endpoints: `GraphQLQueryPager` and `AsyncGraphQLQueryPager`. They have very similar APIs, but the latter supports `async/await` for use in asynchronous contexts. 
```

docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,127 @@
...
+- Watching paginated data
+- Forward, reverse, and bidirectional pagination support
+- Multi-query pagination support
+- Support for custom model types
+
+Apollo Pagination provides two classes to interact with paginated endpoints: `GraphQLQueryPager` and `AsyncGraphQLQueryPager`. They have very similar APIs, but the latter is designed for use in asynchronous contexts. 
+
+## Using a `GraphQLQueryPager`
+
+The `GraphQLQueryPager` class is intended to be a simple, flexible, and powerful way to interact with paginated data. While it has a standard initializer, it is recommended to use the convenience initializers, which simplify the process of creating a `GraphQLQueryPager` instance. The convenience initializers are defined as static methods on the `GraphQLQueryPager` class, prefixed with `make`.
```
AnthonyMDev, line 16:
```suggestion
The `GraphQLQueryPager` class is intended to be a simple, flexible, and powerful way to interact with paginated data. While it has a standard initializer, it is recommended to use the convenience initializers, which simplify the process of creating a `GraphQLQueryPager` instance. The convenience initializers are static methods on the `GraphQLQueryPager` class, prefixed with `make`.
```

docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,127 @@
...
+            endCursor: data.values.pageInfo.endCursor
+        )
+    },
+    transform: { data in 
+        data.values.nodes.map { Person(id: $0.id, name: $0.name) }
+    }
+)
+```
+
+In this example, the `GraphQLQueryPager` instance is initialized with a `transform` closure that transforms the raw response into an array of `Person` objects. Alternatively, we can make use of the `GraphQLQueryPager`'s `map` method to achieve the same result:
```
AnthonyMDev, line 98: I think we need to make it clear that this `transform` will be called on the `data` for each page query's response sequentially and then the arrays will be concatenated together.

And alternatively, you can use the function with the `transform` signature `([PaginatedQuery.Data], InitialQuery.Data, [PaginatedQuery.Data]) throws -> Model` to get all of the page data at once and manually build whatever data structure you want. (With a quick example of what this looks like as well.)

docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,127 @@
...
+        // Handle the error
+    }
+}
+```
+
+## Fetching Data
+
+The `GraphQLQueryPager` class provides several methods to fetch paginated data: `fetch`, `refetch`, `loadNext`, `loadPrevious`, and `loadAll`. 
+
+The `fetch` method is used to fetch the first page of data and must be called before `loadNext` or `loadPrevious` can be called. The `refetch` method is used to refetch the first page of data, and additionally resets the pager to its initial state. The `loadNext` method is used to fetch the next page of data, and the `loadPrevious` method is used to fetch the previous page of data. The `loadAll` method is used to fetch all pages of data.
```
AnthonyMDev, line 117: Explain that `loadAll` will continue to fetch pages until the pageInfo returns `hasNext = false`.

docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+## Fetching Data
+
+The `GraphQLQueryPager` class provides several methods to fetch paginated data: `fetch`, `refetch`, `loadNext`, `loadPrevious`, and `loadAll`. 
+
+The `fetch` method is used to fetch the first page of data and must be called before `loadNext` or `loadPrevious` can be called. The `refetch` method is used to refetch the first page of data, and additionally resets the pager to its initial state. The `loadNext` method is used to fetch the next page of data, and the `loadPrevious` method is used to fetch the previous page of data. The `loadAll` method is used to fetch all pages of data. 
+
+## Cancelling a Pager
+
+The `GraphQLQueryPager` class provides a `cancel` method, which can be used to cancel any ongoing fetch operations as well as cease any watcher subscriptions. Once cancelled, the pager's state is reset. Any subscriber that was subscribed to the pager will remain subscribed and will continue to receive updates.
```
AnthonyMDev, line 121: > Any subscriber that was subscribed to the pager will remain subscribed and will continue to receive updates.

If the watchers have been canceled, what updates would they receive. This is unclear to me.

I wonder if on `cancel`, we should send a termination signal to the subscribers?

docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,127 @@
...
+
+The `fetch` method is used to fetch the first page of data and must be called before `loadNext` or `loadPrevious` can be called. The `refetch` method is used to refetch the first page of data, and additionally resets the pager to its initial state. The `loadNext` method is used to fetch the next page of data, and the `loadPrevious` method is used to fetch the previous page of data. The `loadAll` method is used to fetch all pages of data. 
+
+## Cancelling a Pager
+
+The `GraphQLQueryPager` class provides a `cancel` method, which can be used to cancel any ongoing fetch operations as well as cease any watcher subscriptions. Once cancelled, the pager's state is reset. Any subscriber that was subscribed to the pager will remain subscribed and will continue to receive updates. 
+
+## Error handling
+
+Network errors are exposed to the caller via the `Result` type. Usage errors are thrown as `Error` types or exposed as callbacks in each fetch method.
```
AnthonyMDev, line 125: I know what you are talking about here because we discussed it. But I think this needs a little more detail.

Clarity on the difference between network errors and usage errors as well as maybe a code snippet showing where they show up.

AnthonyMDev (changes requested):
This is an awesome start! I've got a few minor suggestions and questions.

I also want to start a running list of things to be added to documentation (can be done in the future):
[] Example of Reverse/Bidirectional Pagination
[] Example of Offset based pagination
[] Example of using different queries for initial query vs additional pages

Iron-Ham:
Happy to address these – @AnthonyMDev do you have advice for running this locally, so I can see how the changes would present on the web? I wasn't able to get it running locally, which certainly made things a little tougher.

AnthonyMDev:
Our docs deployment repo is actually public, so you can clone it, and then use the instructions here to run the docs locally.
https://github.com/apollographql/docs?tab=readme-ov-file#developing-a-single-docset

Iron-Ham:
cc: @AnthonyMDev I've gone ahead and added 3 more pages & added a single example of an offset pager within the Introduction.

docs/source/pagination/introduction.mdx
AnthonyMDev: Hmmm I see. In FRP, I think the most common pattern is that a `cancel()` should terminate the publisher and dispose of all subscriptions. I'm concerned that deviating from that might cause confusion.

For the case of `refetch`, we don't want to do that of course, but in my mind, that means `refetch` should be calling a private function that cancels the internal network calls and existing watchers without terminating the publisher stream. `refetch` is managing some internal state to begin a refetch, but the subscribers will continue to get events because the pager itself isn't cancelled. While `cancel` should actually cancel the pager.

Especially considering the pager itself actually conforms to `Publisher`, I think `cancel` should do what you expect cancelling an FRP stream would do.

docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,129 @@
...
+Apollo Pagination provides two classes to interact with paginated endpoints: `GraphQLQueryPager` and `AsyncGraphQLQueryPager`. They have very similar APIs, but the latter supports `async`/`await` for use in asynchronous contexts.
+
+## Using a `GraphQLQueryPager`
+
+The `GraphQLQueryPager` class is intended to be a simple, flexible, and powerful way to interact with paginated data. While it has a standard initializer, it is recommended to use the convenience initializers, which simplify the process of creating a `GraphQLQueryPager` instance. The convenience initializers are static methods on the `GraphQLQueryPager` class, prefixed with `make`.
+
+For example, in order to initialize a `GraphQLQueryPager` that paginates a single query in the forward direction with a cursor-based pagination, you can use the `makeForwardCursorQueryPager` method:
+
+```swift
+let pager = GraphQLQueryPager.makeForwardCursorQueryPager(
```
AnthonyMDev, line 21: We'll need to update this with the new functions.

docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,129 @@
...
+    }
+)
+```
+
+In this example, the `GraphQLQueryPager` instance is initialized with a closure that provides the query to be executed, and another closure that extracts the pagination information from the query result. The `GraphQLQueryPager` instance can then be used to fetch the paginated data, and to watch for changes to the paginated data.
+
+Alternatively, we can create a forward-paginating offset-based `GraphQLQueryPager` instance using the `makeForwardOffsetQueryPager` method:
+
+```swift
+let pager = GraphQLQueryPager.makeForwardOffsetQueryPager(
```
AnthonyMDev, line 40: And this

docs/source/pagination/directional-pagers.mdx
```diff
@@ -0,0 +1,105 @@
+---
+title: Forward, Reverse, and Bi-directional Pagination
```
AnthonyMDev, line 2: Bidirectional doesn't need to be hyphenated.
```suggestion
title: Forward, Reverse, and Bidirectional Pagination
```

docs/source/config.json
```diff
@@ -111,6 +112,15 @@
       },
       true
     ],
+    "Pagination": [
+      {
+        "Introduction": "/pagination/introduction",
+        "Using Custom Response Models": "/pagination/custom-types",
+        "Forward, Reverse, and Bi-directional Pagination": "/pagination/directional-pagers",
```
AnthonyMDev, line 119: This is going to look really long in the side navigation bar.

Since forward pagination is pretty straightforward and explained in the intro page, I think maybe we could change this to "Reverse Pagination" or "Directional Pagination"? Not sure what makes the most sense.

AnthonyMDev (changes requested):

docs/source/pagination/introduction.mdx
Iron-Ham: Hmm...

I wonder how we would want to disambiguate this, and whether this is worth doing prior to first public release. What are your thoughts here?

AnthonyMDev: I mean, this is an 0.1 release, so we are allowed to make breaking changes. But this is a pretty significantly breaking change...

I don't think it would be too hard to disambiguate though. Let's chat about this one.

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager+Convenience.swift
```diff
@@ -27,7 +27,7 @@ public extension GraphQLQueryPager {
     client: ApolloClientProtocol,
     watcherDispatchQueue: DispatchQueue = .main,
     initialQuery: InitialQuery,
-    pageResolver: @escaping (P, PaginationDirection) -> InitialQuery,
+    pageResolver: @escaping (P, PaginationDirection) -> InitialQuery?,
```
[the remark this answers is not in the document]
AnthonyMDev: Nah, that's fine!

docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+For example, in order to initialize a `GraphQLQueryPager` that paginates a single query in the forward direction with a cursor-based pagination, you can use the `makeQueryPager` method:
+
+```swift
+let initialQuery = MyQuery(first: 10, after: nil)
+let pager = GraphQLQueryPager.makeQueryPager(
+    client: client,
+    initialQuery: initialQuery,
+    pageResolver: { page, paginationDirection in
+        // As we only want to support forward pagination, we can return `nil` for reverse pagination
```
AnthonyMDev, line 26: For the convenience functions that only support single direction paging, can we just omit the pagination direction from the `pageResolver` closure signature? It should only be needed for bidirectional paging, right?

docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,123 @@
...
+
+## Fetching Data
+
+The `GraphQLQueryPager` class provides several methods to fetch paginated data: `fetch`, `refetch`, `loadNext`, `loadPrevious`, and `loadAll`.
+
+`fetch`: Fetches the first page of data. Must be called before `loadNext` or `loadPrevious` can be called.
+`refetch`: Cancels all ongoing fetch operations and resets the pager to its initial state. Fetches the first page of data.
+`loadNext`: Fetches the next page of data. Can only be called after `fetch` has been called.
+`loadPrevious`: Fetches the previous page of data. Can only be called after `fetch` has been called.
+`loadAll`: Fetches all pages of data. If no initial page is detected, it will first call `fetch` to fetch the first page of data. Will continue to fethc all pages until a `PageInfo` object indicates that there are no more pages to fetch. This function is compatible with forward, reverse, and bidirectional pagination.
```
AnthonyMDev, line 85: I think these need to be bulleted or else they will all appear as one paragraph. Also had a typo!

```suggestion
- `fetch`: Fetches the first page of data. Must be called before `loadNext` or `loadPrevious` can be called.
- `refetch`: Cancels all ongoing fetch operations and resets the pager to its initial state. Fetches the first page of data.
- `loadNext`: Fetches the next page of data. Can only be called after `fetch` has been called.
- `loadPrevious`: Fetches the previous page of data. Can only be called after `fetch` has been called.
- `loadAll`: Fetches all pages of data. If no initial page is detected, it will first call `fetch` to fetch the first page of data. Will continue to fetch all pages until a `PageInfo` object indicates that there are no more pages to fetch. This function is compatible with forward, reverse, and bidirectional pagination.
```

docs/source/pagination/custom-types.mdx
```diff
@@ -0,0 +1,126 @@
+---
+title: Using custom response models
+---
+
+The `GraphQLQueryPager` supports not only returning collections of a query's `Data` type, but also custom response types. The custom response types can either be a collection or a single object. This is useful when you want to transform the raw response into custom models for consumption in your application.
+
+## Transforming existing responses types
+
+The `GraphQLQueryPager` class conforms to the `Publisher` protocol, and provides a `map` method that allows you to transform the raw response into a custom response type. When used this way, the `GraphQLQueryPager` will operate over generated `Data` types, but allow each `Data` type to betransformed into a custom response type. This is especially useful when you want to reuse a single pager in multiple places, transforming the raw data into view models.
```
AnthonyMDev, line 9:
```suggestion
The `GraphQLQueryPager` class conforms to the `Publisher` protocol, and provides a `map` method that allows you to transform the raw response into a custom response type. When used this way, the `GraphQLQueryPager` will operate over generated `Data` types, but allow each `Data` type to be transformed into a custom response type. This is especially useful when you want to reuse a single pager in multiple places, transforming the raw data into view models.
```

docs/source/pagination/custom-types.mdx
```diff
@@ -0,0 +1,126 @@
...
+}
+```
+
+## Creating `GraphQLQueryPager` that returns custom response types
+
+The `GraphQLQueryPager` class can also be initialized with a custom type in mind. When used this way, the `GraphQLQueryPager` will operate over the custom response type, and not the raw `Data` type. This is useful when you want to create a pager that returns a custom response type, and you don't want to transform the raw response into a custom response type.
+
+We provide a convenience method for transforming collection types in this manner. 
+
+In this example, we take a `GraphQLQueryPager` that returns a `Result<PaginationOutput<MyQuery.Data, MyQuery.Data>>`, and transform the raw response into a custom response type of `Result<([Person], UpdateSource), Error>`.
```
AnthonyMDev, line 66:
```suggestion
In this example, we create a `GraphQLQueryPager` that  transform the raw response into a custom response type of `Result<([Person], UpdateSource), Error>`. The `transform` closure will be called on the data from each paginated query and the resulting arrays will be concatenated together.
```

docs/source/pagination/custom-types.mdx
```diff
@@ -0,0 +1,126 @@
...
+    },
+    transform: { data in
+        data.values.nodes.compactMap { node in
+            Person(id: node.id, name: node.name)
+        }
+    }
+)
+```
+
+This supports single object types as well, such as a `Result<(MyResponseModel, UpdateSource), Error>`:
```
AnthonyMDev, line 96:
```suggestion
Alternatively, you can also return a single model type, such as a `Result<(MyResponseModel, UpdateSource), Error>`. To do this, we use a `transform` closure with three parameters, `previousPages, initialPage, nextPages`:
```

docs/source/pagination/directional-pagers.mdx
```diff
@@ -0,0 +1,158 @@
...
+  { extractionData in
+    switch extractionData {
+    case .initial(let value, _), .paginated(let value, _):
+      return transform(value)
+    }
+  }
+}
+
+private func pageExtraction<InitialQuery: GraphQLQuery, PaginatedQuery: GraphQLQuery, P: PaginationInfo, T>(
+  initialTransfom: @escaping (InitialQuery.Data) -> P,
```
AnthonyMDev, line 146:
```suggestion
  initialTransform: @escaping (InitialQuery.Data) -> P,
```

docs/source/pagination/directional-pagers.mdx
```diff
@@ -0,0 +1,158 @@
+---
+title: Directional Pagination
+---
+
+`GraphQLQueryPager` supports pagination in both the forward and reverse direction, as well as both at once. 
+
+## Forward Pagination
+
+Forward pagination is the most common form of pagination. It is used to fetch the next `n` items in a list. We can use the convenience `make` functions to create a configured `GraphQLQueryPager`. While we have many options depending on our requirements -- whether we use one query or two, whether we want to use a cursor or an offset, whether we want to transform the results, etc. -- we will examine using a cursor with a single query.
```
AnthonyMDev, line 9: Let's talk about my question related to omitting `paginationDirection` in single directional convenience functions. If we can go that route, then I have some suggestions to tweak this.

docs/source/pagination/multi-query.mdx
```diff
@@ -0,0 +1,34 @@
+---
+title: Multi-Query Pagination
+---
+
+`GraphQLQueryPager` supports multi-query pagination. This means that you can use separate queries for fetching the initial page of data and for fetching subsequent pages of data. 
+
+In order to use multi-query pagination, you must configure the `GraphQLQueryPager` in order to support multiple queries. It is recommended to use an appropriate convenience `make` function to create the `GraphQLQueryPager` instance. In the following example, we will use the `makeQueryPager` function to create a `GraphQLQueryPager` that uses forward cursor-based pagination.
```
AnthonyMDev, line 7:
```suggestion
In order to use multi-query pagination, you must configure the `GraphQLQueryPager` in order to support multiple queries. It is recommended to use an appropriate convenience `make` function to create the `GraphQLQueryPager` instance. The convenience `make` functions that support multi-query pagination will have both an `extractInitialPageInfo` and `extractNextPageInfo` function.

In the following example, we will use the `makeQueryPager` function to create a `GraphQLQueryPager` that uses forward cursor-based pagination.
```

docs/source/pagination/multi-query.mdx
```diff
@@ -0,0 +1,34 @@
...
+let initialQuery = MyQuery(first: 10, after: nil)
+let pager = GraphQLQueryPager.makeQueryPager(
+    client: client,
+    initialQuery: initialQuery,
+    extractInitialPageInfo: { data in
+        // Extract a `CursorBasedPagination.Forward` instance from the initial query's `Data`
+    },
+    extractNextPageInfo: { data in
+        // Extract a `CursorBasedPagination.Forward` instance from the paginated query's `Data`
+    },
```
AnthonyMDev, line 19:
```suggestion
    extractInitialPageInfo: { initialQueryData in
        // Extract a `CursorBasedPagination.Forward` instance from the initial query's `Data`
        CursorBasedPagination.Forward(
             hasNext: initialQueryData.values.pageInfo.hasNextPage ?? false,
             endCursor: initialQueryData.values.pageInfo.endCursor
         )
    },
    extractNextPageInfo: { paginatedQueryData in
        // Extract a `CursorBasedPagination.Forward` instance from the paginated query's `Data`
        CursorBasedPagination.Forward(
             hasNext: paginatedQueryData.values.pageInfo.hasNextPage ?? false,
             endCursor: paginatedQueryData.values.pageInfo.endCursor
         )
    },
```

docs/source/pagination/multi-query.mdx
```diff
@@ -0,0 +1,34 @@
...
+        case .previous:
+            return nil
+        }
+    }
+)
+```
+
+This example demonstrates how to create a `GraphQLQueryPager` that uses separate queries for fetching the initial page of data and for fetching subsequent pages of data. The `GraphQLQueryPager` is configured to use forward cursor-based pagination. The `extractInitialPageInfo` and `extractNextPageInfo` closures are used to extract pagination information from the initial page of data and from the next page of data, respectively. The `nextPageResolver` closure is used to resolve the next page query given a `CursorBasedPagination.Forward` instance.
+
+The `GraphQLQueryPager` instance can be used in the same way as a single-query pager. The `loadNextPage` method will automatically use the `nextPageResolver` closure to fetch the next page of data.
```
AnthonyMDev, line 34:
```suggestion
This example demonstrates how to create a `GraphQLQueryPager` that uses separate queries for fetching the initial page of data and for fetching subsequent pages of data. The `GraphQLQueryPager` is configured to use forward cursor-based pagination. The `extractInitialPageInfo` and `extractNextPageInfo` closures are used to extract pagination information from the initial page of data and from the next page of data, respectively. The `pageResolver` closure is used to resolve the next page query given a `CursorBasedPagination.Forward` instance.

The `GraphQLQueryPager` instance can be used in the same way as a single-query pager. The `loadNextPage` method will automatically use the `pageResolver` closure to fetch the next page of data.
```

AnthonyMDev (commented):
Wow, reading through these docs, this library is just so incredibly feature rich already! What an awesome job @Iron-Ham!

Couple of edits to add some more information and minor fixes!

docs/source/pagination/introduction.mdx
Iron-Ham: Discussed this on our call just now – but this can be an improvement for 0.2.x

We could do it by declaring `ForwardPageInfo` and `ReversePageInfo` protocols, and then declaring the convenience functions around those protocols.

apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
```diff
@@ -237,6 +237,8 @@ public class AsyncGraphQLQueryPager<Model>: Publisher {
   /// Resets pagination state and cancels further updates from the pager.
   public func cancel() async {
     await pager.cancel()
+    _subject.send(completion: .finished)
```
AnthonyMDev, line 240: Do we need to update unit tests for this? Can also be done in a separate PR later.

Iron-Ham: I think that'd be wise, yes. 
As written, I think this would require us to re-init a new pager – in the same sense that cancelling a `watcher` stops it from being used again.

docs/source/pagination/introduction.mdx
```diff
@@ -0,0 +1,125 @@
...
+- `refetch`: Cancels all ongoing fetch operations and resets the pager to its initial state. Fetches the first page of data.
+- `loadNext`: Fetches the next page of data. Can only be called after `fetch` has been called.
+- `loadPrevious`: Fetches the previous page of data. Can only be called after `fetch` has been called.
+- `loadAll`: Fetches all pages of data. If no initial page is detected, it will first call `fetch` to fetch the first page of data. Will continue to fetch all pages until a `PageInfo` object indicates that there are no more pages to fetch. This function is compatible with forward, reverse, and bidirectional pagination.
+
+## Cancelling ongoing requests
+
+The `GraphQLQueryPager` class provides a `cancel` method, which can be used to cancel any ongoing fetch operations as well as cease all subscriptions. Once cancelled, the pager's state is reset. Any subscriber that was subscribed to the pager will receive a termination signal, and will no longer receive updates. After calling the `cancel` method, you msut resubscribe to the pager in order to receive updates.
```
AnthonyMDev:
```suggestion
The `GraphQLQueryPager` class provides a `cancel` method, which can be used to cancel any ongoing fetch operations as well as cease all subscriptions. Once cancelled, the pager's state is reset. Any subscriber that was subscribed to the pager will receive a completion signal, and will no longer receive updates. After calling the `cancel` method the pager is invalid. You will need to create a new pager in order to receive updates again.
```

AnthonyMDev (approved):

AnthonyMDev (approved):

PR #265, opened 2024-02-12
[ApolloPagination] Fix memory leak in `GraphQLQueryPager`'s `Publisher` implementation

Iron-Ham:
We've detected a memory leak in the implementation of the `Publisher` protocol within `GraphQLQueryPager`. 

Instead of treating this as a custom publisher – which the previous implementation certainly did – we just pare it down and forward to the internal `publisher` property.

A special thank you to @esahmed21 for confirming the leak and making note that it didn't exist in the previous version.

> [!NOTE]
> **Why am I electing to call `subscribe(_:)` instead of `receive(_:)`?**
>
> `subscribe(_:)` in Combine has some additional observation logic that gets triggered before it forwards a subscription, which will ultimately invoke `receive(subscription:)` directly. That logic is important, as it prevents memory cycles & other unexpected outcomes. Apple's Combine docs include the incredible line: `/// Always call this function instead of Publisher/receive(subscriber:)`

### DIFF
--- a/apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
+++ b/apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
@@ -171,27 +171,6 @@ public class AsyncGraphQLQueryPager<Model>: Publisher {
   public func receive<S>(
     subscriber: S
   ) where S: Subscriber, Never == S.Failure, Result<(Model, UpdateSource), Error> == S.Input {
-    let subscription = PagerSubscription(pager: self, subscriber: subscriber)
-    subscriber.receive(subscription: subscription)
-  }
-}
-
-private class PagerSubscription<SubscriberType: Subscriber, Pager: AsyncGraphQLQueryPager<Model>, Model>: Subscription where SubscriberType.Input == Pager.Output {
-  private var subscriber: SubscriberType?
-  private var pager: Pager
-  private var cancellable: AnyCancellable?
-
-  init(pager: Pager, subscriber: SubscriberType) {
-    self.subscriber = subscriber
-    self.pager = pager
-    cancellable = pager.publisher.sink(receiveValue: {
-      _ = subscriber.receive($0)
-    })
-  }
-
-  func request(_ demand: Subscribers.Demand) { }
-
-  func cancel() {
-    subscriber = nil
+    publisher.subscribe(subscriber)
   }
 }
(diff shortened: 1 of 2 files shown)


calvincestari (approved):
I think this looks good and the changes make sense. Is there any way to validate this in testing vs. manual memory tests?

Iron-Ham:
@calvincestari I'm struggling to think of a way, as `Publisher` has no public property that details its subscribers. 
The net impact of the previous code was that pager's `deinit` was never called when a `UIViewController` gets deinitialized. 

My first thought was that we could test this by:
- Declaring a `UIViewController` that owns a `GraphQLQueryPager`. The view controller subscribes to results.
- Deinitialize the `UIViewController`
- Ensure that the `GraphQLQueryPager` gets deinitialized

The problem is that we can't store an external reference to that pager, or we'll tick its reference count up and it won't deinit. 😬 

I'll noodle on it some more.

calvincestari:
> I'm struggling to think of a way, as Publisher has no public property that details its subscribers.

I agree, it's a tricky one. My comment was a question of whether it's even possible, rather than "I think we should add a test".

Iron-Ham:
> > I'm struggling to think of a way, as Publisher has no public property that details its subscribers.
> 
> I agree, it's a tricky one. My comment was a question of whether it's even possible, rather than "I think we should add a test".

I hear ya – but I really _do_ wonder if it's possible 🤔

AnthonyMDev:
You can test this by storing it in a weak property, then setting the original reference to it to nil and checking that the weak var is now also nil.

Iron-Ham:
> You can test this by storing it in a weak property, then setting the original reference to it to nil and checking that the weak var is now also nil.

I've tried something like this:

```swift
  #if os(iOS)
  func test_deinitCalled_combinePublisher() {
    var vc: TestViewController? = TestViewController(client: client)
    weak var pager = vc?.pager
    XCTAssertNotNil(vc?.pager)
    vc = nil
    XCTAssertNil(pager)
  }
  #endif
```

But this seems to return a pass-result with both implementations.

AnthonyMDev (approved):
Always happy to delete unnecessary code!

DanielRay1982:
Can someone get me the fucking email to Wiz i am the person who found the fucked up money thats being returned

DanielRay1982:
Like i dont understand any of this bullshit and its really hard trying to figure the all out. Nothing about the shit is simple i have a 8th grade education a fucking simple email support would be greay for the shit to be added on here

PR #276, opened 2024-02-17
Schema type renaming

BobaFetters:
Adds the ability to customize names of schema types using the configurations options added in #242


BobaFetters:
Updated this draft with the initial pass at the new IR types to represent the schema types, all changes are under the `IR` source folder, everything else is left over from previous iteration that will be removed/changed

apollo-ios-codegen/Sources/ApolloCodegenLib/ApolloCodegen.swift
BobaFetters, line 1: Most of the changes in this file are currently old or unfinished

apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLNamedType+NameFormatting.swift
BobaFetters, line 1: This file should no longer be needed once this work is complete

apollo-ios-codegen/Sources/GraphQLCompiler/GraphQLSchema.swift
BobaFetters, line 1: Still need to undo the old changes to this file

apollo-ios-codegen/Sources/IR/SchemaTypes/IR+Name.swift
```diff
@@ -0,0 +1,39 @@
+import Foundation
+
+public protocol NamedItem {
```
BobaFetters, line 3: Open to different naming for this protocol.

apollo-ios-codegen/Sources/IR/IR+EntitySelectionTree.swift
```diff
@@ -224,6 +224,23 @@ class EntitySelectionTree {
           targetSelections.mergeIn(scopeSelections, from: source)
         }
 
+        if let conditionalScopes = scopeConditions {
```
BobaFetters, line 227: Not sure why these changes are here? Not related to the work I am doing

BobaFetters:
Have this updated with what _should_ be the final implementation for handling schema type renaming after going through a few iterations. Moving on to implementing tests now and will see from there if any further tweaks are needed.

Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
```diff
@@ -484,4 +489,36 @@ class EnumTemplateTests: XCTestCase {
...
+      ]
+    )
+
+    let expected = """
+    \(subject.graphqlEnum.name.typeNameDocumentation)
+    enum MyCustomEnum: String, EnumType {
+      case caseOne = "caseOne"
+      // This type has been renamed from the schema type 'myCase'
+      // using schema customization configuration.
+      case myCustomCase = "myCase"
```
AnthonyMDev, line 512: I don't know if "using schema customization configuration" is going to be clear to users. I'm also not sure if it's necessary that we actually explain _why_ it's renamed, just that it has been. What do you think?

What about:

```suggestion
      // Renamed from GraphQL schema value: 'myCase'
      case myCustomCase = "myCase"
```

Tests/ApolloCodegenTests/CodeGeneration/Templates/InputObjectTemplateTests.swift
```diff
@@ -2540,4 +2545,67 @@ class InputObjectTemplateTests: XCTestCase {
...
+        get { __data["fieldOne"] }
+        set { __data["fieldOne"] = newValue }
+      }
+
+      // This type has been renamed from the schema type 'myField'
+      // using schema customization configuration.
+      public var myCustomField: String {
+        get { __data["myField"] }
+        set { __data["myField"] = newValue }
+      }
```
AnthonyMDev, line 2600: Same opinion here:

```suggestion
      // Renamed from GraphQL schema value: 'myField'      
      public var myCustomField: String {
        get { __data["myField"] }
        set { __data["myField"] = newValue }
      }
```

AnthonyMDev: You're also using the generic message "This _type_ has been renamed", but enum cases and fields on input objects aren't "types", which is why I just changed it to the generic term "value".

Tests/ApolloCodegenTests/CodeGeneration/Templates/InterfaceTemplateTests.swift
```diff
@@ -135,4 +139,25 @@ class InterfaceTemplateTests: XCTestCase {
...
+  
+  func test__render__givenInterface_withCustomName_shouldRenderWithCustomName() throws {
+    // given
+    buildSubject(
+      name: "MyInterface",
+      customName: "MyCustomInterface"
+    )
+    
+    let expected = """
+    \(subject.graphqlInterface.name.typeNameDocumentation)
```
AnthonyMDev, line 152: We shouldn't be using interpolations in the test expectations.

Tests/TestCodeGenConfigurations/SchemaCustomization/Package.swift
```diff
@@ -0,0 +1,37 @@
...
+    .package(name: "apollo-ios", path: "../../../apollo-ios"),
+    .package(name: "AnimalKingdomAPI", path: "./AnimalKingdomAPI")
+  ],
+  targets: [
+    .target(
+      name: "TestApp",
+      dependencies: [
+        .product(name: "Apollo", package: "apollo-ios"),
+        .product(name: "AnimalKingdomAPI", package: "AnimalKingdomAPI")
+      ]
```
AnthonyMDev, line 26: Since we are adding the `ExistentialAny` requirement in #379:

```suggestion
      name: "TestApp",
      dependencies: [
        .product(name: "Apollo", package: "apollo-ios"),
        .product(name: "AnimalKingdomAPI", package: "AnimalKingdomAPI")
      ],
      swiftSettings: [.enableUpcomingFeature("ExistentialAny")]
```

Tests/TestCodeGenConfigurations/SchemaCustomization/Package.swift
AnthonyMDev, line 1: I'm wondering if adding an entire new test configuration for schema customizations is the right approach. Don't we want to make sure that schema customizations work with _all_ of the different namespacing constraints of the different configurations?

I think it would be better to add schema customizations to the all of the existing configurations instead of creating a new one.

Tests/TestCodeGenConfigurations/SchemaCustomization/Package.swift
```diff
@@ -0,0 +1,37 @@
...
+    ),
+    .testTarget(
+      name: "SwiftPackageTests",
+      dependencies: [
+        .product(name: "Apollo", package: "apollo-ios"),
+        .product(name: "ApolloTestSupport", package: "apollo-ios"),
+        .product(name: "AnimalKingdomAPITestMocks", package: "AnimalKingdomAPI")
+      ]
+    )
```
AnthonyMDev, line 35:
```suggestion
    .testTarget(
      name: "SwiftPackageTests",
      dependencies: [
        .product(name: "Apollo", package: "apollo-ios"),
        .product(name: "ApolloTestSupport", package: "apollo-ios"),
        .product(name: "AnimalKingdomAPITestMocks", package: "AnimalKingdomAPI")
      ],
      swiftSettings: [.enableUpcomingFeature("ExistentialAny")]
    )
```

apollo-ios-codegen/Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -118,10 +118,12 @@ public class ApolloCodegen {
     try config.validateConfigValues()
 
     let compilationResult = try await compileGraphQLResult()
-
+
```
AnthonyMDev, line 121:
```suggestion

```

apollo-ios-codegen/Sources/ApolloCodegenLib/ApolloCodegen.swift
```diff
@@ -333,21 +401,21 @@ public class ApolloCodegen {
 
     nonFatalErrors.merge(
       try await nonFatalErrorCollectingTaskGroup() { group in
-        for graphQLObject in ir.schema.referencedTypes.objects {
+        for graphqlObject in ir.schema.referencedTypes.objects {
```
AnthonyMDev, line 404: We have always used `graphQL` across the project. I see why you want to change this, because it's hard to read and parse where the word `Object` begins and the word `graphQL` ends. It's just not the convention we've always been using. I'm happy with this change, but we should consider doing this going forward and changing old code as we come across it.

Pinging @calvincestari for visibility.

apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/EnumTemplate.swift
```diff
@@ -43,15 +47,19 @@ struct EnumTemplate: TemplateRenderer {
       \(if: shouldRenderDocumentation, "///")
       \(documentation: "**Deprecated**: \($0.escapedSwiftStringSpecialCharacters())")
       """ })
+    \(if: graphqlEnumValue.name.shouldRenderDocumentation, """
+      \(graphqlEnumValue.name.typeNameDocumentation)
+      """
+    )
```
AnthonyMDev, line 53: Could we just make `typeNameDocumentation` return `nil` if it shouldn't be rendered? The `TemplateString` will remove the line if an optional is appended and it is nil.

Then this could just be:
```suggestion
    \(graphqlEnumValue.name.typeNameDocumentation) 
```

apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/InputObjectTemplate.swift
```diff
@@ -21,8 +21,12 @@ struct InputObjectTemplate: TemplateRenderer {
     return TemplateString(
     """
     \(documentation: graphqlInputObject.documentation, config: config)
+    \(if: graphqlInputObject.name.shouldRenderDocumentation, """
```
AnthonyMDev, line 24: Same as other comment.

apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLName+RenderingHelper.swift
AnthonyMDev, line 1: This is really clean. Nice job!

apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/UnionTemplate.swift
```diff
@@ -18,8 +18,12 @@ struct UnionTemplate: TemplateRenderer {
     TemplateString(
     """
     \(documentation: graphqlUnion.documentation, config: config)
-    static let \(graphqlUnion.formattedName) = Union(
-      name: "\(graphqlUnion.name)",
+    \(if: graphqlUnion.name.shouldRenderDocumentation, """
```
AnthonyMDev, line 21: Same as above

apollo-ios-codegen/Sources/GraphQLCompiler/GraphQLName.swift
```diff
@@ -0,0 +1,50 @@
...
+    }
+    return false
+  }
+  
+  public var typeNameDocumentation: String {
+    """
+    // This type has been renamed from the schema type '\(schemaName)'
+    // using schema customization configuration.
+    """
+  }
```
AnthonyMDev, line 32:
```suggestion
  private var shouldRenderDocumentation: Bool {
    if let customName, !customName.isEmpty {
      return true
    }
    return false
  }
  
  public var typeNameDocumentation: TemplateString? {
    guard shouldRenderDocumentation else { return nil }
    return """
    // This type has been renamed from the schema type '\(schemaName)'
    // using schema customization configuration.
    """
  }
```

apollo-ios-codegen/Sources/IR/IR+EntitySelectionTree.swift
AnthonyMDev: This is related to the fragment field merging work I'm doing. How the heck did this get here? Are you able to revert this back? I don't think this version is correct, and the one I'm working on is pretty heavily modified from this.

AnthonyMDev (dismissed):
This is looking great! Just a few suggestions. Thanks for the work on this @BobaFetters .

Tests/TestCodeGenConfigurations/SchemaCustomization/Package.swift
BobaFetters: Previously we had talked about adding new test configuration projects for different features instead of always adding them into the existing ones. So I think we just need to decide if/when its beneficial to have something in all the existing ones vs having its own. I guess my question here would be does name spacing matter for the purposes of testing this feature? Technically the namespace is a separate piece from the typename which would be customized, so are we gaining a lot by doing testing this in all the existing projects? I am open to doing so just want to make sure we will actually benefit from it.

apollo-ios-codegen/Sources/ApolloCodegenLib/ApolloCodegen.swift
BobaFetters: So if I remember correctly there were some spots where this was done as `graphQL` and some where it was `graphql` so I just brought them all in sync with the one I thought to be more correct. Wasn't aware there was a standard in place for doing `graphQL` but open to switching back if we decide to do so. Agreed that if we keep this we need to ensure to update any other areas where this is the case.

AnthonyMDev: Yeah, there may have been some inconsistent places, but I have been using `graphQL`. But looking at this now, I think your way is an improvement on readability.

apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/RenderingHelpers/GraphQLName+RenderingHelper.swift
```diff
@@ -0,0 +1,161 @@
...
+    case typename
+  }
+  
+  func render(
+    as context: RenderContext
+  ) -> String {
+    //If the name has been customized return it unchanged
+    if let customName = name.customName {
+      return customName
+  }
```
AnthonyMDev, line 17:
```suggestion
    }
```

AnthonyMDev: Indent was incorrect here

apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/EnumTemplate.swift
```diff
@@ -43,15 +44,16 @@ struct EnumTemplate: TemplateRenderer {
       \(if: shouldRenderDocumentation, "///")
       \(documentation: "**Deprecated**: \($0.escapedSwiftStringSpecialCharacters())")
       """ })
+    \(graphqlEnumValue.name.typeNameDocumentation)
```
AnthonyMDev, line 47: Haha this was a silly one.

There was an extra whitespace character at the end of this line. The `TemplateString` removes the line only if the line is empty, and since it wasn't that's where your issue was coming from.

This should make it work correctly.

```suggestion
    \(graphqlEnumValue.name.typeNameDocumentation)
```

Tests/TestCodeGenConfigurations/SchemaCustomization/Package.swift
AnthonyMDev: My first concern is that we run the test code gen configurations sequentially during CI jobs. This is already the longest CI job by a wide margin, so adding more configurations just makes all of our CI runs longer, which slows us down when waiting to merge things.

I hear you that the namespace is not something that is likely to conflict with this work, but over time we've caught a lot of weird bugs we never would have anticipated would have an issue with different codegen configs/package types.

I don't think that we should be creating new test projects for _features_. Features that should work across different types of configs should be tested to work with all of them. I think that if we added some new "feature" that was really a new way of organizing your projects, only then should it have a new test projects. Something like allowing you to merge all of the generated files into a single API file for example (which I don't think we should ever actually do, but just as an example).

BobaFetters: Yea that makes sense I can get behind that, I'll make the adjustments.

apollo-ios-codegen/Sources/ApolloCodegenLib/ApolloCodegen.swift
calvincestari: I think @BobaFetters is right and it has been inconsistent. "GraphQL" is the whole name so I agree that lowercasing the entire thing feels more correct. We can make the change across the rest of the project in a separate PR though, it might just add noise to this one.

Tests/TestCodeGenConfigurations/SchemaCustomization/Package.swift
calvincestari: The test codegen configuration projects were meant for the different modularity types supported in codegen, not specific features. So yes, a feature should be implemented in all of them so we get the confidence that it's working correctly across all modularity configurations.

AnthonyMDev (approved):
This looks great! Thanks for all the work on this Zach. This is going to resolve so many problems for so many users going forward!

PR #281, opened 2024-03-04
Feature/ContextIdentifier for the mutate queries

VladimirK-ah:
Add contextIdentifier for the mutate queries in order to be able to add headers in the queries

### DIFF
--- a/apollo-ios/Sources/Apollo/ApolloClientProtocol.swift
+++ b/apollo-ios/Sources/Apollo/ApolloClientProtocol.swift
@@ -54,12 +54,14 @@ public protocol ApolloClientProtocol: AnyObject {
   /// - Parameters:
   ///   - mutation: The mutation to perform.
   ///   - publishResultToStore: If `true`, this will publish the result returned from the operation to the cache store. Default is `true`.
+  ///   - contextIdentifier: [optional] A unique identifier for this request, to help with deduping cache hits for watchers. Should default to `nil`.
   ///   - context: [optional] A context that is being passed through the request chain. Should default to `nil`.
   ///   - queue: A dispatch queue on which the result handler will be called. Should default to the main queue.
   ///   - resultHandler: An optional closure that is called when mutation results are available or when an error occurs.
   /// - Returns: An object that can be used to cancel an in progress mutation.
   func perform<Mutation: GraphQLMutation>(mutation: Mutation,
                                           publishResultToStore: Bool,
+                                          contextIdentifier: UUID?,
                                           context: RequestContext?,
                                           queue: DispatchQueue,
                                           resultHandler: GraphQLResultHandler<Mutation.Data>?) -> Cancellable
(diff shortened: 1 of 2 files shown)


AnthonyMDev (approved):
Thanks for the addition!

calvincestari:
This PR closed https://github.com/apollographql/apollo-ios/issues/2514.

apollo-ios/Sources/Apollo/ApolloClientProtocol.swift
```diff
@@ -54,12 +54,14 @@ public protocol ApolloClientProtocol: AnyObject {
...
   ///   - mutation: The mutation to perform.
   ///   - publishResultToStore: If `true`, this will publish the result returned from the operation to the cache store. Default is `true`.
+  ///   - contextIdentifier: [optional] A unique identifier for this request, to help with deduping cache hits for watchers. Should default to `nil`.
   ///   - context: [optional] A context that is being passed through the request chain. Should default to `nil`.
   ///   - queue: A dispatch queue on which the result handler will be called. Should default to the main queue.
   ///   - resultHandler: An optional closure that is called when mutation results are available or when an error occurs.
   /// - Returns: An object that can be used to cancel an in progress mutation.
   func perform<Mutation: GraphQLMutation>(mutation: Mutation,
                                           publishResultToStore: Bool,
+                                          contextIdentifier: UUID?,
```
michaelloo, line 64: Hey guys, this is actually a breaking change that has been introduced between 1.9.0 and 1.9.1

My team who used with 1.9.0 earlier this week is having issues pushing internal libraries using Cocoapods `~> 1.9.0` because it grabs the newer version 1.9.1 and classes that were using `ApolloClientProtocol` now aren't able to compile due to the missing `contextIdentifier`

AnthonyMDev: Ooof. Yeah we missed this. Sorry for the difficulties that caused. We'll need to roll this out as a minor version bump at least and put out a new 1.9 patch that reverts this change. I'll handle that tomorrow morning.

michaelloo: e.g.: if we were using dependency injection to inject a `ApolloClientProtocol` to a class or struct, it would now not compile in 1.9.1 because `ApolloClientProtocol` now has `contextIdentifier` as an additional parameter.

This was probably not intended, maybe something like this would prevent this to be a breaking change in the future:
```swift
public extension ApolloClientProtocol {
  func perform<Mutation: GraphQLMutation>(
    mutation: Mutation,
    publishResultToStore: Bool,
    context: RequestContext?,
    queue: DispatchQueue,
    resultHandler: GraphQLResultHandler<Mutation.Data>?
  ) -> Cancellable {
    perform(
      mutation: mutation,
      publishResultToStore: publishResultToStore,
      contextIdentifier: nil,
      context: context,
      queue: queue,
      resultHandler: resultHandler
    )
  }
}
```
Otherwise a major version bump would have been expected.

AnthonyMDev: I've release 1.9.2 with your suggested change here. Thanks @michaelloo

michaelloo: Thanks @AnthonyMDev for the quick fix!

PR #282, opened 2024-03-05
Pagination - Change cancel() to reset()

AnthonyMDev:
- Changing cancel() to reset()
  - Removing the sending of completion: .finished when calling reset()
  - Don’t remove subscribers on reset()
- Changing convenience functions
  - from static func to convenience init
  - Reorder parameters extractPageInfo before pageResolver
- Update docs for these changes


Tuist/ProjectDescriptionHelpers/Targets/Target+ApolloPaginationTests.swift
```diff
@@ -19,7 +19,8 @@ extension Target {
                 .package(product: "Apollo"),
                 .package(product: "ApolloAPI"),
                 .package(product: "ApolloTestSupport"),
-                .package(product: "ApolloPagination")
+                .package(product: "ApolloPagination"),
+                .package(product: "Nimble")
```
Iron-Ham, line 23: 👀 
Nimble isn't in use here yet, as far as I can tell. Excited to see where and how it gets used!

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager+Convenience.swift
```diff
@@ -5,33 +5,37 @@ import Foundation
 // MARK: - GraphQLQueryPager Convenience Functions
 
 public extension GraphQLQueryPager {
-  /// Convenience function for creating a pager that has a single query and does not transform output responses.
-  static func makeQueryPager<InitialQuery: GraphQLQuery, P: PaginationInfo>(
+
+  /// Convenience initializer for creating a pager that has a single query and does not 
+  /// transform output responses.
```
Iron-Ham, line 10: Just for my reference – do y'all use an 80 character line divider here? I've been operating I think with 100 or 120 as my divider.

docs/source/pagination/directional-pagers.mdx
```diff
@@ -48,46 +54,40 @@ let pager = GraphQLQueryPager.makeQueryPager(
...
+Generally, it's recommended to use a convenience initializers to create a configured `GraphQLQueryPager`. However, if you need to customize the configuration, you can use the `init` method directly. If your application only uses a specific type of pagination, it is simple to declare a custom `GraphQLQueryPager` convenience initializer that can be used throughout your application. This allows you to easily instantiate a pager without worrying about cases that your application does not support. If your pagination scheme only supports forward pagination and offsets, you can declare a custom `GraphQLQueryPager` convenience initializer that only supports forward pagination and offsets. In this example, we declare two initializers for forward offset pagination -- one with a `transform`, and one without:
```
Iron-Ham, line 90:
```suggestion
Generally, it's recommended to use a convenience initializer to create a configured `GraphQLQueryPager`. However, if you need to customize the configuration, you can use the `init` method directly. If your application only uses a specific type of pagination, it is simple to declare a custom `GraphQLQueryPager` convenience initializer that can be used throughout your application. This allows you to easily instantiate a pager without worrying about cases that your application does not support. If your pagination scheme only supports forward pagination and offsets, you can declare a custom `GraphQLQueryPager` convenience initializer that only supports forward pagination and offsets. In this example, we declare two initializers for forward offset pagination -- one with a `transform`, and one without:
```

docs/source/pagination/introduction.mdx
```diff
@@ -81,14 +81,14 @@ pager.sink { result in
...
 - `loadNext`: Fetches the next page of data. Can only be called after `fetch` has been called.
 - `loadPrevious`: Fetches the previous page of data. Can only be called after `fetch` has been called.
 - `loadAll`: Fetches all pages of data. If no initial page is detected, it will first call `fetch` to fetch the first page of data. Will continue to fetch all pages until a `PageInfo` object indicates that there are no more pages to fetch. This function is compatible with forward, reverse, and bidirectional pagination.
 
 ## Cancelling ongoing requests
 
-The `GraphQLQueryPager` class provides a `cancel` method, which can be used to cancel any ongoing fetch operations as well as cease all subscriptions. Once cancelled, the pager's state is reset. Any subscriber that was subscribed to the pager will receive a termination signal, and will no longer receive updates. In order to resume fetching data, a new `GraphQLQueryPager` instance must be created.
+The `GraphQLQueryPager` class provides a `reset` method, which can be used to cancel all in-flight fetch operations and stop watching for changes. Once the pager's state is reset, you can call `fetch` to being receiving updates again.
```
Iron-Ham, line 91: I wonder if the `stop watching for changes` is a little ambiguous? It's true – in that we stop the underlying watchers. Should we clarify that any object subscribed to the pager would remains subscribed?

Iron-Ham (commented):
Seems like `AsyncGraphQLQueryPagerTests` need updates to match these changes – but overall, looks good!

Tuist/ProjectDescriptionHelpers/Targets/Target+ApolloPaginationTests.swift
AnthonyMDev: Yeah, I planned on using it in some tests for the original approach, but then when we went with this approach, I deleted those tests. But regardless, it's we've always used for test matchers and will use in the future, so it's fine to add it here for now.

apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager+Convenience.swift
AnthonyMDev: We use 100! I really need to get a style guide and a swift-format configuration set up for the whole project one of these days.

docs/source/pagination/introduction.mdx
AnthonyMDev: Fixed

PR #286, opened 2024-03-07
Session and task descriptions

hishma:
This pull request updates the Apollo `URLSessionClient` adding the ability to optionally provide a session description and/or task description to the `URLSession` , and any subsequent `URLSessionTask` ’s created by the session.

These are helpful when debugging, especially when [analyzing HTTP traffic with Instruments](https://developer.apple.com/documentation/foundation/url_loading_system/analyzing_http_traffic_with_instruments). 

See also:
- [sessionDescription](https://developer.apple.com/documentation/foundation/urlsession/1408277-sessiondescription) 
- [taskDescription](https://developer.apple.com/documentation/foundation/urlsessiontask/1409798-taskdescription)

### DIFF
--- a/Tests/ApolloInternalTestHelpers/MockURLSession.swift
+++ b/Tests/ApolloInternalTestHelpers/MockURLSession.swift
@@ -28,6 +28,7 @@ public final class MockURLSessionClient: URLSessionClient {
   }
 
   public override func sendRequest(_ request: URLRequest,
+                                   taskDescription: String? = nil,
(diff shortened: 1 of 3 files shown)


AnthonyMDev (approved):
This seems good to me! Thanks for the addition and for adding tests for it!

AnthonyMDev:
Actually, this may be a breaking change this way since `sendRequest` is an `open` function (overrides in subclasses will need to be changed). We need to keep the old version of the function and add this as a second overload.

I can make those changes though.

PR #287, opened 2024-03-08
update: Expand test suite for multipart subscriptions

calvincestari:
* Expands the multipart subscriptions test suite to cover all edge cases detailed in https://github.com/apollographql/team-prometheus/pull/261.
* Refactors the parsing logic to handle null `errors`.

### DIFF
--- a/Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
+++ b/Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
@@ -10,7 +10,7 @@ final class MultipartResponseSubscriptionParserTests: XCTestCase {
 
   // MARK: - Error tests
 
-  func test__error__givenChunk_withIncorrectContentType_shouldReturnError() throws {
+  func test__error__givenIncorrectContentType_shouldReturnError() throws {
     let subject = InterceptorTester(interceptor: MultipartResponseParsingInterceptor())
 
     let expectation = expectation(description: "Received callback")
@@ -46,7 +46,7 @@ final class MultipartResponseSubscriptionParserTests: XCTestCase {
     wait(for: [expectation], timeout: defaultTimeout)
   }
 
-  func test__error__givenChunk_withTransportError_shouldReturnError() throws {
+  func test__error__givenTransportError_shouldReturnError() throws {
     let subject = InterceptorTester(interceptor: MultipartResponseParsingInterceptor())
 
     let expectation = expectation(description: "Received callback")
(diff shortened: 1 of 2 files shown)


calvincestari:
@AnthonyMDev, this PR is ready for review. There are failing tests in the 'Codegen Test Configurations' job but that's failing due to warnings about unused variables. This PR is fine to merge with those because it's merging into `feature/defer-execution-networking` and those warnings are fixed in https://github.com/apollographql/apollo-ios-dev/pull/217 which also merges into `feature/defer-execution-networking`. So once this PR is merged I'll rebase the branch for https://github.com/apollographql/apollo-ios-dev/pull/217.

Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
```diff
@@ -215,7 +306,29 @@ final class MultipartResponseSubscriptionParserTests: XCTestCase {
...
 
-    let expectation = expectation(description: "Payload (null) ignored")
+    let expectation = expectation(description: "Null payload ignored")
+    expectation.isInverted = true
+
+    _ = network.send(operation: MockSubscription<Time>()) { result in
+      expectation.fulfill()
+    }
+
+    wait(for: [expectation], timeout: defaultTimeout)
```
AnthonyMDev, line 316: I hate tests that _have_ to wait for their timeout. They slow down test runs and you never are actually 100% sure if they are actually verifying behavior or just not failing in time. I know sometimes they are unavoidable, but I wonder if we could avoid this...

Is it guaranteed that payloads are processed in the order they are received, even if received in parallel? If so, could we change this test to send 3 payloads: [1, null, 3] and verify that we just get 2 completion callbacks with the correct values?

Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
```diff
@@ -215,7 +306,29 @@ final class MultipartResponseSubscriptionParserTests: XCTestCase {
...
+  }
+
+  func test__parsing__givenNullErrors_shouldIgnore() throws {
```
AnthonyMDev, line 319: Same as above.

AnthonyMDev (approved):
LGTM, just a question on a few tests!

Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
calvincestari: Sure, I can change the test to not have to rely on the timeout; you make a good point.

As for payload order processing, yes I believe they will be processed in the order received. I don't think it's possible to receive payloads in parallel because the network processing queue serial.

Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
calvincestari: Donein https://github.com/apollographql/apollo-ios-dev/pull/287/commits/15f0a48cef37b554a3d980b009229068b831f88f.

Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/287/commits/15f0a48cef37b554a3d980b009229068b831f88f.

calvincestari:
> There are failing tests in the 'Codegen Test Configurations' job but that's failing due to warnings about unused variables. This PR is fine to merge with those because it's merging into feature/defer-execution-networking and those warnings are fixed in https://github.com/apollographql/apollo-ios-dev/pull/217 which also merges into feature/defer-execution-networking. So once this PR is merged I'll rebase the branch for https://github.com/apollographql/apollo-ios-dev/pull/217.

As mentioned before re. the failing tests - once this PR was merged I then merged `feature/defer-execution-networking` into `defer/partial-and-incremental-execution` and the test suite passes as expected.

PR #292, opened 2024-03-09
[ApolloPagination] Add completion callbacks to fetch and refetch in GraphQLQueryPager for API Consistency

Iron-Ham:
For API consistency, it makes sense that both `fetch` and `refetch` have completion blocks to indicate when their operations are completed. If we think about the way that the `Async` variant works, we always have a "completion" signal in the form of the `await` – whereas we don't for `fetch` and `refetch` in the `GraphQLQueryPager`. 

____

Related: https://github.com/apollographql/apollo-ios/issues/3349

___

Changes in `GraphQLQueryPager.swift`:

* [`public class GraphQLQueryPager<Model>: Publisher {`](diffhunk://#diff-15b3e5571b28a433475df3ef139318cd9aa539ef1302cd4e4600639fbb8d57aaL232-R245): The `refetch` and `fetch` methods have been updated to include two new parameters: `callbackQueue` and `completion`. This allows users to specify a dispatch queue on which the completion callback should be fired and a completion block that triggers after the operation execution.

Changes in `GraphQLQueryPagerCoordinator.swift`:

* [`public protocol PagerType {`](diffhunk://#diff-55a381cd001c8f61f1ea8b3de40ecb0cfd97d4fadf68641fa77317dcb0ca8d2cL28-R36): The `refetch` and `fetch` methods in the `PagerType` protocol have been updated to include the `callbackQueue` and `completion` parameters, aligning with the changes made in the `GraphQLQueryPager` class.
* `class GraphQLQueryPagerCoordinator<InitialQuery: GraphQLQuery, PaginatedQuery: G`: The `refetch` and `fetch` methods have been updated to include the `callbackQueue` and `completion` parameters. This allows users to specify a dispatch queue on which the completion callback should be fired and a completion block that triggers after the operation execution. [[1]](diffhunk://#diff-55a381cd001c8f61f1ea8b3de40ecb0cfd97d4fadf68641fa77317dcb0ca8d2cL154-R171) [[2]](diffhunk://#diff-55a381cd001c8f61f1ea8b3de40ecb0cfd97d4fadf68641fa77317dcb0ca8d2cL165-R188)

### DIFF
--- a/apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
+++ b/apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift
@@ -228,7 +228,10 @@ public class GraphQLQueryPager<Model>: Publisher {
   }
 
   /// Discards pagination state and fetches the first page from scratch.
-  /// - Parameter cachePolicy: The apollo cache policy to trigger the first fetch with. Defaults to `fetchIgnoringCacheData`.
+  /// - Parameters:
+  ///   - cachePolicy: The apollo cache policy to trigger the first fetch with. Defaults to `fetchIgnoringCacheData`.
+  ///   - callbackQueue: The `DispatchQueue` that the `completion` fires on. Defaults to `main`.
+  ///   - completion: A completion block that will always trigger after the execution of this  operation.
   public func refetch(
     cachePolicy: CachePolicy = .fetchIgnoringCacheData,
     callbackQueue: DispatchQueue = .main,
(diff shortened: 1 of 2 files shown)


AnthonyMDev (approved):
Do we need to update the usage docs for this? Maybe we should create a new branch for 0.2 and do this there, so that those docs don't get deployed until we push the 0.2 version release?

Iron-Ham:
> Do we need to update the usage docs for this? Maybe we should create a new branch for 0.2 and do this there, so that those docs don't get deployed until we push the 0.2 version release?

We probably should re-generate the docset, yeah. _But_ since this is a non-breaking change, I can see an argument for a 0.1.1

SopheaSao:
SopheaSao Cambodia Enjoy with you ♥️

PR #298, opened 2024-03-12
[ApolloPagination] Add conditional `Equatable` conformance to pagers

Iron-Ham:
A bit of an idea I had to ease usage in SwiftUI.

If an `@Observable` view model owns a pager, we cannot observe value changes (`onChange(...)`) of the whole `viewModel`, as the pager breaks conformance. Whether that's a good pattern or not (I think _probably not?_) is a bit besides the point, as I can see some users doing it anyways. 

There are various values stored on the pager classes – the underlying coordinator, the cancellables, etc – but at its core, the underlying data state is what matters, i.e., the value of `_subject`. 

Note: I don't think I can return `lhs._subject.value == rhs._subject.value` since the value is a `Result<(Model, UpdateSource), Error>`. Even if `Model`, `UpdateSource`, and `Error`, are equatable, the tuple value of `(Model, UpdateSource)` is not equatable – thus failing to translate conditional equatability to the `Result`.

___

🤖 Copilot Generated 🤖 

This pull request includes changes to the Apollo iOS Pagination package to improve the comparability and hashability of certain classes and enums. The most important changes include adding `Equatable` conformance to `AsyncGraphQLQueryPager` and `GraphQLQueryPager` classes, and making the `UpdateSource` enum `Hashable`.

Here are the most significant changes:

* [`apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift`](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1R248-R263): Added `Equatable` protocol conformance to `AsyncGraphQLQueryPager`. This allows for equality comparison between instances of `AsyncGraphQLQueryPager` where `Model` is also `Equatable`. The equality is determined by comparing the values and sources of the subjects of the two `AsyncGraphQLQueryPager` instances, or by comparing the localized descriptions of their errors if they both have errors.

* [`apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager.swift`](diffhunk://#diff-15b3e5571b28a433475df3ef139318cd9aa539ef1302cd4e4600639fbb8d57aaR252-R267): Similarly, the `GraphQLQueryPager` class now conforms to the `Equatable` protocol. The equality comparison is done in the same way as for `AsyncGraphQLQueryPager`.

* [`apollo-ios-pagination/Sources/ApolloPagination/UpdateSource.swift`](diffhunk://#diff-1b96ca532f31d2f77a8a78b720248a4cb0e29f641c22f887b0d3d9756648ed56L1-R1): The `UpdateSource` enum now conforms to the `Hashable` protocol. This allows instances of `UpdateSource` to be used in sets or as dictionary keys.

### DIFF
--- a/Tests/ApolloPaginationTests/AsyncGraphQLQueryPagerTests.swift
+++ b/Tests/ApolloPaginationTests/AsyncGraphQLQueryPagerTests.swift
@@ -354,6 +354,37 @@ final class AsyncGraphQLQueryPagerTests: XCTestCase {
     subscriber.cancel()
   }
 
+  func test_equatable() async {
+    let pagerA = await AsyncGraphQLQueryPager(pager: createPager(), transform: { previous, initial, next in
+      let allPages = previous + [initial] + next
+      return allPages.flatMap { data in
+        data.hero.friendsConnection.friends.map { $0.name }
+      }
+    })
+
+    let pagerB = await AsyncGraphQLQueryPager(pager: createPager(), transform: { previous, initial, next in
+      let allPages = previous + [initial] + next
+      return allPages.flatMap { data in
+        data.hero.friendsConnection.friends.map { $0.name }
+      }
+    })
+
+    XCTAssertEqual(pagerA, pagerB)
+
+    pagerA._subject.send(.success((["Al-Khwarizmi", "Al-Jaziri", "Charles Babbage", "Ada Lovelace"], .cache)))
+    XCTAssertNotEqual(pagerA, pagerB)
+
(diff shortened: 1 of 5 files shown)


AnthonyMDev (approved):
Do you feel like this is worth writing unit tests for? I'm 50/50 on it. Otherwise, LGTM

Iron-Ham:
Gotta love unit tests. Sometimes the code seems simple enough, and then you write a test case and realized that with the previous setup, `nil != nil`.

AnthonyMDev (approved):

PR #299, opened 2024-03-12
[ApolloPagination] Remove `async` requirement from `AsyncGraphQLQueryPager` initializers

Iron-Ham:
**This is a breaking change, and would require a 0.2.0 release**. 

In order to make instantiation of the `AsyncGraphQLQueryPager` easier, we can remove the `async` requirement of the initializer _if_ we also ensure that accesses to `cancellables` becomes atomic. This pull request primarily focuses on the removal of asynchronous initializers and the introduction of `Task` in the `AsyncGraphQLQueryPager` class and its related test classes. The most significant changes include the removal of `await` from `AsyncGraphQLQueryPager` initializers, the introduction of `Task` for subscribing to pager results, and the use of `@Atomic` for thread-safe mutation of `cancellables`.

___

#### Why?

Honestly, it just feels better to instantiate it this way: it opens up the reality of instantiating the object in a `SwiftUI` view model, and then using it in an asynchronous context via `.task` modifiers in the view as the `async` entry-point. We could technically do that by declaring a new `Task` to instantiate the pager in, but then we are relegated to an optional-only pager. 

The effective change to SwiftUI view models is something like this:

Before:

```swift
@Observable final class MyViewModel {
  var pager: AsyncGraphQLQueryPager<[MyCustomModel]>?

  init() {
    Task {
      pager = await AsyncGraphQLQueryPager(...)
    }
  }
}
```

After:

```swift
@Observable final class MyViewModel {
  var pager: AsyncGraphQLQueryPager<[MyCustomModel]>

  init() {
    pager = AsyncGraphQLQueryPager(...)
  }
}
```

___

I ran the `AsyncGraphQLQueryPagerTests` suite 100 times with the address sanitizer:
 ✅  `Executed 700 tests, with 0 failures (0 unexpected) in 74.311 (74.597) seconds` 

100 times with the thread sanitizer:
✅  `Executed 700 tests, with 0 failures (0 unexpected) in 79.664 (79.981) seconds`

___

Changes to `AsyncGraphQLQueryPager`:

* [`apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift`](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L12-R12): Removed `async` from the initialization of `AsyncGraphQLQueryPager` and its related methods. The `cancellables` property has been made atomic to ensure thread safety. Subscription to the pager's results has been modified to store the cancellable internally. [[1]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L12-R12) [[2]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L21-R24) [[3]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L40-R51) [[4]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L59-R65) [[5]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L71-R78) [[6]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L96-R102) [[7]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L111-R117) [[8]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L140-R154) [[9]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L164-R170) [[10]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L179-R185) [[11]](diffhunk://#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L194-R203)

Changes to `AsyncGraphQLQueryPagerTests`:

* [`Tests/ApolloPaginationTests/AsyncGraphQLQueryPagerTests.swift`](diffhunk://#diff-59a184662b7113ca556c2ff1481eaee2d6fa8dc3f06860cda2cf19acf6fc218cL31-R31): Adjusted the tests to reflect the removal of `async` from the initialization of `AsyncGraphQLQueryPager`. [[1]](diffhunk://#diff-59a184662b7113ca556c2ff1481eaee2d6fa8dc3f06860cda2cf19acf6fc218cL31-R31) [[2]](diffhunk://#diff-59a184662b7113ca556c2ff1481eaee2d6fa8dc3f06860cda2cf19acf6fc218cL87-R87) [[3]](diffhunk://#diff-59a184662b7113ca556c2ff1481eaee2d6fa8dc3f06860cda2cf19acf6fc218cL149-R149) [[4]](diffhunk://#diff-59a184662b7113ca556c2ff1481eaee2d6fa8dc3f06860cda2cf19acf6fc218cL211-R211) [[5]](diffhunk://#diff-59a184662b7113ca556c2ff1481eaee2d6fa8dc3f06860cda2cf19acf6fc218cL276-R276) [[6]](diffhunk://#diff-59a184662b7113ca556c2ff1481eaee2d6fa8dc3f06860cda2cf19acf6fc218cL309-R309) [[7]](diffhunk://#diff-849d1616469c55001c042bc3f8f1c168399a12a377a5522115b5da5bebf6559dL61-R61)

Changes to `PagerCoordinator+Erase`:

* [`Tests/ApolloPaginationTests/PagerCoordinator+Erase.swift`](diffhunk://#diff-74ba0cde07e55b1bddc4aa253b14f6f0a0ca4a662e33083f77bad0befa5ea905L35-R36): Adjusted the tests to reflect the removal of `async` from the initialization of `AsyncGraphQLQueryPager`. [[1]](diffhunk://#diff-74ba0cde07e55b1bddc4aa253b14f6f0a0ca4a662e33083f77bad0befa5ea905L35-R36) [[2]](diffhunk://#diff-74ba0cde07e55b1bddc4aa253b14f6f0a0ca4a662e33083f77bad0befa5ea905L45-R46) [[3]](diffhunk://#diff-74ba0cde07e55b1bddc4aa253b14f6f0a0ca4a662e33083f77bad0befa5ea905L55-R56)

Changes to `GraphQLQueryPager+Convenience`:

* [`apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPager+Convenience.swift`](diffhunk://#diff-1ab0c1f28ec542beeb66edd5a206a3857033e09c4102a474bc306b9c3b48ab43L162-R163): Adjusted the convenience methods to reflect the removal of `async` from the initialization of `AsyncGraphQLQueryPager`. [[1]](diffhunk://#diff-1ab0c1f28ec542beeb66edd5a206a3857033e09c4102a474bc306b9c3b48ab43L162-R163) [[2]](diffhunk://#diff-1ab0c1f28ec542beeb66edd5a206a3857033e09c4102a474bc306b9c3b48ab43L182-R183) [[3]](diffhunk://#diff-1ab0c1f28ec542beeb66edd5a206a3857033e09c4102a474bc306b9c3b48ab43L204-R205) [[4]](diffhunk://#diff-1ab0c1f28ec542beeb66edd5a206a3857033e09c4102a474bc306b9c3b48ab43L227-R228) [[5]](diffhunk://#diff-1ab0c1f28ec542beeb66edd5a206a3857033e09c4102a474bc306b9c3b48ab43L252-R253) [[6]](diffhunk://#diff-1ab0c1f28ec542beeb66edd5a206a3857033e09c4102a474bc306b9c3b48ab43L279-R280)

### DIFF
--- a/apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
+++ b/apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
@@ -9,7 +9,7 @@ public class AsyncGraphQLQueryPager<Model>: Publisher {
   public typealias Output = Result<(Model, UpdateSource), Error>
   let _subject: CurrentValueSubject<Output?, Never> = .init(nil)
   var publisher: AnyPublisher<Output, Never> { _subject.compactMap({ $0 }).eraseToAnyPublisher() }
-  public var cancellables: Set<AnyCancellable> = []
+  @Atomic public var cancellables: Set<AnyCancellable> = []
   public let pager: any AsyncPagerType
 
   public var canLoadNext: Bool { get async { await pager.canLoadNext } }
@@ -18,45 +18,51 @@ public class AsyncGraphQLQueryPager<Model>: Publisher {
   init<Pager: AsyncGraphQLQueryPagerCoordinator<InitialQuery, PaginatedQuery>, InitialQuery, PaginatedQuery>(
     pager: Pager,
     transform: @escaping ([PaginatedQuery.Data], InitialQuery.Data, [PaginatedQuery.Data]) throws -> Model
-  ) async {
+  ) {
     self.pager = pager
-    await pager.subscribe { [weak self] result in
-      guard let self else { return }
-      let returnValue: Output
+    Task {
+      let cancellable = await pager.subscribe { [weak self] result in
+        guard let self else { return }
+        let returnValue: Output
 
-      switch result {
-      case let .success((output, source)):
-        do {
-          let transformedModels = try transform(output.previousPages, output.initialPage, output.nextPages)
-          returnValue = .success((transformedModels, source))
-        } catch {
+        switch result {
+        case let .success((output, source)):
+          do {
+            let transformedModels = try transform(output.previousPages, output.initialPage, output.nextPages)
+            returnValue = .success((transformedModels, source))
+          } catch {
+            returnValue = .failure(error)
+          }
+        case let .failure(error):
           returnValue = .failure(error)
         }
-      case let .failure(error):
-        returnValue = .failure(error)
-      }
 
-      _subject.send(returnValue)
-    }.store(in: &cancellables)
+        _subject.send(returnValue)
+      }
+      _ = $cancellables.mutate { $0.insert(cancellable) }
+    }
   }
 
   init<Pager: AsyncGraphQLQueryPagerCoordinator<InitialQuery, PaginatedQuery>, InitialQuery, PaginatedQuery>(
     pager: Pager
-  ) async where Model == PaginationOutput<InitialQuery, PaginatedQuery> {
+  ) where Model == PaginationOutput<InitialQuery, PaginatedQuery> {
     self.pager = pager
-    await pager.subscribe { [weak self] result in
-      guard let self else { return }
-      let returnValue: Output
+    Task {
+      let cancellable = await pager.subscribe { [weak self] result in
+        guard let self else { return }
+        let returnValue: Output
 
-      switch result {
-      case let .success((output, source)):
-        returnValue = .success((output, source))
-      case let .failure(error):
-        returnValue = .failure(error)
+        switch result {
+        case let .success((output, source)):
+          returnValue = .success((output, source))
+        case let .failure(error):
+          returnValue = .failure(error)
+        }
+
+        _subject.send(returnValue)
       }
-
-      _subject.send(returnValue)
-    }.store(in: &cancellables)
+      _ = $cancellables.mutate { $0.insert(cancellable) }
+    }
   }
 
   convenience init<
@@ -68,8 +74,8 @@ public class AsyncGraphQLQueryPager<Model>: Publisher {
     pager: Pager,
     initialTransform: @escaping (InitialQuery.Data) throws -> Model,
     pageTransform: @escaping (PaginatedQuery.Data) throws -> Model
-  ) async where Model: RangeReplaceableCollection, Model.Element == Element {
-    await self.init(
+  ) where Model: RangeReplaceableCollection, Model.Element == Element {
+    self.init(
       pager: pager,
       transform: { previousData, initialData, nextData in
         let previous = try previousData.flatMap { try pageTransform($0) }
@@ -93,7 +99,7 @@ public class AsyncGraphQLQueryPager<Model>: Publisher {
     pageResolver: ((P, PaginationDirection) -> PaginatedQuery?)?,
     initialTransform: @escaping (InitialQuery.Data) throws -> Model,
     pageTransform: @escaping (PaginatedQuery.Data) throws -> Model
-  ) async where Model: RangeReplaceableCollection, Model.Element == Element {
+  ) where Model: RangeReplaceableCollection, Model.Element == Element {
     let pager = AsyncGraphQLQueryPagerCoordinator(
       client: client,
       initialQuery: initialQuery,
@@ -108,7 +114,7 @@ public class AsyncGraphQLQueryPager<Model>: Publisher {
       },
       pageResolver: pageResolver
     )
-    await self.init(
+    self.init(
       pager: pager,
       initialTransform: initialTransform,
       pageTransform: pageTransform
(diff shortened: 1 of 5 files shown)


apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
```diff
@@ -191,9 +197,10 @@ public class AsyncGraphQLQueryPager<Model>: Publisher {
   /// Subscribe to the results of the pager, with the management of the subscriber being stored internally to the `AnyGraphQLQueryPager`.
   /// - Parameter completion: The closure to trigger when new values come in.
   public func subscribe(completion: @MainActor @escaping (Output) -> Void) {
-    publisher.sink { result in
+    let cancellable = publisher.sink { result in
       Task { await completion(result) }
-    }.store(in: &cancellables)
+    }
+    _ = $cancellables.mutate { $0.insert(cancellable) }
```
AnthonyMDev, line 203: Does this introduce a possible race condition where the `cancellable` set is cancelled between the time this one is created and the time when it is inserted into the set?

apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
```diff
@@ -18,45 +18,51 @@ public class AsyncGraphQLQueryPager<Model>: Publisher {
...
         }
-      case let .failure(error):
-        returnValue = .failure(error)
-      }
 
-      _subject.send(returnValue)
-    }.store(in: &cancellables)
+        _subject.send(returnValue)
+      }
+      _ = $cancellables.mutate { $0.insert(cancellable) }
```
AnthonyMDev, line 42: Same question as below

AnthonyMDev (commented):
I like the intent of this PR, and if it works correctly, then let's ship it. But I'm concerned about creating race conditions with the `cancellables`.

apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
Iron-Ham: It seems unlikely, but I'm also not sure if it's a consequential effect?

This is valid:

```swift
let cancellables: Set<AnyCancellable> = []
let cancellable: AnyCancellable = ... 
cancellable.cancel()
cancellables.insert(cancellable)
```

Iron-Ham: On further examination – I think that the only thing this change really does is remove the guarantee that we're setup to publish values prior to a `subscribe` being called. 

i.e., with code like this:

```swift
let pager = AsyncGraphQLQueryPager(...)
pager.subscribe { ... } 
```

There's a chance the act of a subscriber creating an `AnyCancellable` to be internally managed by the `AsyncGraphQLQueryPager` is executed _prior_ to the `init` setting up to forward results from the `AsyncGraphQLQueryPagerCoordinator`. There's no negative impact to this.

Even if a networking operation somehow begins and completes execution prior to either `subscribe` or `init`'s `Task` completing – its data will still be forwarded the moment that the Task does complete.

Iron-Ham: This may be a better option than what's in `main` currently – especially for edge cases in SwiftUI. I haven't reproduced this, but this is a bit of a thought exercise.

Imagine we have a simple `ViewModel` that instantiates an `AsyncGraphQLQueryPager`, and has functions for `fetch`, `refetch` and `loadNext`:

```swift
@Observable class ViewModel {
  let pager: AsyncGraphQLQueryPager<[String]>!
  var names: [String] = []
  init() {
    Task {
      pager = AsyncGraphQLQueryPager(...)
    }
  }

  func fetch() async {
    await pager.fetch()
  }
  func refetch() async {
    await pager.fetch()
  }
  func loadNext() async {
    try? await pager.loadNext()
  }
}
```

Note the instantiation – we do so in a `Task` within the `init`. 

A common pattern would be to use a top-level task on the top-level view to perform the initial `fetch`:

```swift
struct SwiftUIView: View {
  @StateObject var viewModel = ViewModel()
  var body: some View {
    List($viewModel.names) { name in
      Text(name)
        .task {
          guard name == viewModel.names.last else { return }
          await viewModel.loadNext()
        }
    }
    .task {
      await viewModel.fetch()
    }
    .refreshable {
      await viewModel.refetch()
    }
  }
}
```

There's actually more concern in this existing code than I would have for the proposed code. Specifically: the `task` modifier on the top-level view can be treated as analogous to `viewWillAppear`. What happens if we call `viewModel.fetch()`, and the view model's `init`'s `Task` hasn't completed yet? There would be no pager to fetch from.

The proposed code would move the initialization of a `pager` outside of a `Task`. That means that in a similar scenario to the one above, the pager would begin fetching – and then forward data when the pager's `init` `Task` completes.

AnthonyMDev:
Thanks for all the elaboration on this!

PR #300, opened 2024-03-15
Provide a direct means to observe changes in ApolloStore

jamesonwilliams:
(Full context in https://github.com/apollographql/apollo-ios/issues/3350.)

Apollo Kotlin currently provides a public [`changedKeys`](https://github.com/apollographql/apollo-kotlin/blob/v4.0.0-beta.5/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo3/cache/normalized/ApolloStore.kt#L27-L30) property to get notified when objects in the ApolloStore change. Internally, iOS has had a similar facility, too -- but it hasn't been `public` until now.

### DIFF
--- a/Tests/ApolloTests/Cache/StoreSubscriptionTests.swift
+++ b/Tests/ApolloTests/Cache/StoreSubscriptionTests.swift
@@ -1,0 +1,83 @@
+import XCTest
+@testable import Apollo
+
+class StoreSubscriptionTests: XCTestCase {
+  static let defaultWaitTimeout: TimeInterval = 1
+
+  var store: ApolloStore!
+
+  override func setUpWithError() throws {
+    try super.setUpWithError()
+    store = ApolloStore()
+  }
+
+  override func tearDownWithError() throws {
+    store = nil
+    try super.tearDownWithError()
+  }
+
+  // MARK: - Tests
+
+  func testSubscriberIsNotifiedOfStoreUpdate() throws {
+    let cacheKeyChangeExpectation = XCTestExpectation(description: "Subscriber is notified of cache key change")
+    let expectedChangeKeySet: Set<String> = ["QUERY_ROOT.__typename", "QUERY_ROOT.name"]
+    let subscriber = SimpleSubscriber(cacheKeyChangeExpectation, expectedChangeKeySet)
+
+    store.subscribe(subscriber)
+    addTeardownBlock { self.store.unsubscribe(subscriber) }
+
+    store.publish(
+      records: [
+        "QUERY_ROOT": [
+          "__typename": "Hero",
+          "name": "Han Solo"
+        ]
+      ]
+    )
+
+    wait(for: [cacheKeyChangeExpectation], timeout: Self.defaultWaitTimeout)
+  }
+
+  func testUnsubscribedSubscriberDoesNotReceiveStoreUpdate() throws {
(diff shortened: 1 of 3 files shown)


Tests/ApolloTests/Cache/StoreSubscriptionTests.swift
```diff
@@ -0,0 +1,83 @@
...
+          "__typename": "Hero",
+          "name": "Han Solo"
+        ]
+      ]
+    )
+
+    wait(for: [cacheKeyChangeExpectation], timeout: Self.defaultWaitTimeout)
+  }
+
+  func testUnsubscribedSubscriberDoesNotReceiveStoreUpdate() throws {
```
AnthonyMDev, line 41: I know this test is correct as far as testing all the behavior thoroughly, but we try to avoid testing that something _doesn't_ happen in asynchronous context because it requires the test to run and wait for the timeout. Having multiple tests like that ends up making your test runs take too long to complete.

Instead, let's just change this to confirm that after unsubscribe is called, the subscriber is no longer in the `ApolloStore.subscribers` set. I think that should be sufficient.

You'll need to make the `subscribers` property `internal` instead of `private`. And since that is done async, the easiest way to check for it is with `Nimble`. Just add `import Nimble` to the top of this test file, then the test assertion is `expect(self.store.subscribers).toEventually(beEmpty())`.

AnthonyMDev (changes requested):
Code for this looks good! Just one request on the tests and then I'm good to merge this. Thanks so much for the PR!

Tests/ApolloTests/Cache/StoreSubscriptionTests.swift
jamesonwilliams: Makes sense! Thanks for the actionable feedback - updating now

apollo-ios/Sources/Apollo/ApolloStore.swift
```diff
@@ -83,13 +85,23 @@ public class ApolloStore {
     }
   }
 
-  func subscribe(_ subscriber: ApolloStoreSubscriber) {
+  /// Subscribes to notifications of ApolloStore content changes
+  ///
+  /// - Parameters:
+  ///    - subscriber: A subscriber to receive content change notificatons. To avoid a retain cycle,
+  ///                  ensure you call `unsubscribe` on this subcriber before it goes out of scope.
```
calvincestari, line 92:
```suggestion
  ///                  ensure you call `unsubscribe` on this subscriber before it goes out of scope.
```

AnthonyMDev (approved):

apollo-ios/Sources/Apollo/ApolloStore.swift
jamesonwilliams: This fix ended up in https://github.com/apollographql/apollo-ios-dev/pull/300/commits/6f66f7be3dcce250c43c20b11bde2b53df2b615f

Tests/ApolloTests/Cache/StoreSubscriptionTests.swift
jamesonwilliams: Updated in https://github.com/apollographql/apollo-ios-dev/pull/300/commits/6f66f7be3dcce250c43c20b11bde2b53df2b615f

PR #303, opened 2024-03-18
Add GH action to check PR metadata

calvincestari:
This adds a GitHub Action that will validate some of the metadata for pull requests. Note that it will skip of the PR is a draft and will then only validate once it's marked as "ready for review". As we want to do more to validate PRs we can add to this action under the umbrella of "metadata".


AnthonyMDev (approved):
LGTM

PR #304, opened 2024-03-18
fix: Updates Swift `Package.resolved` file version format

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3355.

This is the fix to accept version `3` as a valid format for the version checker.

### DIFF
--- a/Tests/CodegenCLITests/VersionCheckerTests.swift
+++ b/Tests/CodegenCLITests/VersionCheckerTests.swift
@@ -35,47 +35,78 @@ class VersionCheckerTests: XCTestCase {
     fileManager = nil
   }
 
-  private func version1PackageResolvedFileBody(apolloVersion version: String) -> String {
-    return """
-    {
-      "object": {
-        "pins": [
-          {
-            "package": "Apollo",
-            "repositoryURL": "https://github.com/apollographql/apollo-ios.git",
-            "state": {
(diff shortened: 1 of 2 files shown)


Tests/CodegenCLITests/VersionCheckerTests.swift
```diff
@@ -35,216 +35,268 @@ class VersionCheckerTests: XCTestCase {
...
+        // then
+        expect(result).to(equal(
+          .versionMismatch(cliVersion: Constants.CLIVersion, apolloVersion: apolloVersion)
+        ))
+      }
+    }
   }
   
-  func test__matchCLIVersionToApolloVersion__givenPackageResolvedFileInXcodeWorkspaceAndProject_withVersion2FileFormat_hasMatchingVersion_returns_versionMatch_fromWorkspace() throws {
+  func test__matchCLIVersionToApolloVersion__givenPackageResolvedFileInXcodeWorkspaceAndProject_withKnownResolvedFileFormats_hasMatchingVersion_returns_versionMatch_fromWorkspace() throws {
```
calvincestari, line 251: This test was changed to have both the workspace and project use the same Apollo version. Previously they were different and that could also affect the outcome. So this test was changed and there is a new test below that has mismatched versions to ensure that the workspace one is still selected.

PR #319, opened 2024-03-28
test-mocks.mdx - fix spelling

evliu:
the definite article should be a relative pronoun, meaning "to ensure"


calvincestari (approved):
Thanks for the correction @evliu.

PR #326, opened 2024-04-02
Fix GraphQLOperationVariableValue conformance

guillian-balisi:
- [GraphQLOperationVariableValue](https://github.com/apollographql/apollo-ios-dev/blob/550bf291989d0b52cbf5992b7a8b341f77f5a123/apollo-ios/Sources/ApolloAPI/GraphQLOperation.swift#L114-L116) protocol defined with property `_jsonEncodableValue` (containing underscore)
- [InputObject](https://github.com/apollographql/apollo-ios-dev/blob/550bf291989d0b52cbf5992b7a8b341f77f5a123/apollo-ios/Sources/ApolloAPI/SchemaTypes/InputObject.swift#L5-L11) conformance to `GraphQLOperationVariableValue` with `jsonEncodableValue` (not containing underscore)
- This PR adds the underscore to fix the protocol conformance

### DIFF
--- a/apollo-ios/Sources/ApolloAPI/SchemaTypes/InputObject.swift
+++ b/apollo-ios/Sources/ApolloAPI/SchemaTypes/InputObject.swift
@@ -7,8 +7,8 @@ public protocol InputObject: GraphQLOperationVariableValue, JSONEncodable, Hasha
 }
 
 extension InputObject {
-  public var _jsonValue: JSONValue { jsonEncodableValue?._jsonValue }
-  public var jsonEncodableValue: (any JSONEncodable)? { __data._jsonEncodableValue }
+  public var _jsonValue: JSONValue { _jsonEncodableValue?._jsonValue }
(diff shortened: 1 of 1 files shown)


calvincestari:
Hi @guillian-balisi - is this solving an issue you're experiencing?

I don't believe this change is needed for two reasons:
* We use underscored function names to indicate that they are private, could change at any time and should not be depended on. That isn't the case with the `jsonEncodableValue` function (no underscore), i.e.: we're indicating that it can be used and trusted.
* `InputObject` already conforms to `GraphQLOperationVariableValue` through the [extension on `JSONEncodable`](https://github.com/apollographql/apollo-ios-dev/blob/main/apollo-ios/Sources/ApolloAPI/GraphQLOperation.swift#L149).

I'm going to close this PR since it's not needed but if you're experiencing an issue that requires this change please comment here again and we'll continue the discussion.

PR #327, opened 2024-04-04
tests: Adds missing tests and enables skipped tests for `@defer` [1/2]

calvincestari:
Supersedes https://github.com/apollographql/apollo-ios-dev/pull/219.

* Refactors test infrastructure to support `@defer`
* Refactors selection set tests for `@defer`
* Fixes a bug with fragment accessor separators in selection set template for `@defer`


Tests/ApolloCodegenInternalTestHelpers/DirectiveHelpers.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+extension String {
+  // This function will only be needed until @defer is merged into the GraphQL spec and is
+  // considered a first-class directive in graphql-js. Right now it is a valid directive but must
+  // be 'enabled' through explicit declaration in the schema.
+  public mutating func appendDeferDirective() {
```
calvincestari, line 7: Until the `@defer` directive is enabled by default in graphql-js the schema needs to declare it. This function enables us to use local small schemas within tests. We can remove this helper once `@defer` is available to all schemas by default.

Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
```diff
@@ -40,7 +40,7 @@ class IRRootFieldBuilderTests: XCTestCase {
   // MARK: - Helpers
 
   func buildSubjectRootField(operationName: String? = nil) async throws {
-    addDeferDirective()
+    schemaSDL.appendDeferDirective()
```
calvincestari, line 43: We can either do this centrally, as it is now, for all localized test schemas OR be explicit about which schemas we want to support `@defer`. I'm open to the latter if you think it'd be a cleaner way to do this.

Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
```diff
@@ -5166,12 +5138,10 @@ class IRRootFieldBuilderTests: XCTestCase {
       )
     ))
 
-    expect(allAnimals_asDog_deferredAsRoot).to(beNil())
+    expect(allAnimals_asDog?.containsDeferredChildFragment).to(beFalse())
```
calvincestari, line 5141: Testing for nil didn't mean there wasn't any deferred child, just that you didn't find it at the type info you tried it on. This new check better ensures no deferred children in the tree.

apollo-ios-codegen/Sources/IR/IR+NamedFragmentSpread.swift
```diff
@@ -18,7 +18,7 @@ public class NamedFragmentSpread: Hashable, CustomDebugStringConvertible {
   /// Indicates the location where the fragment has been "spread into" its enclosing
   /// operation/fragment. It's `scopePath` and `entity` reference are scoped to the operation it
   /// belongs to.
-  let typeInfo: SelectionSet.TypeInfo
+  public let typeInfo: SelectionSet.TypeInfo
```
calvincestari, line 21: I don't think there is any reason this shouldn't be a public constant. Doing so matches what we make available on `InlineFragment` and means we don't need to have a `public` method just to get at `deferCondition`. The type info is also used in the deferred fragment metadata template, but that may change.

AnthonyMDev (approved):

Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
calvincestari: I thought about this more last night and have decided to remove it from the common `buildSubjectRootField` function and instead use it at each schema that needs the defer directive added. My reasoning was that there are more tests that don't need it and using it on each test is more declarative of what's needed. Plus it's going to be easy to forget about this little addition if it's hidden away in a common function but at least now it's more visible and I'm hoping we'll take note of it more once the directive is available in graphql-js like all other directives.

PR #331, opened 2024-04-08
docs: In-memory cache note

calvincestari:
Related to https://github.com/apollographql/apollo-ios/issues/3365.

* Adds a note to the [`InMemoryNormalizedCache ` cache documentation](https://www.apollographql.com/docs/ios/caching/cache-setup#inmemorynormalizedcache) re. auto-eviction policies
* Updates another note in the same page to use the new `<note>` tag


calvincestari:
This is what the note ends up looking like:
Screenshot 2024-04-08 at 3 20 55 PM

PR #333, opened 2024-04-15
feature: Enable visionOS Support

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3320.

With [SQLite.swift being updated to support visionOS](https://github.com/stephencelis/SQLite.swift/releases/tag/0.15.1) this:
* Updates our SQLite.swift dependency version
* Adds `visionOS` platform declarations
* Updates other dependency versions

### DIFF
--- a/apollo-ios-codegen/Package.swift
+++ b/apollo-ios-codegen/Package.swift
@@ -19,10 +19,10 @@ let package = Package(
       .upToNextMajor(from: "1.0.0")),
     .package(
       url: "https://github.com/apple/swift-collections",
-      .upToNextMajor(from: "1.0.6")),
+      .upToNextMajor(from: "1.1.0")),
     .package(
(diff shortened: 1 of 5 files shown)


apollo-ios-codegen/Package.swift
```diff
@@ -19,10 +19,10 @@ let package = Package(
       .upToNextMajor(from: "1.0.0")),
     .package(
       url: "https://github.com/apple/swift-collections",
-      .upToNextMajor(from: "1.0.6")),
+      .upToNextMajor(from: "1.1.0")),
```
calvincestari, line 22: I'm on the fence about whether we should include these here or not? I could easily be persuaded to revert this change.

AnthonyMDev (approved):
Excited to get this done!

loganblevins:
Nice work!!

PR #337, opened 2024-04-17
defer: Deprecate `legacyResponse`

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3370.

* Deprecates `legacyResponse`
* Stores the cache keys which are a result of the JSON interceptor parsing


apollo-ios/Sources/Apollo/HTTPResponse.swift
```diff
@@ -12,22 +12,31 @@ public class HTTPResponse<Operation: GraphQLOperation> {
...
   
-  /// [optional] The data as parsed into a `GraphQLResult`, which can eventually be returned to the UI. Will be nil if not yet parsed.
+  /// [optional] The data as parsed into a `GraphQLResult`, which can eventually be returned to the UI. Will be nil 
+  /// if not yet parsed.
   public var parsedResponse: GraphQLResult<Operation.Data>?
   
-  /// [optional] The data as parsed into a `GraphQLResponse` for legacy caching purposes. If you're not using the `JSONResponseParsingInterceptor`, you probably shouldn't be using this property.
+  /// [optional] The data as parsed into a `GraphQLResponse` for legacy caching purposes. If you're not using the 
+  /// `JSONResponseParsingInterceptor`, you probably shouldn't be using this property.
   /// **NOTE:** This property will be removed when the transition to the Swift Codegen is complete.
```
AnthonyMDev, line 21: We should remove this `**NOTE**` since we've don't that transition and now are only not removing it for backwards compatibility.

apollo-ios/Sources/Apollo/IncrementalGraphQLResponse.swift
```diff
@@ -27,20 +27,73 @@ final class IncrementalGraphQLResponse<Operation: GraphQLOperation> {
   private let base: AnyGraphQLResponse
 
   public init(operation: Operation, body: JSONObject) throws {
-    guard let path = body["path"] as? [JSONValue] else {
-      throw ResponseError.missingPath
-    }
+    guard let path = body["path"] as? [JSONValue] else { throw ResponseError.missingPath }
```
AnthonyMDev, line 30: Why did you remove a bunch of new line indentation throughout this PR?

AnthonyMDev (approved):
This LGTM! Crazy how little code was actually needed to do this and get such a big performance optimization. We should have done this a long time ago. Thanks so much @calvincestari!

I'm not sure why you changed the indentation styling of a bunch of the code in this PR as well. I prefer the style with params on individual lines, but since we haven't adopted an official style guideline and linter rules, I'm not going to request changes for that. That is something we really should do though 😅 .

apollo-ios/Sources/Apollo/HTTPResponse.swift
calvincestari: Ah, good catch!

apollo-ios/Sources/Apollo/IncrementalGraphQLResponse.swift
calvincestari: Just personal preference, I'll revert and change my new code to match the style it was before.

My preferences:
* If a function call can fit on a single line within the 120 column width then make it one line.
* If there is only a single instruction between `{` and `}` put it all on a single line too provided it fits within the 120 column limit.

calvincestari:
> This LGTM! Crazy how little code was actually needed to do this and get such a big performance optimization. We should have done this a long time ago.

I still need to:
* add fast parsing tests for `IncrementalGraphQLResponse`
* clean up the failing codegen tests due to warnings about an unused parameter in the incremental interceptor
* capture performance tests of eliminating the double parsing when writing to the cache

apollo-ios/Sources/Apollo/IncrementalGraphQLResponse.swift
```diff
@@ -40,9 +40,72 @@ final class IncrementalGraphQLResponse<Operation: GraphQLOperation> {
...
+    case .fetchIgnoringCacheCompletely:
+      // There is no cache, so we don't need to get any info on dependencies. Use fast parsing.
+      return (try parseIncrementalResultFast(), nil)
+
+    default:
+      return try parseIncrementalResult()
+    }
+  }
+
+  private func parseIncrementalResult() throws -> (IncrementalGraphQLResult, RecordSet?) {
```
calvincestari, line 63: Ideally the `parseResult` and `parseResultFast` functions in GraphQLResult would be `internal` rather than the `public` scope they have now but that would be a potentially breaking change. These are scoped minimally for now but we can always loosen that if needed.

PR #345, opened 2024-04-24
Watcher refetch policy

jimisaacs:
Sorry, I don't have much of a description here. I've just been having some difficulty rationalizing the Apollo-iOS watcher with Apollo-Kotlin's. I am aware that Apollo-Kotlin has a concept of both a fetchPolicy and a refetchPolicy, and the watcher of Apollo-Kotlin defaults to cache-only for refetch. This is what we want on iOS. We want to be able to invoke client.watch(), while providing both an initial fetch cache policy, and the refetch policy for how to determine what to do when the keys change. Basically allow the network on initial fetch, but never hit the network when handling a change.
I was looking through the Apollo-iOS code to see if this could be a non-breaking contribution, but I am realizing the difficulty of that and am about to write my own watcher to handle our needs. Any thoughts or suggestions would be appreciated.

My current implementation only tries to allow wrapping implementations to inject a refetchPolicy but not have any sort of default value so that `nil` means that it just uses current behavior. However what would actually make sense to lean into the same defaults as Apollo-Kotlin, i.e. cache-only for a watcher refetch.

### DIFF
--- a/apollo-ios/Sources/Apollo/ApolloClient.swift
+++ b/apollo-ios/Sources/Apollo/ApolloClient.swift
@@ -94,11 +94,13 @@ extension ApolloClient: ApolloClientProtocol {
 
   public func watch<Query: GraphQLQuery>(query: Query,
                                          cachePolicy: CachePolicy = .default,
+                                         refetchCachePolicy: CachePolicy? = nil,
                                          context: RequestContext? = nil,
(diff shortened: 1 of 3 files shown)


AnthonyMDev:
After conversations with @jimisaacs , we're going to go in a different direction with this. Closing this PR. Thanks!

PR #346, opened 2024-04-25
GraphQLQueryWatcher: `publisher` property

Iron-Ham:

I noticed that in situations where we have a large number of local cache updates and network cache updates, that we end up refreshing the UI far too often. By being able to treat the results of cache updates as a `Publisher`, we can place `throttle` or `debounce` functionality into the watcher's update stream. 

This PR introduces the following changes:
* [`apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift`](diffhunk://#diff-642389f551252b1671c22b4d281628bd3472fe810c25974f01a72257b781ad4dR178-R218): The `GraphQLQueryWatcher` class was extended to include a `CachePublisher` struct and a `WatcherSubscription` class, enabling the watcher to publish cache updates to subscribers. A `publisher()` method was also added to allow access to the `CachePublisher`.
  - This was chosen as an alternative to making the `GraphQLQueryWatcher` a `Publisher` itself, as it is an observer of the underlying cache. 
* [`apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift`](diffhunk://#diff-642389f551252b1671c22b4d281628bd3472fe810c25974f01a72257b781ad4dR33-R97): The `GraphQLQueryWatcher` class was refactored to include convenience initializers and a private initializer. The `resultHandler` was updated to trigger subscriptions when a result is received.
  - The original initializer was effectively left unchanged, meaning that there is no breaking change to existing consumers. 
  - A new initializer was added that does not take a result handler. This initializer expects users to then utilize the `publisher` property. 
* [`apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift`](diffhunk://#diff-642389f551252b1671c22b4d281628bd3472fe810c25974f01a72257b781ad4dL12-R13): The `resultHandler` property in the `GraphQLQueryWatcher` class was changed from a constant to an optional variable.
* [`apollo-ios/Package.swift`](diffhunk://#diff-3a2cd2d70a6f8cf86d3d3783c8f4d3c6ab2b29e3fc6524f9c9ca12dd513b6603L10-R10): **The minimum macOS version for the Apollo package was updated from 10.14 to 10.15**. This is because `Combine` is unsupported on 10.14. **This constitutes a breaking change for macOS 10.14 deployments**. 

Employing this on the `GraphQLQueryWatcher` feels like the correct choice, as opposed to doing it in a [roundabout manner](https://github.com/joel-perry/ApolloCombine/tree/master?tab=readme-ov-file#usage) on the `ApolloClient`, but would require a version minimum increase to macOS 10_15. 

**NOTE**: I didn't tackle any macOS 10.15 deprecations. 

### Alternatives Considered:

- As per the OSS `ApolloCombine` project: an extension on `ApolloClient`. I chose to avoid this route because: it wouldn't integrate well with `ApolloPagination`, it would be wasteful if you want multiple subscribers on the same watcher.
- Pulling the behavior of `GraphQLQueryWatcher` into a protocol / creating a subclass of `GraphQLQueryWatcher` that has this behavior. I chose not to do this since it seems less discoverable, and perhaps a bit arbitrary. It has the advantage of not needing a version bump.

### DIFF
--- a/Tests/ApolloTests/Cache/WatchQueryTests.swift
+++ b/Tests/ApolloTests/Cache/WatchQueryTests.swift
@@ -120,7 +120,115 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
       wait(for: [serverRequestExpectation, refetchedWatcherResultExpectation], timeout: Self.defaultWaitTimeout)
     }
   }
-  
+
+  func testRefetchWatchedQueryFromServerThroughWatcherReturnsRefetchedResults_sink() throws {
+    class SimpleMockSelectionSet: MockSelectionSet {
+      override class var __selections: [Selection] { [
+        .field("hero", Hero.self)
+      ]}
+
+      class Hero: MockSelectionSet {
+        override class var __selections: [Selection] {[
+          .field("__typename", String.self),
+          .field("name", String.self)
+        ]}
+      }
+    }
+
+    let watchedQuery = MockQuery<SimpleMockSelectionSet>()
+
+    let resultObserver = makeResultObserver(for: watchedQuery)
+
+    let watcher = GraphQLQueryWatcher(client: client, query: watchedQuery, resultHandler: resultObserver.handler)
+    var results: [GraphQLQueryWatcher<MockQuery<SimpleMockSelectionSet>>.CachePublisher.Output] = []
+    var subscription = watcher.publisher().sink { result in
+      results.append(result)
+    }
+    addTeardownBlock {
+      subscription.cancel()
+      watcher.cancel()
+    }
+
+    runActivity("Initial fetch from server") { _ in
+      let serverRequestExpectation =
+        server.expect(MockQuery<SimpleMockSelectionSet>.self) { request in
+          [
+            "data": [
+              "hero": [
+                "name": "R2-D2",
+                "__typename": "Droid"
+              ]
+            ]
+          ]
+        }
+
+      let initialWatcherResultExpectation =
+        resultObserver.expectation(
+          description: "Watcher received initial result from server"
+        ) { result in
+          try XCTAssertSuccessResult(result) { graphQLResult in
+            XCTAssertEqual(graphQLResult.source, .server)
+            XCTAssertNil(graphQLResult.errors)
+
+            let data = try XCTUnwrap(graphQLResult.data)
+            XCTAssertEqual(data.hero?.name, "R2-D2")
+          }
+
+          let latestSinkResult = try XCTUnwrap(results.last)
+          try XCTAssertSuccessResult(latestSinkResult) { graphQLResult in
+            XCTAssertEqual(graphQLResult.source, .server)
+            XCTAssertNil(graphQLResult.errors)
+
+            let data = try XCTUnwrap(graphQLResult.data)
+            XCTAssertEqual(data.hero?.name, "R2-D2")
+          }
+      }
+
+      watcher.fetch(cachePolicy: .fetchIgnoringCacheData)
+
+      wait(for: [serverRequestExpectation, initialWatcherResultExpectation], timeout: Self.defaultWaitTimeout)
+    }
+
+    runActivity("Refetch from server") { _ in
+      let serverRequestExpectation =
+        server.expect(MockQuery<SimpleMockSelectionSet>.self) { request in
+          [
+            "data": [
+              "hero": [
+                "name": "Artoo",
+                "__typename": "Droid"
+              ]
+            ]
+          ]
+        }
+
+      let refetchedWatcherResultExpectation = resultObserver.expectation(
+        description: "Watcher received refetched result from server"
+      ) { result in
+        try XCTAssertSuccessResult(result) { graphQLResult in
+          XCTAssertEqual(graphQLResult.source, .server)
+          XCTAssertNil(graphQLResult.errors)
+
+          let data = try XCTUnwrap(graphQLResult.data)
+          XCTAssertEqual(data.hero?.name, "Artoo")
+        }
+
+        let latestSinkResult = try XCTUnwrap(results.last)
+        try XCTAssertSuccessResult(latestSinkResult) { graphQLResult in
+          XCTAssertEqual(graphQLResult.source, .server)
+          XCTAssertNil(graphQLResult.errors)
+
+          let data = try XCTUnwrap(graphQLResult.data)
+          XCTAssertEqual(data.hero?.name, "Artoo")
+        }
+      }
+
+      watcher.refetch()
+
+      wait(for: [serverRequestExpectation, refetchedWatcherResultExpectation], timeout: Self.defaultWaitTimeout)
+    }
+  }
+
   func testWatchedQueryGetsUpdatedAfterFetchingSameQueryWithChangedData() throws {
     class SimpleMockSelectionSet: MockSelectionSet {
       override class var __selections: [Selection] { [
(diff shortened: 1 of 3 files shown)


apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -29,26 +30,71 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
...
+      handler: nil
+    )
+  }
+
+  init(
+    client: ApolloClientProtocol,
+    query: Query,
+    context: RequestContext? = nil,
+    callbackQueue: DispatchQueue = .main,
+    handler: GraphQLResultHandler<Query.Data>?
```
AnthonyMDev, line 85: I would think the handler would need to be `@escaping` here also. What am I missing?

apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -129,3 +175,44 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
     }
   }
 }
+
+extension GraphQLQueryWatcher {
+  public struct CachePublisher: Publisher {
+    public typealias Output = Result<GraphQLResult<Query.Data>, Error>
+    public typealias Failure = Never
+
+    var watcher: GraphQLQueryWatcher
```
AnthonyMDev, line 184: Thinking this should be a `let` right?

Tests/ApolloTests/Cache/WatchQueryTests.swift
```diff
@@ -120,7 +120,115 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
...
+      }
+    }
+
+    let watchedQuery = MockQuery<SimpleMockSelectionSet>()
+
+    let resultObserver = makeResultObserver(for: watchedQuery)
+
+    let watcher = GraphQLQueryWatcher(client: client, query: watchedQuery, resultHandler: resultObserver.handler)
+    var results: [GraphQLQueryWatcher<MockQuery<SimpleMockSelectionSet>>.CachePublisher.Output] = []
+    var subscription = watcher.publisher().sink { result in
```
AnthonyMDev, line 144: Make this `let` please.

AnthonyMDev (changes requested):
I'm not opposed to this addition, but I want to make sure we think through it. 

I'm unclear on how easy it is to create a retain cycle here. There will definitely be a retain on the watcher, but since the watcher holds on to the `subscriptions` will that cause a retain cycle in any situation? What if the subscriber block accesses the `watcher` itself? I guess that would be broken by calling `cancel` though, right?

Iron-Ham:
🤔 
While this is potentially directionally correct – I'm not sure it's the best solution for the problem. I spoke with @eliperkins earlier today, who suggested a mechanism like FMDB's [queueing mechanism](https://github.com/ccgus/fmdb?tab=readme-ov-file#nullability-and-swift-optionals), in conjunction with the concept of a cache update timestamp.

___

Imagine the following scenario:

A user takes five rapid actions, each triggering an optimistic cache update _and_ queuing a network request which makes the mutation. The mutation responds with the updated data, and in the event of failure rolls back the optimistic update. The user actions must have optimistic updates attached, as the UI would otherwise be out-of-sync. The server responses, however, would contain stale data relative to other optimistic updates – which would blow out the changes.

This PR attempts to solve this problem by giving consumers of a watcher the ability to debounce or throttle outputs. While that may be a desired behavior in another context, it seems like it shouldn't be the primary solution. Instead, what if each updated value had an `updated_at` timestamp that gets committed by a new `Client` method? That way, we could have an API that looks something like this:

```swift
func perform<Mutation: GraphQLMutation, T>(
  mutation: Mutation,
  publishResultToStore: Bool,
  context: RequestContext?,
  queue: DispatchQueue,
  optimisticUpdate: ((ReadWriteTransaction) throws -> T)?, // NEW: optional `optimisticUpdate` parameter
  resultHandler: GraphQLResultHandler<Mutation.Data>?
) -> Cancellable
```

This way, the client can:
- Trigger the optimistic update, with an internally stored (or stored within sqlite?) timestamp of when the field was updated.
- When the mutation response is received, it uses the timestamp of the optimistic mutation (if one was provided; else, it uses the current time) to determine if a given value should be overwritten. 


With all of this in mind, the (success) path would look something like this:

A user takes five rapid actions, each triggering an optimistic cache update _and_ queuing a network request which makes the mutation. Each optimistic event is tied to a given timestamp, and only data whose timestamp is equal (or later than) the previous data's timestamp can overwrite. Thus, when a mutation responds, it only overwrites data that has no timestamp (existing cached data) or confirms the optimistic update by overwriting data with a timestamp equal-to-or-less-than its timestamp. After five mutations, the cache state remains correct throughout the process, even for actions that have not yet been confirmed by the server. If the consumer of the API wants throttling or debouncing, that seems like a separate (and separately solved) problem.

There's still the matter of rollbacks, and that's something to think about in this API. We should have a way of exposing a rollback to the user: this is a common use-case and I'm certain we can take notes from how `apollo-react` has gone about this problem. 

What do you think, @AnthonyMDev?

Iron-Ham:
🤔 
While this is potentially directionally correct – I'm not sure it's the best solution for the problem. I spoke with @eliperkins earlier today, who suggested a mechanism like FMDB's [queueing mechanism](https://github.com/ccgus/fmdb?tab=readme-ov-file#nullability-and-swift-optionals), in conjunction with the concept of a cache update timestamp.

___

Imagine the following scenario:

A user takes five rapid actions, each triggering an optimistic cache update _and_ queuing a network request which makes the mutation. The mutation responds with the updated data, and in the event of failure rolls back the optimistic update. The user actions must have optimistic updates attached, as the UI would otherwise be out-of-sync. The server responses, however, would contain stale data relative to other optimistic updates – which would blow out the changes.

This PR attempts to solve this problem by giving consumers of a watcher the ability to debounce or throttle outputs. While that may be a desired behavior in another context, it seems like it shouldn't be the primary solution. Instead, what if each updated value had an `updated_at` timestamp that gets committed by a new `Client` method? That way, we could have an API that looks something like this:

```swift
func perform<Mutation: GraphQLMutation, T>(
  mutation: Mutation,
  publishResultToStore: Bool,
  context: RequestContext?,
  queue: DispatchQueue,
  optimisticUpdate: ((ReadWriteTransaction) throws -> T)?, // NEW: optional `optimisticUpdate` parameter
  resultHandler: GraphQLResultHandler<Mutation.Data>?
) -> Cancellable
```

This way, the client can:
- Trigger the optimistic update, with an internally stored (or stored within sqlite?) timestamp of when the field was updated.
- When the mutation response is received, it uses the timestamp of the optimistic mutation (if one was provided; else, it uses the current time) to determine if a given value should be overwritten. 


With all of this in mind, the (success) path would look something like this:

A user takes five rapid actions, each triggering an optimistic cache update _and_ queuing a network request which makes the mutation. Each optimistic event is tied to a given timestamp, and only data whose timestamp is equal (or later than) the previous data's timestamp can overwrite. Thus, when a mutation responds, it only overwrites data that has no timestamp (existing cached data) or confirms the optimistic update by overwriting data with a timestamp equal-to-or-less-than its timestamp. After five mutations, the cache state remains correct throughout the process, even for actions that have not yet been confirmed by the server. If the consumer of the API wants throttling or debouncing, that seems like a separate (and separately solved) problem.

There's still the matter of rollbacks, and that's something to think on in this API. We should have a way of exposing a rollback to the user: this is a common use-case and I'm certain we can take notes from how `apollo-react` has gone about this problem.

PR #347, opened 2024-04-25
Add refetchOnFailedUpdates to query watcher

AnthonyMDev:
This PR adds a `refetchOnFailedUpdates` to `GraphQLQueryWatcher`. When this is set to `false`, network fetches will no longer occur if the cache load for updates to a watched object fails. This features was added at the request of @jimisaacs.

### DIFF
--- a/Tests/ApolloTests/Cache/WatchQueryTests.swift
+++ b/Tests/ApolloTests/Cache/WatchQueryTests.swift
@@ -826,7 +826,7 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
     }
   }
 
-  func testWatchedQuery_givenRefetchOnFailedUpdates_false_doesNotrefetchFromServerAfterOtherQueryUpdatesListWithIncompleteObject() throws {
+  func testWatchedQuery_givenRefetchOnFailedUpdates_false_doesNotRefetchFromServerAfterOtherQueryUpdatesListWithIncompleteObject() throws {
     class HeroAndFriendsIDsSelectionSet: MockSelectionSet {
       override class var __selections: [Selection] {[
         .field("hero", Hero?.self)
--- a/Tests/ApolloTests/Cache/WatchQueryTests.swift
+++ b/Tests/ApolloTests/Cache/WatchQueryTests.swift
@@ -826,6 +826,146 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
     }
   }
 
+  func testWatchedQuery_givenRefetchOnFailedUpdates_false_doesNotrefetchFromServerAfterOtherQueryUpdatesListWithIncompleteObject() throws {
+    class HeroAndFriendsIDsSelectionSet: MockSelectionSet {
+      override class var __selections: [Selection] {[
+        .field("hero", Hero?.self)
+      ]}
+
+      var hero: Hero? { __data["hero"] }
+
+      class Hero: MockSelectionSet {
+        override class var __selections: [Selection] {[
+          .field("__typename", String.self),
+          .field("id", String.self),
+          .field("friends", [Friend]?.self),
+        ]}
+
+        var friends: [Friend]? { __data["friends"] }
+
+        class Friend: MockSelectionSet {
+          override class var __selections: [Selection] {[
+            .field("__typename", String.self),
+            .field("id", String.self),
+          ]}
+        }
+      }
+    }
+
+    class HeroAndFriendsNameWithIDsSelectionSet: MockSelectionSet {
+      override class var __selections: [Selection] {[
+        .field("hero", Hero?.self)
+      ]}
+
+      var hero: Hero? { __data["hero"] }
+
+      class Hero: MockSelectionSet {
+        override class var __selections: [Selection] {[
+          .field("__typename", String.self),
+          .field("id", String.self),
+          .field("name", String.self),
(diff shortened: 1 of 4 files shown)


apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -35,16 +42,20 @@ public final class GraphQLQueryWatcher<Query: GraphQLQuery>: Cancellable, Apollo
...
   ///   - client: The client protocol to pass in.
   ///   - query: The query to watch.
+  ///   - refetchOnFailedUpdates: Should the watcher perform a network fetch when it's watched
+  ///     objects have changed, but reloading them from the cache fails. Defaults to `true`.
   ///   - context: [optional] A context that is being passed through the request chain. Defaults to `nil`.
   ///   - callbackQueue: The queue for the result handler. Defaults to the main queue.
   ///   - resultHandler: The result handler to call with changes.
   public init(client: ApolloClientProtocol,
               query: Query,
+              refetchOnFailedUpdates: Bool = true,
```
jimisaacs, line 52: @AnthonyMDev What do you think of `ApolloClient.watch` passing this? If you have opinions against it, I can skip calling a wrapped `ApolloClient` straight to use `GraphQLQueryWatcher` directly.

AnthonyMDev: Oh good call. I should definitely add that to this PR!

apollo-ios/Sources/Apollo/ApolloClientProtocol.swift
```diff
@@ -39,12 +39,15 @@ public protocol ApolloClientProtocol: AnyObject {
...
   ///   - cachePolicy: A cache policy that specifies when results should be fetched from the server or from the local cache.
+  ///   - refetchOnFailedUpdates: Should the watcher perform a network fetch when it's watched
+  ///     objects have changed, but reloading them from the cache fails. Should default to `true`.
   ///   - context: [optional] A context that is being passed through the request chain. Should default to `nil`.
   ///   - callbackQueue: A dispatch queue on which the result handler will be called. Should default to the main queue.
   ///   - resultHandler: [optional] A closure that is called when query results are available or when an error occurs.
   /// - Returns: A query watcher object that can be used to control the watching behavior.
   func watch<Query: GraphQLQuery>(query: Query,
                                   cachePolicy: CachePolicy,
+                                  refetchOnFailedUpdates: Bool,
```
calvincestari, line 50: Are we considering this a breaking change since it would impact users that have written their own `ApolloClientProtocol` conforming client?

Tests/ApolloTests/Cache/WatchQueryTests.swift
```diff
@@ -826,6 +826,146 @@ class WatchQueryTests: XCTestCase, CacheDependentTesting {
     }
   }
 
+  func testWatchedQuery_givenRefetchOnFailedUpdates_false_doesNotrefetchFromServerAfterOtherQueryUpdatesListWithIncompleteObject() throws {
```
calvincestari, line 829:
```suggestion
  func testWatchedQuery_givenRefetchOnFailedUpdates_false_doesNotRefetchFromServerAfterOtherQueryUpdatesListWithIncompleteObject() throws {
```

calvincestari (approved):

apollo-ios/Sources/Apollo/ApolloClientProtocol.swift
AnthonyMDev: Ah that's a great point. I've decided to not add the parameter to `ApolloClientProtocol` since it's pretty minor and fix this with a deprecated function on `ApolloClient`. How do you feel about this fix? 

It's a little awkward that the new parameter isn't on the protocol, but honestly, the whole protocol is kind of a painful solution to a problem that should be fixed by us building a `MockApolloClient` for testing and maybe just making `ApolloClient` `open`. That's work for 2.0 though.

calvincestari: > I've decided to not add the parameter to ApolloClientProtocol since it's pretty minor and fix this with a deprecated function on ApolloClient. How do you feel about this fix?

Looks good. I think it's a better solution than breaking existing implementations for something that isn't crucial.

jimisaacs: @AnthonyMDev @calvincestari 

We have wrapped `ApolloClient`, but are currently accepting an `ApolloClientProtocol`. Though I think we can change that to just `ApolloClient`.

I am realizing how _thin_ `ApolloClient` actually is, and have been considering just foregoing wrapping an `ApolloClient` and `ApolloClientProtocol` at all, and just having our own `NetworkTransport` that wraps `RequestChainNetworkTransport` internally.

Just letting you know, if you feel like you'd like to just keep the `ApolloClient` change out (in another commit to main of course), I am fine just using `GraphQLQueryWatcher` directly if you can be certain that `ApolloClient` can remain as thin as it is now until a 2.0/3.0 level of change.

jimisaacs: Actually a rescind that statement, `GraphQLQueryWatcher` requires an argument conforming to `ApolloClientProtocol`.

PR #349, opened 2024-05-01
defer: Adds incremental response cache writes

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3362.

With only minor changes we've got cache writing successfully working for the partial and incremental responses of operations using `@defer`.
* Adds storage for cache records to `IncrementalJSONResponseParsingInterceptor` so they can be merged when multiple incremental results are received in the same response.
* Removes the early-exit in `CacheWriteInterceptor` to disable cache writes for operations using `@defer`.
* Adds tests for incremental cache writes.

### DIFF
--- a/Tests/ApolloTests/Interceptors/IncrementalJSONResponseParsingInterceptorTests.swift
+++ b/Tests/ApolloTests/Interceptors/IncrementalJSONResponseParsingInterceptorTests.swift
@@ -367,7 +367,105 @@ class IncrementalJSONResponseParsingInterceptorTests: XCTestCase {
 
   // MARK: Cache Records Tests
 
-  func test__cacheRecords__givenMultipleIncrementalResults_shouldComputeCacheRecordsMatchingResults() throws {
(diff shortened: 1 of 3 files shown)


AnthonyMDev (approved):
This looks good! It just seems suspicious to me that this was so simple haha :) I have some worries that we are missing some edge cases or something here, but I can't figure out where right now, so let's move forward and address them when/if they arise.

calvincestari:
> This looks good! It just seems suspicious to me that this was so simple haha :)

Much easier than I anticipated too. There is definitely more complexity in seeding a defer operation-fetch with cache data, which is still in progress.

> I have some worries that we are missing some edge cases or something here, but I can't figure out where right now, so let's move forward and address them when/if they arise.

I agree, I don't think this should stop us merging and moving forward. Possible future features for defer operation-caching is maybe the user doesn't ever want to cache a particular deferred response. Let's wait to gather feedback on that first though.

PR #353, opened 2024-05-09
fix (defer): deferred fragment condition evaluation

calvincestari:
Conditions on a deferred fragment were not correctly being evaluated during field collection. This meant that deferred fragments would always be collected into `_deferredFragments` regardless of any condition. It's now fixed and deferred fragments are not collected if the condition would evaluate to false meaning the deferred fragment should never be expected to be received.

### DIFF
--- a/apollo-ios/Sources/Apollo/FieldSelectionCollector.swift
+++ b/apollo-ios/Sources/Apollo/FieldSelectionCollector.swift
@@ -101,10 +101,12 @@ struct DefaultFieldSelectionCollector: FieldSelectionCollector {
         //
         // The deferred fragment identifiers still need to be collected becuase that is how the
         // user determines the state of the deferred fragment via the @Deferred property wrapper.
-        var collectDeferredFragment: Bool = true
-        if let condition {
-          collectDeferredFragment = condition.evaluate(with: info.variables)
-        }
+        let isDeferred: Bool = {
+          if let condition, !condition.evaluate(with: info.variables) {
+            return false
+          }
+          return true
+        }()
 
         if collectDeferredFragment {
           groupedFields.addDeferredFragment(typeCase)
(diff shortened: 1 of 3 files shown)


AnthonyMDev (commented):
This behavior is still wrong I think. If the conditions evaluate to `false`, the fragment will still be executed, it just won't be deferred. We need to treat it as a non-deferred fragment in that case, not ignore it completely.

calvincestari:
Yes, you're right. 🤦‍♂️

I'll correct this when I'm back on Monday.

calvincestari:
@AnthonyMDev this is now fixed with an additional test to ensure that a `missingValue` error is thrown during execution if the deferred data is not present with a `false` condition.

apollo-ios/Sources/Apollo/FieldSelectionCollector.swift
```diff
@@ -93,15 +93,26 @@ struct DefaultFieldSelectionCollector: FieldSelectionCollector {
...
+      case let .deferred(condition, typeCase, _):
         // In Apollo's implementation (Router + Server) of deferSpec=20220824 ALL defer directives
         // will be honoured and sent as separate incremental responses. This means deferred
         // selection fields never need to be collected because they are parsed with the incremental
         // data, at which time they are no longer deferred.
         //
         // The deferred fragment identifiers still need to be collected becuase that is how the
         // user determines the state of the deferred fragment via the @Deferred property wrapper.
-        groupedFields.addDeferredFragment(typeCase)
+        var collectDeferredFragment: Bool = true
```
AnthonyMDev, line 104: I think we should change this to be named `isDeferred` with a comment saying, if the condition resolves to false, the fragment is collected normally and not deferred.

apollo-ios/Sources/Apollo/FieldSelectionCollector.swift
```diff
@@ -93,15 +93,26 @@ struct DefaultFieldSelectionCollector: FieldSelectionCollector {
...
+        if let condition {
+          collectDeferredFragment = condition.evaluate(with: info.variables)
+        }
```
AnthonyMDev, line 107:
```suggestion
        let isDeferred: Bool = {
          if let condition, !condition.evaluate(with: info.variables) {
            return false
          }
          return true
        }()
```

AnthonyMDev: I usually prefer a computed `let` to a `var` in situations like these (though that's not always possible).

AnthonyMDev (approved):
Looks good now! Minor suggestions on the implementation, but the functionality is correct.

PR #358, opened 2024-05-14
Generate nil default value for input field w/default value in schema

AnthonyMDev:
Fixes [#2997](https://github.com/apollographql/apollo-ios/issues/2997).

The Kotlin codegen generates the parameter for that field with a default value of nil, so that if the user doesn’t pass a value, the field is omitted from the request and the schema side default value is used. This provides parity with Kotlin client functionality.


calvincestari (approved):
Nice one! Thank you so much for doing this. ❤️

PR #359, opened 2024-05-14
Namespace ApolloAPI in InterfaceTemplate

AnthonyMDev:
Fixes [#3375](https://github.com/apollographql/apollo-ios/issues/3375)

`Interface` should be namespaced to `ApolloAPI`.


calvincestari (approved):
Looks good, thank you. Are we just sticking with fixing `Interface` for now and wait to see if the others are a problem before a preemptive fix? `Union`, `InputObject` and maybe one or two others could have the same problem.

PR #361, opened 2024-05-16
chore: Update deprecated `--schema-name` property references

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3382.

There isn't much detail in that issue on what the actual problem is so I did an audit on where we might still be using `--schema-name` and found two outdated references; migration guide and CI action. This PR simply updates those to use the newer `--schema-namespace`.


calvincestari:
Merging without approval since it's a known change to documentation and a CI command.

PR #363, opened 2024-05-17
Support ID scalar as a custom scalar

AnthonyMDev:
This changes the generation of the ID scalar to be treated as a custom scalar that can be modified by the user.

This is implemented by request from [#3379](https://github.com/apollographql/apollo-ios/issues/3379).


calvincestari:
This looks good, thanks for doing it so quickly. It seems to be failing on one of the codegen test configurations though. 🤔

calvincestari (approved):

apollo-bot2:

PR #369, opened 2024-05-23
ci: Update environment to latest versions

calvincestari:
When adding the CI job to #367 I noticed that our CI uses older versions than what we use on our development machines and also older than what is available on GitHub.

This PR includes:
* Update to use the latest macos image available instead of a static version
* Update to use Xcode 15.4


calvincestari:
@BobaFetters is there a way for us to force all CI jobs to run and not depend on the file tree changes? Looks like some are being skipped here but I'd prefer the confidence of having them all run.

BobaFetters:
@calvincestari Yea we can definitely do that, just need to determine the parameters of when we would want to do that vs what we do for most PRs. Like our release-check workflow always runs everything, so if we have something we can key off of in a PR like a specific title etc we could set it up pretty easy

BobaFetters (approved):

PR #371, opened 2024-05-25
[ApolloPagination] Make `isLoadingAll` publicly accessible, remove special logic

Iron-Ham:
> [!CAUTION]
> This is a breaking change!

This pull request introduces the `isLoadingAll` property to both `GraphQLQueryPager` and `AsyncGraphQLQueryPager`.  Additionally, it removes the logic in the `AsyncGraphQLQueryPagerCoordinator`'s `subscribe` function which intentionally does not forward updates to subscribers while `loadingAll`. 

The existing logic was built on the assumption that callers of `loadAll` will not want any updates until we can furnish all data. However, with a public `isLoadingAll` property, consumers of the API can manage this behavior themselves. This does mean that users will receive each page as it comes in via the `loadAll` function unless they specifically want to defer the update (through use of the `isLoadingAll` property).

### DIFF
--- a/Tests/ApolloPaginationTests/AsyncGraphQLQueryPagerTests.swift
+++ b/Tests/ApolloPaginationTests/AsyncGraphQLQueryPagerTests.swift
@@ -347,7 +347,9 @@ final class AsyncGraphQLQueryPagerTests: XCTestCase {
     let lastPageExpectation = Mocks.Hero.FriendsQuery.expectationForSecondPage(server: server)
     try await pager.loadAll()
     await fulfillment(of: [firstPageExpectation, lastPageExpectation], timeout: 5)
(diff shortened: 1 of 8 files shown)


AnthonyMDev:
Let's continue to discuss the desired behavior for this in [#3387](https://github.com/apollographql/apollo-ios/issues/3387) before deciding on how to proceed here.

calvincestari:
@Iron-Ham - is this still dependent on work being done in other issues/PRs, is there anything you need from us here?

Iron-Ham:
@calvincestari I think this is still the right thing to do – but I can re-open this after https://github.com/apollographql/apollo-ios-dev/pull/428 is merged.

PR #372, opened 2024-05-25
Highlight unused code in Pull Requests as a warning

Iron-Ham:
This pull request adds a new action which runs Periphery to identify unused code within a PR. 

Screenshot 2024-05-25 at 2 32 15 PM


Iron-Ham:
Screenshot 2024-05-25 at 2 32 15 PM

calvincestari:
Looks like the new job didn't get triggered in this PR, I'd like to see how long it takes to execute. Is this solving a specific problem you've noticed or just a general codebase hygiene tool you use elsewhere?

Iron-Ham:
Apologies for the late response @calvincestari, travel will do that 😓 

It ran here: https://github.com/apollographql/apollo-ios-dev/actions/runs/9237413673/job/25414361878 and took about two minutes. It could be optimized further if we cached the Periphery install from Homebrew. I've set it so that it only runs on PRs that include Swift changes.

> Is this solving a specific problem you've noticed or just a general codebase hygiene tool you use elsewhere?

A bit of both! There isn't a _ton_ of unused code in Apollo (at least, with the configuration I've set in the `yml` file), but there is some. As far as the tool being used elsewhere, I've seen it used locally in various projects – but decided to automate it on CI via Actions shortly before leaving for vacation.

I would argue that before we consider merging this, we should update to the [new release](https://github.com/peripheryapp/periphery/releases/tag/2.20.0) which includes the [Actions formatter](https://github.com/peripheryapp/periphery/pull/746) so that we don't have to pipe this through a ruby script at all.

calvincestari:
I can see the value in it but I don't think this is a major issue in the codebase. I'm not super keen to have this enforced in every PR but wouldn't mind if it were moved to be a git hook instead so that whoever wants to can install and run locally for their PRs.

PR #375, opened 2024-05-28
Default task descriptions

ejensen:
This PR adds task descriptions based on the operation type, name and identifier. [Task descriptions](https://developer.apple.com/documentation/foundation/urlsessiontask/1409798-taskdescription) are helpful in disambiguation `URLSessionTask`s in Instruments and debugging in Xcode.

### DIFF
--- a/apollo-ios/Sources/Apollo/NetworkFetchInterceptor.swift
+++ b/apollo-ios/Sources/Apollo/NetworkFetchInterceptor.swift
@@ -36,7 +36,7 @@ public class NetworkFetchInterceptor: ApolloInterceptor, Cancellable {
       return
     }
     
-    var taskDescription = Operation.operationName
+    var taskDescription = "\(Operation.operationType) \(Operation.operationName)"
     if let operationIdentifier = Operation.operationIdentifier {
         taskDescription += " \(operationIdentifier)"
     }
--- a/apollo-ios/Sources/Apollo/NetworkFetchInterceptor.swift
+++ b/apollo-ios/Sources/Apollo/NetworkFetchInterceptor.swift
@@ -36,7 +36,11 @@ public class NetworkFetchInterceptor: ApolloInterceptor, Cancellable {
       return
     }
     
-    let task = self.client.sendRequest(urlRequest) { [weak self] result in
+    var taskDescription = Operation.operationName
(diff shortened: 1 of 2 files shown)


apollo-ios/Sources/Apollo/NetworkFetchInterceptor.swift
```diff
@@ -36,7 +36,11 @@ public class NetworkFetchInterceptor: ApolloInterceptor, Cancellable {
       return
     }
     
-    let task = self.client.sendRequest(urlRequest) { [weak self] result in
+    var taskDescription = "\(Operation.operationType) \(Operation.operationName)"
+    if let operationIdentifier = Operation.operationIdentifier {
```
AnthonyMDev, line 40: These operation identifiers are just the id's used for persisted queries. I don't think they are really useful in this context.

AnthonyMDev (changes requested):
Thanks so much for the PR! I think that we shouldn't add the `operationIdentifier` to the `taskDescription`, but otherwise, I'm good with this. If you make that change, I'm happy to merge this.

apollo-ios/Sources/Apollo/URLSessionClient.swift
```diff
@@ -152,7 +152,7 @@ open class URLSessionClient: NSObject, URLSessionDelegate, URLSessionTaskDelegat
     sendRequest(
       request,
       taskDescription: nil,
-      rawTaskCompletionHandler: nil,
+      rawTaskCompletionHandler: rawTaskCompletionHandler,
```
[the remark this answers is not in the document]
ejensen: It is indirectly related. I realized the `rawTaskCompletionHandler` parameter was not being forwarded correctly when overriding the `sendRequest()` method to construct a default `taskDescription` when the `taskDescription` parameter is `nil`.

apollo-ios/Sources/Apollo/NetworkFetchInterceptor.swift
ejensen: Sounds good to me. I removed it in commit 897be2f

AnthonyMDev (approved):
Thanks for this PR @ejensen!

PR #377, opened 2024-05-30
refactor: extracted DataDict.asJSONDictionary()

TizianoCoroneo:
Hey there 👋 

recently I found the need to encode a `DataDict` instance into a JSON object. I found that this functionality is already provided but only for `GraphQLResult` instances instead.

 I refactored the relevant code to also expose a `DataDict.asJSONDictionary()` method that performs the same logic as the one in `GraphQLResult`.

### DIFF
--- a/apollo-ios/Sources/Apollo/GraphQLResult.swift
+++ b/apollo-ios/Sources/Apollo/GraphQLResult.swift
@@ -55,10 +55,19 @@ extension GraphQLResult {
   /// - Returns: A `[String: Any]` JSON dictionary representing the ``GraphQLResult``.
   public func asJSONDictionary() -> [String: Any] {
     var dict: [String: Any] = [:]
-    if let data { dict["data"] = convert(value: data.__data) }
+    if let data { dict["data"] = data.__data.asJSONDictionary() }
     if let errors { dict["errors"] = errors.map { $0.asJSONDictionary() } }
     if let extensions { dict["extensions"] = extensions }
     return dict
+  }
+}
+
+extension DataDict {
+  /// Converts a ``DataDict`` into a basic JSON dictionary for use.
+  ///
+  /// - Returns: A `[String: Any]` JSON dictionary representing the ``DataDict``.
+  public func asJSONDictionary() -> [String: Any] {
+    _data.mapValues(convert(value:))
   }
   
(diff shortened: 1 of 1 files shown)


AnthonyMDev (changes requested):
Thanks for the PR @TizianoCoroneo!

The team discussed this and we'd like to request a change before merging this. The `DataDict` is not really intended for public consumption. It has to be public in order for your generated models to access it and compile properly, but we don't want to encourage people to use it directly.

We think this would be improved by making this accessible directly from `SelectionSet` instead of `DataDict`.

We also try not to cloud the namespace of `SelectionSet` too much because of naming collisions and code completion. I haven't talked to the team about this part yet, but I'm wondering if it would be better to have a global function or static struct for this. Something like `Apollo.JSONConverter.convert(_ selectionSet: ApolloAPI.SelectionSet) -> [String: Any]`.

We definitely want to do this on `SelectionSet`, not `DataDict`, but lets discuss the actual API for this before finalizing and merging.

@calvincestari @BobaFetters, any opinions/suggestions?

calvincestari:
> I'm wondering if it would be better to have a global function or static struct for this. Something like `Apollo.JSONConverter.convert(_ selectionSet: ApolloAPI.SelectionSet) -> [String: Any]`.

I hadn't considered the package namespace for this but I'm not opposed to it. I think it could help with clarity on where this feature is available.

BobaFetters:
Yea I like the idea of having available as a global like you suggested
`Apollo.JSONConverter.convert(_ selectionSet: ApolloAPI.SelectionSet) -> [String: Any]`

And I think we should use the `JSONConverter` to hold all of our helper functions that handle JSON conversion, maybe we go through in future PRs and deprecate some existing functions and bring it all into one place.

TizianoCoroneo:
I updated the PR. Now there is an enum with static methods to perform the conversions like this:
```swift
let jsonObject = Apollo.JSONConvert.selectionSetToJSONObject(myHero)
```

I also added a function to convert a `GraphQLResult` into a JSONObject that internally uses `asJSONDictionary`, just because the sight of one empty enum with one static method kind of hurts.

AnthonyMDev:
I've branched off of this PR and made a few changes. New PR is up #380. Thanks for the help @TizianoCoroneo!

PR #378, opened 2024-05-31
Set integration tests to run on macos-13

AnthonyMDev:
Integration tests need node 12.22.10 which does not have an image for arm64 devices. We need to rollback to using the macos-13 image for this job.


calvincestari (approved):
LGTM!

PR #379, opened 2024-06-04
Added Existential Any requirement

AnthonyMDev:
This PR adds the `-enable-upcoming-feature ExistentialAny` flag to all targets in the project and adds the `any` keyword at all necessary call sites.

The code generation templates are also updated to add the `any` keyword where needed.


apollo-ios-codegen/Package.swift
```diff
@@ -1,4 +1,4 @@
-// swift-tools-version:5.7
+// swift-tools-version:5.9
```
calvincestari, line 1: Can we think of any particular reason we this was still at 5.7?

calvincestari (commented):
Looks good, a couple questions:
1. Should we standardize on `swift-tools-version` of `5.9` across all packages in the repo? Eg: `apollo-ios-pagination` still on `5.8`.
2. I think we should get the `ExistentialAny` setting into the codegen test configuration projects too. Knowing those build correctly will be valuable.

apollo-ios-codegen/Package.swift
BobaFetters: I feel like sometime last year I tried to bump us forward on this and someone had an issue with it and we decided not to move forward but I could be thinking of something different. I'll see if I can find a closed PR or something for it.

BobaFetters:
> 1. Should we standardize on `swift-tools-version` of `5.9` across all packages in the repo? Eg: `apollo-ios-pagination` still on `5.8`.

I think it makes sense to standardize this across our packages.

> 2. I think we should get the `ExistentialAny` setting into the codegen test configuration projects too. Knowing those build correctly will be valuable.

+1

apollo-ios-codegen/Package.swift
BobaFetters: Looks like we had a conversation about this last year and decided to hold off for a bit because it would require Xcode 15 https://apollograph.slack.com/archives/C054R4FEW7L/p1695928304850209

I did update the `apollo-ios` package to 5.9 in January when we did the update of the InstallCLI plugin though - #132

AnthonyMDev: Yeah, I don't think we have any reason to wait any longer. Thanks for looking into it @BobaFetters

Tests/TestCodeGenConfigurations/Other-CocoaPods/CocoaPodsProject.xcodeproj/project.pbxproj
```diff
@@ -704,6 +704,7 @@
 				MTL_ENABLE_DEBUG_INFO = INCLUDE_SOURCE;
 				MTL_FAST_MATH = YES;
 				ONLY_ACTIVE_ARCH = YES;
+				OTHER_SWIFT_FLAGS = "\"-enable-upcoming-feature ExistentialAny\"";
```
calvincestari, line 707: Doesn't this one need the enclosing `"` removed too? There is one more at [L763](https://github.com/apollographql/apollo-ios-dev/pull/379/files#diff-55e3f5e7faa355914df4319b13344ca8cac090f7d15a748b24019a41dd395c82R763) that has them too.

calvincestari (approved):
Looks good, thanks for adding the test config setting. One last comment on the swift flag in one of the cocoapods test projects.

PR #380, opened 2024-06-05
JSONConverter helper

AnthonyMDev:
This is an alternative API for the changes in #377. It also cleans up the unit tests so we don't need to import `StarWarsAPI` in more places (hoping we can get rid of it completely eventually).

@TizianoCoroneo @calvincestari @BobaFetters I think this is an improvement on the API in the other PR. Let me know what you think.


apollo-ios/Sources/Apollo/JSONConverter.swift
```diff
@@ -0,0 +1,52 @@
...
+#if !COCOAPODS
+import ApolloAPI
+#endif
+
+public enum JSONConverter {
+  
+  /// Converts a ``SelectionSet`` into a basic JSON dictionary for use.
+  ///
+  /// - Returns: A `[String: Any]` JSON dictionary representing the ``SelectionSet``.
+  public static func convert(_ selectionSet: some SelectionSet) -> [String: Any] {
```
TizianoCoroneo, line 11: as you might have guessed from my PR, I'm not a fan of relying on the compiler to select the right overload, but that's my personal preference. The changes in this PR still work fine for my use case. Thanks for the help!

calvincestari (approved):
LGTM! Do we have a test for the `GraphQLError` conversion anywhere already?

AnthonyMDev:
> LGTM! Do we have a test for the `GraphQLError` conversion anywhere already?

Yes, it's at GraphQLResultTests.swift:69

PR #386, opened 2024-06-10
Fix building ApolloSQLite with Xcode 16b1

rastersize:
In Xcode 16, or rather Foundation in the iOS 18/etc. SDKs, there’s a new [`Expression` type](https://developer.apple.com/documentation/foundation/expression). The SQLite library also expose this type. This caused an issue compiling the `SQLiteDotSwiftDatabase.swift` file, as it imports both libraries. This PR disambiguates it for the compiler.

With this change the project once again builds with Xcode 16.

Screenshot 2024-06-10 at 3 59 48 PM

### DIFF
--- a/Tests/ApolloTests/RequestChainTests.swift
+++ b/Tests/ApolloTests/RequestChainTests.swift
@@ -870,6 +870,8 @@ class RequestChainTests: XCTestCase {
   }
 
   func test__memory_management__givenOperation_withEarlyAndFinalInterceptorChainExit_shouldNotHaveRetainCycle_andShouldNotCrash() throws {
+    throw XCTSkip("Flaky test skipped in PR #386- must be refactored or fixed in a separate PR.")
+
     // given
     let store = ApolloStore(cache: InMemoryNormalizedCache(records: [
       "QUERY_ROOT": [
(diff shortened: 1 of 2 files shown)


calvincestari:
Looks like it failed the `test__memory_management__givenOperation_withEarlyAndFinalInterceptorChainExit_shouldNotHaveRetainCycle_andShouldNotCrash` test which is a known flaky test.

Mordil:
Do the initializers need to be updated as well since they have explicit references to the types?

calvincestari:
> Do the initializers need to be updated as well since they have explicit references to the types?

I thought that too but then thought that the fully qualified property declaration on line 11 and 12 is enough for the compiler to know which type you actually want to use at those call sites in the initializers. I don't have Xcode 16 yet so I'm guessing at that theory.

rastersize:
@Mordil @calvincestari They don’t need to be update to build (I tested in Xcode 16b1). I can add it there as well though if you prefer?

calvincestari:
> I can add it there as well though if you prefer?

I don't think it's necessary if the compiler is not complaining about it. There's a chance that it becomes a requirement in a future version of Xcode 16 but if that happens we can add it then.

calvincestari (approved):

calvincestari:
I've confirmed this fix with Xcode 16.0 beta (16A5171c) and skipped the flaky test so we can get this PR merged.

calvincestari:
I've noticed there are some other build errors with Xcode 16 but they're limited to our test suites which I'll get fixed in a separate PR. This PR alone should be enough to get the dependency building again in your projects. Thanks for the contribution @rastersize.

PR #387, opened 2024-06-10
dev: Xcode 16 updates for test suite

calvincestari:
This is a work in progress to get our test suite building with Xcode 16. Xcode 16 is not available on GitHub Actions so this has to be a manual verification for now.


calvincestari:
Going to be included with the work to support `Sendable` instead.

PR #381, opened 2024-06-08
Schema renaming documentation

BobaFetters:
Adding documentation for the schema type renaming feature.

PR #388, opened 2024-06-13
Feature: Schema Type Renaming

BobaFetters:
This PR contains the finished schema type renaming feature that was done across multiple PRs, ready to be merged into main.

Work is from the following PRs:
- #242 
- #276 
- #381 

Closes apollographql/apollo-ios#3283


calvincestari (approved):
From the commit list it looks like this is the three PRs we've reviewed already right? That and the merges from`main`. LGTM!

PR #390, opened 2024-06-14
defer: Refactor execution for deferred fragment cache reads

calvincestari:
This changeset enables cache-based execution of operations that use `@defer`. Since there is only a single response from a cache read all available deferred fragments will be returned with the initial (partial) data. Any deferred fragments that cannot be fulfilled will remain as 'deferred' in the result.

The executor now enumerates through deferred fragments that were collected into `_deferredFragments`, matches the identifier to a fragment cache entry, and attempts to immediately resolve the fragment fields. If successful then the deferred fragment is marked as fulfilled. If the deferred fragment cannot resolved because of something such as a missing field, then the fragment remains as 'deferred' in the result.


apollo-ios/Sources/Apollo/ExecutionSources/NetworkResponseExecutionSource.swift
```diff
@@ -9,6 +9,8 @@ public struct NetworkResponseExecutionSource: GraphQLExecutionSource, CacheKeyCo
   public typealias RawObjectData = JSONObject
   public typealias FieldCollector = DefaultFieldSelectionCollector
 
+  public var shouldAttemptDeferredFragmentExecution: Bool { false }
```
calvincestari, line 12: I wonder if we should work this property into the network response parsing of incremental responses? Looking at this plainly you'd assume that network responses should not execute deferred fragments but they do because currently that's decoupled from this flag.

apollo-ios/Sources/Apollo/GraphQLExecutor.swift
```diff
@@ -265,27 +265,76 @@ public final class GraphQLExecutor<Source: GraphQLExecutionSource> {
...
         fieldEntries.append(fieldEntry)
       }
-      
-      return compactLazilyEvaluateAll(fieldEntries).map {
-        try accumulator.accept(fieldEntries: $0, info: info)
+
+      if executionSource.shouldAttemptDeferredFragmentExecution {
+        for deferredFragment in groupedFields.deferredFragments {
+          guard let fragmentType = groupedFields.cachedFragmentIdentifiers[deferredFragment] else {
+            info.deferredFragments.insert(deferredFragment)
```
calvincestari, line 305: In the case that the fragment type cannot be found in the cache I think it's best that it remains 'deferred'; it's not going to break execution.

apollo-ios/Sources/Apollo/FieldSelectionCollector.swift
```diff
@@ -4,10 +4,11 @@ import ApolloAPI
 #endif
 
 @_spi(Execution)
-public struct FieldSelectionGrouping: Sequence {
-  private var fieldInfoList: [String: FieldExecutionInfo] = [:]
+public struct FieldSelectionGrouping {
+  fileprivate(set) var fieldInfoList: [String: FieldExecutionInfo] = [:]
   fileprivate(set) var fulfilledFragments: Set<ObjectIdentifier> = []
   fileprivate(set) var deferredFragments: Set<ObjectIdentifier> = []
+  fileprivate(set) var cachedFragmentIdentifiers: [ObjectIdentifier: any SelectionSet.Type] = [:]
```
calvincestari, line 11: Might rename to `cachedFragmentIdentifierTypes`

apollo-ios/Sources/Apollo/ExecutionSources/NetworkResponseExecutionSource.swift
AnthonyMDev: I'm confused. How are network responses executing deferred fragments?

apollo-ios/Sources/Apollo/GraphQLExecutionSource.swift
```diff
@@ -17,6 +17,8 @@ public protocol GraphQLExecutionSource {
   /// GraphQL execution.
   associatedtype FieldCollector: FieldSelectionCollector<RawObjectData>
 
+  var shouldAttemptDeferredFragmentExecution: Bool { get }
```
AnthonyMDev, line 20: Can we add some documentation to this to explain what the implications this has on execution are.

AnthonyMDev (approved):
This was a lot of effort to get across the finish line. Great work Calvin! The tests looks great too!

apollo-ios/Sources/Apollo/ExecutionSources/NetworkResponseExecutionSource.swift
calvincestari: It's a few things:
1. If the operation is not a `subscription` then the `multipart/mixed` `accept` header is always added which lets the server respond with multipart responses if it can, such as with `@defer` operations - [here](https://github.com/apollographql/apollo-ios-dev/blob/feature/defer-execution-networking/apollo-ios/Sources/Apollo/RequestChainNetworkTransport.swift#L110).
2. The `IncrementalJSONResponseParsingInterceptor` is used in the request chain if the operation contains deferred fragments - [here](https://github.com/apollographql/apollo-ios-dev/blob/feature/defer-execution-networking/apollo-ios/Sources/Apollo/DefaultInterceptorProvider.swift#L48).
3. Then the partial vs. incremental JSON parsing is determined by whether we have parsed an initial (partial) response yet.

calvincestari: So considering those steps above, they are really for multipart parsing which _should_ be decoupled from defer specifically. The thing that determines `@defer` parsing is the specification we send in the `accept` header and sent back to us in the response which we then use to determine which parser to use [here](https://github.com/apollographql/apollo-ios-dev/blob/feature/defer-execution-networking/apollo-ios/Sources/Apollo/MultipartResponseParsingInterceptor.swift#L67).

It seems odd to have a 'master' flag `shouldAttemptDeferredFragmentExecution` for defer networking parsing; I'm not sure where we'd put the check?

calvincestari: That makes me think that the [check for defer operations](https://github.com/apollographql/apollo-ios-dev/blob/feature/defer-execution-networking/apollo-ios/Sources/Apollo/DefaultInterceptorProvider.swift#L48) to decide whether to use `IncrementalJSONResponseParsingInterceptor` vs. `JSONResponseParsingInterceptor` is also incorrect since subscriptions need incremental parsing too but won't be considered a 'defer' operation.

calvincestari: OK - I'm going to keep those out of this PR and follow up with a defer-multipart-cleanup PR where I'll tackle these issues.

calvincestari: For anyone else reading this - we've cleared this up in the documentation of the new property.

PR #392, opened 2024-06-18
cI: Release Automation Step 1

BobaFetters:
This is the first step in automating our release process as much as possible. The plan is to have this manually triggered CI job that can create the initial PR and handle all the basic tasks associated with that. Then a separate CI job that will be triggered on merge of release PRs to handle the tasks associated with that (tagging, creating release on github repo etc)

Currently planning to leave CHANGELOG handling as a manual process and get as many other parts of the CI process as possible automated and then revisit the possibility of automating CHANGELOG handling as well.

Sample Release PR created by this workflow: https://github.com/apollographql/apollo-ios-dev/pull/394

The "Things to do as part of releasing" section will be the focus of the workflow that triggers from merging the release PR.

### DIFF
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -10,6 +10,7 @@ on:
 env:
   BRANCH_NAME: "release/${{ github.event.inputs.versionNumber }}"
   VERSION_NUMBER: ${{ github.event.inputs.versionNumber }}
+  CURRENT_SHA: ${{ github.sha }}
 
 jobs:
   create-release-pr:
@@ -48,7 +49,7 @@ jobs:
           Release ${{ env.VERSION_NUMBER }}
         body: |
           #### Diff
-          [See diff since last version](https://github.com/apollographql/apollo-ios-dev/compare/${{ env.PREVIOUS_VERSION }}...${{ github.sha }}).
+          [See diff since last version](https://github.com/apollographql/apollo-ios-dev/compare/${{ env.PREVIOUS_VERSION }}...${{ env.CURRENT_SHA }}).
 
           #### Things to do in this PR
           - [ ] Update [`CHANGELOG.md`](https://github.com/apollographql/apollo-ios/blob/main/CHANGELOG.md) with all relevant changes since the prior version. _Please include PR numbers and mention contributors for external PR submissions._
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -9,6 +9,7 @@ on:
 
 env:
   BRANCH_NAME: "release/${{ github.event.inputs.versionNumber }}"
+  VERSION_NUMBER: ${{ github.event.inputs.versionNumber }}
 
 jobs:
   create-release-pr:
@@ -32,7 +33,7 @@ jobs:
       run: |
         previousVersion=$(sh apollo-ios/scripts/get-version.sh)
         echo "PREVIOUS_VERSION=$previousVersion" >> ${GITHUB_ENV}
-        sh ./scripts/set-version.sh ${{ github.event.inputs.versionNumber }}
+        sh ./scripts/set-version.sh ${{ env.VERSION_NUMBER }}
         (cd SwiftScripts && swift run DocumentationGenerator)
         make archive-cli-to-apollo-package
     - name: Create Pull Request
@@ -40,11 +41,11 @@ jobs:
       with:
         branch: ${{ env.BRANCH_NAME }}
         commit-message: |
-          Setting up Release ${{ github.event.inputs.versionNumber }}
+          Setting up Release ${{ env.VERSION_NUMBER }}
 
           Created By GitHub Action Workflow
         title: |
-          Release ${{ github.event.inputs.versionNumber }}
+          Release ${{ env.VERSION_NUMBER }}
         body: |
           #### Diff
           [See diff since last version](https://github.com/apollographql/apollo-ios-dev/compare/${{ env.PREVIOUS_VERSION }}...${{ github.sha }}).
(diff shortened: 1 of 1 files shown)


apollo-bot2:
## :red_circle: Potential Security Issues Found

If you are seeing this message, it means that [the security scanning tool](https://semgrep.dev/) that Apollo uses to test our source has identified a potential security issue in code introduced or updated via your branch. Information about what was detected and steps to move forward are below.

If the scanner detected a legitimate issue, please take action to correct it prior to merging this PR. The action required will vary based on the detection. If the detection is a false positive, please follow the steps below to resolve the issue.

###  Issue Detected 
| Message | Caused CI Failure? | Link | Path | Signature |
|---------|--------------------|------|------|-----------|
| Using variable interpolation `${{...}}` with `github` context data in a `run:` step could allow an attacker to inject their own code into the runner. This would allow them to steal secrets and code. `github` context data can have arbitrary user input and should be treated as untrusted. Instead, use an intermediate environment variable with `env:` to store the data and use the environment variable in the `run:` script. Be sure to use double-quotes the environment variable, like this: "$ENVVAR". | Yes | <a href="https://github.com/apollographql/apollo-ios-dev/blob/4ac2041148ec638984c5c3043e3d93004c134375/.github/workflows/create-release-pr.yml?plain=1#L32" target="_blank">Link</a> | .github/workflows/create-release-  pr.yml | rules.providers.semgrep.security.ya  ml.github-actions.security.run-  shell-injection |
<hr />
<details>
    <summary><h3>False Positive Resolution Process</h3></summary>

The easiest way to resolve a false positive is to add a comment containing `nosemgrep` to the code that triggered the detection. You can add the comment directly to the line triggering the detection or on the line immediately above it. For example:

```python
#!/usr/bin/env python3

def my_function():
    my_code_generating_detection() # nosemgrep
```

<details>
    <summary><h4>What if I can't add a line comment?</h4></summary>

If you can't add a comment because the detection is generated by a file that does not support comments (like JSON), you can use Semgrep's `.semgrepignore` file. For more information about how to create an exclusion via `.semgrepignore`, refer to the [Semgrep documentation](https://semgrep.dev/docs/ignoring-files-folders-code/#defining-ignored-files-and-folders-in-semgrepignore).
</details>

</details>
<hr />
<details> 
    <summary><h3>Getting Help</h3></summary>


The Apollo Security team is available to assist in resolving this issue. Please tag us on this PR using  `@apollographql/security` if you need assistance! 
</details>
<hr />
<details>
    <summary><h3>How do I know I fixed this correctly?</h3></summary>
Once you have resolved the issue completely, this message will disappear! If you're still seeing this message, there is more to do prior to merging.
</details>

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,61 @@
...
+      with:
+        ref: main
+    - name: Setup SSH Keys
+      uses: webfactory/ssh-agent@v0.7.0
+      with:
+        ssh-private-key: |
+          ${{ secrets.APOLLO_IOS_DEV_DEPLOY_KEY }}
+    - name: Configure Git
+      uses: ./.github/actions/configure-git
+    - name: Run Commands
```
AnthonyMDev, line 31: Can we make the name here describe what commands are being run. Or break this into different steps for each command?

AnthonyMDev (approved):
This is awesome! Thanks so much for getting this started!

.github/workflows/create-release-pr.yml
BobaFetters: Sure I can break them out to separate commands

calvincestari:
> Currently planning to leave CHANGELOG handling as a manual process and get as many other parts of the CI process as possible automated and then revisit the possibility of automating CHANGELOG handling as well.

I think the changelog can definitely be automated but we'd need to all adopt PR titles like [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification) specifies. The changelog will be less descriptive than we have now but collecting those explanations is what makes it a manual process at the moment.

calvincestari:
> but we'd need to all adopt PR titles like [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification) specifies.

We could enforce this through our PR validation action so the community has to abide too.

.github/workflows/create-release-pr.yml
```diff
@@ -0,0 +1,71 @@
...
+          #### Diff
+          [See diff since last version](https://github.com/apollographql/apollo-ios-dev/compare/${{ env.PREVIOUS_VERSION }}...${{ github.sha }}).
+
+          #### Things to do in this PR
+          - [ ] Update [`CHANGELOG.md`](https://github.com/apollographql/apollo-ios/blob/main/CHANGELOG.md) with all relevant changes since the prior version. _Please include PR numbers and mention contributors for external PR submissions._
+
+          #### Things to do as part of releasing
+          - [ ] Add tag of format `major.minor.patch` to GitHub.
+          - [ ] Create a release on GitHub with the new tag, using the latest [`CHANGELOG.md`](https://github.com/apollographql/apollo-ios/blob/main/CHANGELOG.md) contents.
+          - [ ] Attach CLI binary to the GitHub release. _Use the `make archive-cli-for-release` command which builds both Intel and ARM architectures, and creates the tar archive for you._
```
calvincestari, line 69: We can probably remove the instruction "_Use the `make archive-cli-for-release` command which builds both Intel and ARM architectures, and creates the tar archive for you._" since that should already have been done in the 'Archive CLI' step. Copying the output binary generated there is enough.

calvincestari (approved):
LGTM!

BobaFetters:
> > but we'd need to all adopt PR titles like [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/#specification) specifies.
> 
> We could enforce this through our PR validation action so the community has to abide too.

Yea I assumed we would do something along these lines, figured some investigation and discussion to decide exactly what we want to would be good.

.github/workflows/create-release-pr.yml
BobaFetters: Good catch!

PR #395, opened 2024-06-18
Fixing broken integration tests

BobaFetters:
Fixing broken integration tests from schema type renaming work


calvincestari (approved):
Thank you!

PR #396, opened 2024-06-21
ci: Setting up release publishing

BobaFetters:
This PR sets up the publishing portion of automating our release process. Some of the steps may (likely will) require some tweaking once we have had a chance to test it in practice to ensure everything is configured exactly correct, using proper tokens etc.


AnthonyMDev (approved):
Unsure if cocoapods setup is sufficient, but everything else looks good to me. I'd say we try it and see what happens.

BobaFetters:
> Unsure if cocoapods setup is sufficient, but everything else looks good to me. I'd say we try it and see what happens.

Yea honestly most of it is iffy if it will work as is or need some extra setup etc but hard to know without giving it a test

PR #401, opened 2024-06-25
Adjusting release workflow

BobaFetters:
Making some minor adjustments to the release workflow based on things noticed testing the workflow with the 1.13.0 release


.github/workflows/pr-subtree-push.yml
```diff
@@ -83,15 +84,11 @@ jobs:
...
         uses: ./.github/actions/configure-git
       - name: Trigger publish-release workflow
         shell: bash
         run: |
-          gh auth login -p ssh
-          gh workflow run publish-release.yml
\ No newline at end of file
+          gh workflow run publish-release.yml
+        env:
+          GH_TOKEN: ${{ secrets.APOLLO_IOS_PAT }}
```
calvincestari, line 94: Is this PAT for a general CI user or tied to one of us? If it's a personal one we should ask the security team what their preference is. _I might have asked this before, I'm not sure._

calvincestari (approved):

calvincestari:
> Wondering if instead of using the `peter-evans/create-pull-request@v6` action for pull requests if we should just manually handle them with the GitHub CLI like we are doing for some other things?

If it's a simple limited number of steps then I always prefer to have us do them rather than the 'magic' of an action; one less dependency too.

.github/workflows/pr-subtree-push.yml
BobaFetters: It was created by me, so yea if we need a general use one we can switch this easy enough

BobaFetters:
> > Wondering if instead of using the `peter-evans/create-pull-request@v6` action for pull requests if we should just manually handle them with the GitHub CLI like we are doing for some other things?
> 
> If it's a simple limited number of steps then I always prefer to have us do them rather than the 'magic' of an action; one less dependency too.

I'll double check the documentation for the cli but it honestly should be pretty simple I think

calvincestari (approved):

PR #403, opened 2024-06-26
refactor: Validate or add `@defer` directive in schema

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3287.

The change here is to:
1. Validate any `@defer` directive that is in the schema document. The validation is to ensure that any schema definition matches the experimental definition of the directive in graphql-js. We do this because some server implementations allow `@defer` to be used in other locations, such as `FIELD`, and Apollo iOS codegen does not support those definitions. If an unsupported `@defer` directive definition is found a warning will be output and the definition will be automatically replaced with a version that we do support. This could result in operation validation failing if the directive was being used in a way that does not match the replaced definition.
2. Automatically add a `@defer` directive if it is not found in the schema. This mimics the eventual state of the defer proposal where the directive is available to be used without explicit definition.


apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/__tests__/deferDirectiveTests.ts
```diff
@@ -16,11 +16,151 @@ import {
...
+  });
+})
+
+describe("test invalid directive definitions", () => {
+
+  // Directive Definition Tests
+
+  describe("given mismatched location in directive", () => {
+    const schemaSDL: string = `
+    directive @defer(label: String, if: Boolean! = true) on FIELD
```
AnthonyMDev, line 108: If the defer directive is _only_ on `FIELD` than we should throw an error, but if it's on _both_ fields and `FRAGMENT_SPREAD | INLINE_FRAGMENT` then I think we should allow it here. We should only fail if they actually try to use a `@defer` directive on a field.

apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/__tests__/deferDirectiveTests.ts
```diff
@@ -16,11 +16,151 @@ import {
...
+      species: String!
+      friend: Animal!
+    }
+    `;
+
+    it("should throw when loading documents", () => {
+      expect(() => {
+        loadSchemaFromSources([new Source(schemaSDL, "Test Schema", { line: 1, column: 1 })]);
+      }).toThrow(
+        Error("Mismatched @defer directive found.")
```
AnthonyMDev, line 129: Can we make the error here more clear to the user of what went wrong. "Mismatched" isn't very helpful for them.

AnthonyMDev (changes requested):

apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/__tests__/deferDirectiveTests.ts
calvincestari: OK, this now sends a warning to the console that the supplied directive definition is unsupported and will be replaced.

apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/__tests__/deferDirectiveTests.ts
calvincestari: As we spoke about earlier in the week, the behaviour here now is to print a warning and automatically replace the unsupported definition with the experimental graphql-js defined directive that we do support.

calvincestari:
@AnthonyMDev this is ready for another review, should be all done.

AnthonyMDev (approved):

PR #412, opened 2024-07-02
Adding debugDescription to SelectionSets

BobaFetters:
- Adding `debugDescription` through `CustomDebugStringConvertible` to `SelectionSet`

Closes apollographql/apollo-ios#3374

### DIFF
--- a/apollo-ios/Sources/ApolloAPI/SelectionSet.swift
+++ b/apollo-ios/Sources/ApolloAPI/SelectionSet.swift
@@ -46,7 +46,7 @@ public protocol CompositeInlineFragment: CompositeSelectionSet, InlineFragment {
 }
 
 // MARK: - SelectionSet
-public protocol SelectionSet: Hashable {
+public protocol SelectionSet: Hashable, CustomDebugStringConvertible {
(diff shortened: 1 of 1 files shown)


apollo-ios/Sources/ApolloAPI/SelectionSet.swift
```diff
@@ -117,6 +117,10 @@ extension SelectionSet {
   @inlinable public static func ==(lhs: Self, rhs: Self) -> Bool {
     return lhs.__data == rhs.__data
   }
+  
+  public var debugDescription: String {
+    return "\(self.__data._data as AnyObject)"
```
BobaFetters, line 122: Should this print more than the data inside the `DataDict`, like the `fulfilledFragments` etc? And are we good with a readable format like provided by casting to `AnyObject` like the screenshot or do we want to manually create an output from keys/values?

Screenshot 2024-07-02 at 12 37 25 AM

calvincestari (approved):
I think this satisfies what was requested in the issue. The other properties of a `DataDict` probably don't need to be output, we can adjust later if needed.

PR #413, opened 2024-07-02
feature: GraphQL execution for `@defer` support

calvincestari:
This PR is to merge the `@defer` required changes to the GraphQL executer into `main`. There is already some of the `@defer` work in the `main` branch but we chose to build the executor changes on a separate branch in isolation not to cause any instability to `main`.

This PR should only consist of other PRs that have already been reviewed and merges from `main` to keep this branch up-to-date with `main`.


john-twigg-ck:
@calvincestari 

I'd like to provide feedback about the the way I would expect to be able to leverage the `@Deferred` property.

The ability to introspect the status of `pending vs final` is useful, however as an end user I would like to be able to `await` on a pending state and eventually receive the final result. 

pseudo example taken from [here](https://github.com/apollographql/apollo-ios-dev/pull/413/files#diff-9d4127f7f6d22e03fa35bd8a04a24199fb71a39dab7bcb5862802c33cacd2778R184)
```
      expect(animal.fragments.$deferredGenus).to(equal(.pending))
        
      let genus = try await animal.fragments.$deferredGenus 
        
      expect(animal.fragments.$deferredGenus).to(equal(.fulfilled)) //its now finished
      expect(animal.fragments.deferredGenus).toBeNonNil() //its non nil
      expect(animal.fragments.deferredGenus).to(equal(genus)) 

```

NOTE: I'd expect something similar with `@stream` where I could receive an `AsyncStream `

calvincestari:
Hi @john-twigg-ck, thanks for the feedback.

> The ability to introspect the status of pending vs fulfilled is useful, however as an end user I would like to be able to await on a pending state and eventually receive the final result.

The implementation of `@defer` is built on the existing network stack which unfortunately does not have any `async`/`await` support. So awaiting on the status won't be possible in 1.x. @AnthonyMDev is currently working on [Swift 6 compatibility](https://github.com/apollographql/apollo-ios-dev/blob/main/apollo-ios/ROADMAP.md#20-release---swift-6-compatibility) which will be a major version bump, and will include proper support for `async`/`await` too so it's possible that in the future we could get that working but it's still got some challenges though due to the way that the models and data is handled, i.e.: models and their data are not persisted so there is nothing to 'update'.

> NOTE: I'd expect something similar with `@stream` where I could receive an `AsyncStream`

There's only bad news here I'm afraid as unfortunately there is no existing support for `@stream`, and it's not in this implementation either. I don't think any of the Apollo clients support `@stream` nor does the Apollo Router. Eventually when the `@defer`/`@stream` proposal is merged into the GraphQL spec we'll have to provide support for it.

PR #414, opened 2024-07-03
defer: Documentation update

calvincestari:
This adds more detail to the `@defer` feature usage including a mention of cache behaviour.

### DIFF
--- a/docs/source/fetching/defer.mdx
+++ b/docs/source/fetching/defer.mdx
@@ -2,13 +2,13 @@
 title: Using the @defer directive in Apollo iOS
 ---
 
-> ⚠️ **The `@defer` directive is currently [experimental](https://www.apollographql.com/docs/resources/release-stages/#experimental-features) in Apollo iOS and is available for use in the `TBD` release.** If you have feedback on the feature, please let us know via [GitHub issues](https://github.com/apollographql/apollo-ios/issues/new?assignees=&labels=bug%2Cneeds+investigation&projects=&template=bug_report.yaml).
+> ⚠️ **The `@defer` directive is currently an [experimental feature](https://www.apollographql.com/docs/resources/release-stages/#experimental-features) in Apollo iOS and is available for use from version x.y.z.** If you have feedback on the feature, please let us know via a new [GitHub issue](https://github.com/apollographql/apollo-ios/issues/new?assignees=&labels=bug%2Cneeds+investigation&projects=&template=bug_report.yaml) or search through any [existing `@defer`-related issues](https://github.com/apollographql/apollo-ios/issues?q=is%3Aissue+is%3Aopen+%22defer%22).
 
-Beginning with version `TBD`, Apollo iOS provides experimental support of [the `@defer` directive](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md), which enables your queries to receive data for specific fields asynchronously. This is helpful whenever some fields in a query take much longer to resolve than the others.
+The [`@defer` directive](https://github.com/graphql/graphql-wg/blob/main/rfcs/DeferStream.md) enables your queries to receive data for specific fields asynchronously. This is helpful whenever some fields in a query take much longer to resolve than the others.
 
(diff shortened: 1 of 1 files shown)


docs/source/fetching/defer.mdx
```diff
@@ -2,13 +2,13 @@
 title: Using the @defer directive in Apollo iOS
 ---
 
-> ⚠️ **The `@defer` directive is currently [experimental](https://www.apollographql.com/docs/resources/release-stages/#experimental-features) in Apollo iOS and is available for use in the `TBD` release.** If you have feedback on the feature, please let us know via [GitHub issues](https://github.com/apollographql/apollo-ios/issues/new?assignees=&labels=bug%2Cneeds+investigation&projects=&template=bug_report.yaml).
+> ⚠️ **The `@defer` directive is currently an [experimental feature](https://www.apollographql.com/docs/resources/release-stages/#experimental-features) in Apollo iOS and is available for use from version x.y.z.** If you have feedback on the feature, please let us know via a new [GitHub issue](https://github.com/apollographql/apollo-ios/issues/new?assignees=&labels=bug%2Cneeds+investigation&projects=&template=bug_report.yaml) or search through any [existing `@defer`-related issues](https://github.com/apollographql/apollo-ios/issues?q=is%3Aissue+is%3Aopen+%22defer%22).
```
AnthonyMDev, line 5: Let's just make sure not to merge this PR in until we are ready to release a version with defer and change that from "version x.y.z" to the appropriate version number.

docs/source/fetching/defer.mdx
```diff
@@ -30,18 +30,18 @@ query PersonQuery($personId: ID!) {
 }
 ```
 
-In the generated code for this query, the `asUser` field for the fragment will be optional. That is because when the initial payload is received from the server, the fields of the fragment are not yet present. A `Person` will be returned with only the basic fields filled in.
+In the generated code for this query the `asUser` type case will be optional, it will also have a `fragments` container in which there will be an optional fragment with the same name as the `label` property value used in the query. The deferred fragment is optional because it will not be returned with the basic fields, it will be returned in a separate response instead.
```
AnthonyMDev, line 33: Will there `asUser` even be generated? I'm not sure if the defer work changes this, but in the past, if the `person` selection set was already on `User`, a type case on the exact matching type would be reduced out by the codegen engine. I think the preferred behavior here would be to just have `data.person.fragments.deferredFriends`.

If this only changed for deferred type cases, I'm okay with it, but if this changes how the codegen engine handles all type cases, then we're going to have existing code break when regenerated.

AnthonyMDev: If person was on some `PersonInterface` type and the typecase was on `User`, then this would be correct though.

AnthonyMDev (approved):

docs/source/fetching/defer.mdx
calvincestari: Currently this is to be merged into the `feature/defer-execution-networking` branch, which will then get merged into `main`. Two options then:
1. Redo this change against main and leave it till we know which version it will be released in.
2. Keep with the current merge path and update the doc before release.

docs/source/fetching/defer.mdx
calvincestari: There isn't any generated code for this example that I was working from, I simply copied the [example from Apollo Kotlin](https://www.apollographql.com/docs/kotlin/fetching/defer/) so the docs are similar.

> If person was on some PersonInterface type and the typecase was on User, then this would be correct though.

That's my understanding of how this generated model would work. `person` is defined as an interface and `friends` only exists on `User` which implements `PersonInterface`. I agree that if `person` was of type `User` then `asUser` would not be generated; defer hasn't changed that.

docs/source/fetching/defer.mdx
calvincestari: I think let's stick with option 2, which is the current merge path. At the point that we merge `feature/defer-execution-networking` into `main` it's ready for release so updating the docs can happen in the release branch.

PR #417, opened 2024-07-05
fix: Local cache mutation build error in Swift 6

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3398. Gets Xcode 16 successfully building local cache mutations.

* I've opted to simply remove the fragment mutation setter and associated `@available` attribute. There is no functional change to the code because the setter was unable to be called anyways. So the only real loss is the convenience message displayed to the user if they did try to mutate the entire fragment instead of properties on the fragment. We could state this explicitly in the local cache mutation documentation if we think that would be helpful instead.
* ~I also removed the "CI Tests (Xcode 16.0 Beta)" action that I added for us to manually run our CI suite on Xcode 16.0 beta. This was not as useful as I initially thought because it's catching all the concurrency (`Sendable`) related issues too which we know are a much more complex problem to resolve.~ This is back but with a reduced test suite that ensures the project can build and run unit tests (not codegen test configuration tests though) with Xcode 16. 

I suggest we continue to monitor the Xcode 16 beta releases and if this turns out to be a beta glitch then we can revert this PR to get back the setter with attribute and users will get the nice warning message again if they try mutate the entire fragment.


BobaFetters (approved):
This seems like a good interim solution to me to get past this issue for Xcode 16 beta builds, agree we should continue to monitor the beta builds to see if this is just a bug in the current beta and can maybe be reverted in the future.

calvincestari:
Thanks @BobaFetters. Let's get @AnthonyMDev's opinion too before merging.

AnthonyMDev:
If this compiles and builds now, I'm good with merging it! When I initially did this work, I don't believe it was possible to include a `_modify` accessor without also including a `set` accessor (though maybe I just assumed that and didn't try?).

calvincestari:
> If this compiles and builds now, I'm good with merging it! When I initially did this work, I don't believe it was possible to include a `_modify` accessor without also including a `set` accessor (though maybe I just assumed that and didn't try?).

Yes, this builds with both Xcode 15.4 and Xcode 16 without the setter.

PR #419, opened 2024-07-09
Add editor config files

BobaFetters:
Adding editor config files for apollo dev and ios/codegen/pagination packages

Closes apollographql/apollo-ios#3404

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,8 @@
+# editorconfig.org
+root = true
+
+[*]
+indent_style = space
+indent_size = 2
+trim_trailing_whitespace = true
+insert_final_newline = true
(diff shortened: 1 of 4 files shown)


BobaFetters:
@AnthonyMDev @calvincestari Thoughts on having a single `.editorconfig` at the `apollo-ios-dev` level for everything vs also having one for each package (ios, codegen, pagination) since currently we do all development through dev?

calvincestari:
> @AnthonyMDev @calvincestari Thoughts on having a single `.editorconfig` at the `apollo-ios-dev` level for everything vs also having one for each package (ios, codegen, pagination) since currently we do all development through dev?

I think one at the Tuist project level should be sufficient; that's where we want everyone to develop from and all sub-projects should have the same formatting.

calvincestari (approved):

PR #428, opened 2024-07-15
[ApolloPagination] Update `PaginationOutput` to operate over `GraphQLResult`s

Iron-Ham:
Closes: https://github.com/apollographql/apollo-ios/issues/3413

> [!CAUTION]
> This is a breaking change.

The goal of this PR is two-fold: 
1. We want to expose errors per-response when fetching data, and handle those errors gracefully where we can. While we do currently expose networking failures to consumers, we are not correctly parsing GraphQL error responses. Additionally, since we were not considering errors within the responses previously, we weren't handling them correctly. This was especially noticeable in `loadAll`, where some cases may lead to continued retries of fetching data. 
2. We want to simplify usage. 

#### Changing the stored values of `PaginationOutput`

We started by changing the stored values of `PaginationOutput` to be `GraphQLResult<GraphQLQuery.Data`, instead of the `GraphQLQuery.Data`: See [this diff](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-9a5fa952a80779b87487bc082ff38c8cdb31ea4cf01d6c68f09418aa75dc8fe0R9-R20) for detail. 

#### Changing the `Output` types of `GraphQLQueryPager`

With the changes to `PaginationOutput` in place, with each page containing a `Source` (i.e., `cache` or `server`) via its `GraphQLResult`, we made the choice to modify the `Output` values of both `GraphQLQueryPager` and `AsyncGraphQLQueryPager` such that those values are no longer tuples of `(Model, UpdateSource)` and are instead just `Model`: [1](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-15b3e5571b28a433475df3ef139318cd9aa539ef1302cd4e4600639fbb8d57aaL9) [2](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1R9). 

#### Improving `loadAll`

To correctly handle errors within `loadAll`, we first detect whether an error occurred and use that error to set `isLoadingAll` to `false` ([1](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-734cf04edf27d84c994976e044fdb587c2f0e1dda821705b55a6162a36c97937R370)), which [forces an exit](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-734cf04edf27d84c994976e044fdb587c2f0e1dda821705b55a6162a36c97937R139) of `loadAll`. 

#### Storing the most recent result

This is important in order to know the most recent `Source` of an update, as well as the `errors` key of the most recent update. [1](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-9a5fa952a80779b87487bc082ff38c8cdb31ea4cf01d6c68f09418aa75dc8fe0R17)

#### Changes to `subscribe`

We marked the `subscribe` methods of both `GraphQLQueryPager` and `AsyncGraphQLQueryPager` as deprecated: [1](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1R105) [2](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-15b3e5571b28a433475df3ef139318cd9aa539ef1302cd4e4600639fbb8d57aaR103). This is because the subscribe method was already redundant: The `GraphQLQueryPager` and `AsyncGraphQLQueryPager` are Combine `Publisher`s, and a consumer of the API can subscribe to them by calling `sink`. 

#### Removal of Initializers

In conjunction with the changes related to errors, we've removed various initializers that were no longer valid given the change in return types. We considered removal of all convenience initializers and initializers that contain a `transform` function, but found them useful in specific applications. For example, some consumers of the library must have an output type that is not constrained to any underlying `Query`, as the underlying `Query` may be different. For those consumers, the `transform:` initializers remain. For all others, however, it's highly recommended that the standard `transform`-less initializer is used, as it retains information about the most recent value, errors within, and so on. 

#### New computed properties

Some users have made note that they commonly employ a bit of code like this one:

```swift
let allPages = previousPages + [initialPage] + nextPages
```

We've added a [computed property](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-9a5fa952a80779b87487bc082ff38c8cdb31ea4cf01d6c68f09418aa75dc8fe0R79-R80) that should automatically do that for you if the pages are of the same type, as well as a [computed property](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-9a5fa952a80779b87487bc082ff38c8cdb31ea4cf01d6c68f09418aa75dc8fe0R75-R77) that surfaces all data.

Similarly, we've added a [computed property](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-9a5fa952a80779b87487bc082ff38c8cdb31ea4cf01d6c68f09418aa75dc8fe0R27-R30) that surfaces all errors from all pages. 

### Other Changes

We've removed conformance to `Equatable` on `GraphQLQueryPager`, as it doesn't make much sense – we were previously comparing the equatability of the current values – but not of the underlying data or state of the pager. [1](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-15b3e5571b28a433475df3ef139318cd9aa539ef1302cd4e4600639fbb8d57aaL266-L282) [2](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-bbd20683d6ef1206d7aae2b688f4b2124f812c95aae7af409761c48c322399a1L256-L272) 

We've renamed the `fetch` case of `UpdateSource` to `server`, to be more in line with `GraphQLResult.Source`. [1](https://github.com/apollographql/apollo-ios-dev/pull/428/files#diff-1b96ca532f31d2f77a8a78b720248a4cb0e29f641c22f887b0d3d9756648ed56L2). 

## Impact

The impact of this PR is:
- Users are now aware of request-level errors when using `GraphQLQueryPager`
- `loadAll` can no longer short-circuit.
- The caller API is simplifying and maturing, with fewer ways to achieve the same results. This also means a net deletion of code within the library, and less surface area to support.


apollo-ios-pagination/Sources/ApolloPagination/AsyncGraphQLQueryPager.swift
```diff
@@ -248,25 +178,12 @@ public class AsyncGraphQLQueryPager<Model>: Publisher {
...
-      return true
-    default:
-      return false
-    }
+  ) where S: Subscriber, Never == S.Failure, Result<Model, any Error> == S.Input {
+    publisher
+      .removeDuplicates(by: { _lhs, _rhs in
+        guard let lhs = try? _lhs.get(), let rhs = try? _rhs.get() else { return false }
+        return lhs == rhs
+      })
```
AnthonyMDev, line 186: Is this the only reason we added the requirement to make models `Equatable`? I'd prefer if we didn't add restrictions that aren't 100% necessary.

Further, I'm not convinced we should even be calling `removeDuplicates` here. Is there a situation caused by this PR where we are now anticipating duplicated updates that we didn't not before?

AnthonyMDev (commented):
I know you’re still working on getting this PR passing CI. But I’m doing some code review now and I’m having a hard time understanding some of the changes made here. Adding this PaginationResult feels like it adds a lot of complexity, both internally and in the consumption of the API in calling code.

---

This also seems to be concatenating all errors from individual pages into one `[errors]` array on the `PaginationResult`. That is not ideal, as it will be difficult to understand which errors came from which requests and trace them back using the error `path` and `location`.

---

As is often the case, this is becoming a lot more involved than was originally anticipated to solve the issues in https://github.com/apollographql/apollo-ios/issues/3413. I'm wondering if there is a different approach to addressing that without making such sweeping changes. 

I'm not sure if this is the right approach either, but would it be possible to keep the `transform` APIs, but change them to expose the errors for each query? That could allow users to determine if they want to ignore and suppress those errors or expose them however they see fit. This could mean we are just pushing off the difficulty of dealing with per-page errors to the users, rather than handling them ourselves though...

Iron-Ham:
cc: @AnthonyMDev I'll push changes with the swap to using `GraphQLResult` within PaginationOutput following this PR

Iron-Ham:
cc: @AnthonyMDev would it be helpful to split this into a few smaller PRs?

AnthonyMDev (approved):

PR #431, opened 2024-07-16
Disable Fragment Field Merging

AnthonyMDev:
This PR is a working implementation of the feature to allow disabling of field merging in generated models. Previous PRs were corrupted during rebasing and will be closed. This unified PR contains the entire feature.

**Field merging configuration is implemented as an experimental feature currently. This needs ample testing in production before being considered stable.**

## Functionality

- Adds FieldMerging option to ApolloCodegenConfiguration
- Implements logic for respecting merging strategy while traversing EntitySelectionTree 
- Updates SelectionSetTemplate to properly generate based on field merging configuration
- Validates that `selectionSetInitializers` and `fieldMerging` are not used together.

## Testing:

This PR is fully unit tested.
Manual testing can be done on the `AnimalKingdomAPI`, `GitHubAPI`, and `StarWarsAPI` by changing the config values in the `SwiftScripts/Codegen` and then running `swift run Codegen`. All packages compile correctly with any combination of field merging values.

## Future Work

Possible future iterations include:
- Allowing field merging to be configured on a per definition or per selection set basis.
- This feature does not work with `selectionSetInitializers`. I currently believe these to be fundamentally incompatible. If a means of making these features compatible can be identified, we should implement it.


calvincestari (approved):
Tests look good, let's get this merged!

PR #435, opened 2024-07-19
chore: Create draft release for `apollo-ios-codegen` when publishing a release

calvincestari:
When doing the release for 1.14.0 I noticed that there was no draft release created in the `apollo-ios-codegen` repo and I had to manually create releases for 1.13.0 and 1.14.0. This should now do that automatically for us.

### DIFF
--- a/.github/workflows/create-release-pr.yml
+++ b/.github/workflows/create-release-pr.yml
@@ -67,4 +67,6 @@ jobs:
           - [ ] Update [CHANGELOG.md](https://github.com/apollographql/apollo-ios/blob/main/CHANGELOG.md) with all relevant changes since the prior version. _Please include PR numbers and mention contributors for external PR submissions._
 
           #### Things to do as part of releasing
-          - [ ] Update and publish the draft release created by CI using the latest [CHANGELOG.md](https://github.com/apollographql/apollo-ios/blob/main/CHANGELOG.md) contents
+          - [ ] Run the []'Publish Release' workflow](https://github.com/apollographql/apollo-ios-dev/actions/workflows/publish-release.yml) in apollo-ios-dev
+          - [ ] Update and publish the [draft release in apollo-ios](https://github.com/apollographql/apollo-ios/releases) created by CI using the latest [CHANGELOG.md](https://github.com/apollographql/apollo-ios/blob/main/CHANGELOG.md) contents
+          - [ ] Update and publish the [draft release in apollo-ios-codegen](https://github.com/apollographql/apollo-ios-codegen/releases) created by CI (copy and paste the previous release contents and update the version number and link)
--- a/.github/workflows/publish-release.yml
+++ b/.github/workflows/publish-release.yml
@@ -62,10 +62,17 @@ jobs:
           git push --tags
 
       # Create Draft Release on GitHub with tag of version
-      - name: Create Draft Release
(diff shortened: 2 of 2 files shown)


.github/workflows/publish-release.yml
```diff
@@ -62,12 +62,19 @@ jobs:
...
           gh release create ${{ env.RELEASE_VERSION }} 'apollo-ios/CLI/apollo-ios-cli.tar.gz#apollo-ios-cli.tar.gz' -d --repo "apollographql/apollo-ios" -t "${{ env.RELEASE_VERSION }}"
         env:
           GH_TOKEN: ${{ secrets.APOLLO_IOS_PAT }}
+      
+      - name: Create Draft Release (`apollo-ios-codegen`)
+        shell: bash
+        run: |
+          gh release create ${{ env.RELEASE_VERSION }} -d --repo "apollographql/apollo-ios-codegen" -t "${{ env.RELEASE_VERSION }}"
+        env:
+          GH_TOKEN: ${{ secrets.APOLLO_IOS_PAT }}
```
calvincestari, line 77: @BobaFetters - do you know if this PAT will allow us to create a release in the `apollo-ios-codegen` repo too? I noticed that we're using it in the step below for triggering the `apollo-ios-xcframework` release workflow so I'm hopeful the answer is yes?

BobaFetters: sorry not sure how I didn't see this PR sooner, yea the PAT works for all the apollo ios repos so we can use the same one

.github/workflows/create-release-pr.yml
```diff
@@ -67,4 +67,6 @@ jobs:
           - [ ] Update [CHANGELOG.md](https://github.com/apollographql/apollo-ios/blob/main/CHANGELOG.md) with all relevant changes since the prior version. _Please include PR numbers and mention contributors for external PR submissions._
 
           #### Things to do as part of releasing
-          - [ ] Update and publish the draft release created by CI using the latest [CHANGELOG.md](https://github.com/apollographql/apollo-ios/blob/main/CHANGELOG.md) contents
\ No newline at end of file
+          - [ ] Run the []'Publish Release' workflow](https://github.com/apollographql/apollo-ios-dev/actions/workflows/publish-release.yml) in apollo-ios-dev
```
BobaFetters, line 70: Do we need this step? The `Publish Release` workflow should run automatically when merging release PRs

calvincestari:
Merging and we'll monitor for any issues when we do the next release.

.github/workflows/create-release-pr.yml
calvincestari: The trigger is set to `workflow_dispatch` which from the [Actions documentation indicates it's a manual trigger](https://github.com/apollographql/apollo-ios-dev/blob/main/.github/workflows/publish-release.yml#L4). Am I interpreting that incorrectly?

BobaFetters: Yea so that means the workflow can be manually triggered, I believe it's also needed to allow dispatching from other workflows (although i could be wrong there) but we have it being dispatched [here](https://github.com/apollographql/apollo-ios-dev/blob/main/.github/workflows/pr-subtree-push.yml#L76)

calvincestari: Ah, I see. OK, I'll get another PR up to remove that step - apologies.

PR #438, opened 2024-07-23
Fix crash when GraphQLError is “too many validation errors”

AnthonyMDev:
When a GraphQLError from the JS parsing step is a “Too many validation errors” error, there is no `source` in the error object. This was causing a crash. We should make this an optional parameter and check for it to avoid this edge case.

Writing a unit test for this is difficult because we would need to write a large schema document with many errors in it. This is a minor edge case, and I think that an inline comment explaining the problem should suffice to prevent regressions here.


calvincestari (approved):
LGTM!

PR #439, opened 2024-07-24
fix: Cache write interceptor should gracefully handle missing cache records

calvincestari:
The defer execution changes merged from PR https://github.com/apollographql/apollo-ios-dev/pull/413 introduced a bug where the cache write interceptor would throw if no cache records were returned. This is incorrect as it is a valid response for the `data` key to be missing and only an `errors` key returned, such as in the case of a [request error](https://spec.graphql.org/draft/#sec-Errors.Request-Errors).

This changeset is to allow the cache write interceptor to gracefully handle missing cache records and only publish records to the cache if they were returned from parsing.
* Adds a test that verifies the default interceptor chain successfully parses a request error
* Removes deprecation attribute for `noResponseToParse` error
* Removes error `missingCacheRecords` since it's no longer valid (breaking change)
* Moves the chain cancellation check earlier in the `interceptAsync` of `CacheWriteInterceptor` method to return at the earliest point

_This was first reported as a result of https://github.com/apollographql/apollo-ios-dev/pull/428 not getting a clean CI build._

### DIFF
--- a/apollo-ios/Sources/Apollo/CacheWriteInterceptor.swift
+++ b/apollo-ios/Sources/Apollo/CacheWriteInterceptor.swift
@@ -7,17 +7,12 @@ import ApolloAPI
 public struct CacheWriteInterceptor: ApolloInterceptor {
 
   public enum CacheWriteError: Error, LocalizedError {
-    @available(*, deprecated, message: "Will be removed in a future version.")
     case noResponseToParse
-
-    case missingCacheRecords
 
     public var errorDescription: String? {
       switch self {
       case .noResponseToParse:
         return "The Cache Write Interceptor was called before a response was received to be parsed. Double-check the order of your interceptors."
-      case .missingCacheRecords:
-        return "The Cache Write Interceptor cannot find any cache records. Double-check the order of your interceptors."
       }
     }
   }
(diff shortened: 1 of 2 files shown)


apollo-ios/Sources/Apollo/CacheWriteInterceptor.swift
```diff
@@ -7,17 +7,12 @@ import ApolloAPI
 public struct CacheWriteInterceptor: ApolloInterceptor {
 
   public enum CacheWriteError: Error, LocalizedError {
-    @available(*, deprecated, message: "Will be removed in a future version.")
```
calvincestari, line 10: This was deprecated because in the defer execution PR it was no longer being used. Had we caught this bug then it would never have been deprecated.

apollo-ios/Sources/Apollo/CacheWriteInterceptor.swift
```diff
@@ -7,17 +7,12 @@ import ApolloAPI
...
     case noResponseToParse
 
-    case missingCacheRecords
```
calvincestari, line 13: Arguably this is a breaking change but given the short availability of release `1.14.0` I don't see this as being too impactful. Like the deprecation of `noResponseToParse` it would not have been introduced if we'd caught the bug earlier.

apollo-ios/Sources/Apollo/CacheWriteInterceptor.swift
```diff
@@ -37,7 +32,11 @@ public struct CacheWriteInterceptor: ApolloInterceptor {
     request: HTTPRequest<Operation>,
     response: HTTPResponse<Operation>?,
     completion: @escaping (Result<GraphQLResult<Operation.Data>, any Error>) -> Void) {
-    
+
+    guard !chain.isCancelled else {
```
calvincestari, line 36: I've moved this earlier in the `interceptAsync` logic because it makes sense that if the chain has already been cancelled you would not want to receive any response, nor should the interceptor chain proceed forward.

AnthonyMDev (approved):
LGMT. Thanks for the fix!

PR #440, opened 2024-07-25
fix: Introspection-based schema download creates duplicate `@defer` directive

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3417

This changeset splits the logic for adding the defer directive to the document between SDL and introspection sources. SDL sources can be inspected as a `DocumentNode` with all declared directives available but introspection sources, once converted to a `DocumentNode`, do not contain all declared directives and instead need to be inspected as the schema built from the introspection source; I do not know whether this is a bug in graphql-js or intended behaviour but I will take that up with them separately.

* Adds JS tests for introspection sources with the same pattern as SDL source tests; valid directive, unsupported directive, and no directive.
* Adds a Swift-based introspection test to the `ApolloSchemaDownloaderInternalTests` to ensure that even though the JS logic has been fixed there isn't some fetch-schema-workflow specific difference to creating the duplicate directive.
* Splits the defer directive addition logic between SDL and introspection sources.


apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/__tests__/deferDirectiveTests.ts
```diff
@@ -461,5 +464,162 @@ describe("given schema with valid defer directive", () => {
...
+        }
+      }
+    }
+    `;
+
+    const document: DocumentNode = parseOperationDocument(
+      new Source(documentString, "Test Query", { line: 1, column: 1 })
+    );
+
+    it("ensure the introspection JSON is correct for test", () => {
```
calvincestari, line 492: Introspection JSON is difficult to parse and read so I've added tests like this one to ensure that the JSON doesn't get changed in a way that would invalidate what the test is meant to be checking for.

BobaFetters (approved):
LGTM

PR #442, opened 2024-07-27
docs: Fixes the tutorial server URL

calvincestari:
The URL we had does not work in explorer. It is also different from the URL used in the schema download configuration JSON sample.


calvincestari:
Merging this without review since it just fixes a URL already referenced elsewhere.

PR #444, opened 2024-07-27
feat: async query pager fetch with cache policy

pondorasti:
resolves https://github.com/apollographql/apollo-ios/issues/3418


BobaFetters:
@AnthonyMDev @Iron-Ham Any reason either of you have that we wouldn't want to do something like this?

Iron-Ham:
> @AnthonyMDev @Iron-Ham Any reason either of you have that we wouldn't want to do something like this?

I'd love to learn more about this use-case. I have some concerns that this would result in unexpected outcomes (when fetched using `returnCacheDataDontFetch` or `returnCacheDataElseFetch`) since the cached pagination information may rely on unstable information (`offset`s, some `cursor` implementations).

pondorasti:
> Is the idea that this data has already been fetched elsewhere in the app (in an existing session) and should be reused?

The app is structured around many View/ViewModel components that are more or less tied to a feature. There's very few environment objects passed down (i.e. user model). So each part of the app defines it's own GraphQL queries and heavily relies on the normalized cache to avoid a crazy number of fetch requests. Also since all the queries are implemented with watchers it's very easy to keep state in sync when the normalized cache is updated directly or indirectly.

In the particular case of pagination, we have a view that can be presented through multiple entry points like you mentioned. One thing to keep in mind though with the View/ViewModel pattern described above is that when views are popped from the screen, swift also garbage collects any state and view models that came with them. So even when you go back and forth on the same View, we end up triggering a bunch of fetch requests with the current apollo pagination package.



> Otherwise... I think we could accomplish the same behavior by removing the `assert` from `refetch`, which allows users to specify their policy?

This would be just as good as my modification to the fetch API. Before I created this PR I did try making everything of based on `refetch(cachePolicy: returnCacheDataDontFetch)` but the asserts where getting in the way so I opened this PR.

pondorasti:
> Can we instead just remove the asserts from refetch?

Done!

> Can we make sure this applies to both the GraphQLQueryPager and AsyncGraphQLQueryPager?

Looks like `GraphQLQueryPagerCoordinator` is built on top of the `AsyncGraphQLQueryPagerCoordinator`, so removing the `assert` from the async pager will propagate to the sync pager too.

https://github.com/apollographql/apollo-ios-dev/blob/94be72076b944f5957dd520f45e2a72868a06a04/apollo-ios-pagination/Sources/ApolloPagination/GraphQLQueryPagerCoordinator.swift#L39-L41

Iron-Ham:
Looks good to me! I'm happy with the `assert` removal as a path forward. It's a bit less opinionated, and makes the minimal change to get this feature in place.

What do y'all think @BobaFetters @AnthonyMDev?

AnthonyMDev (approved):
Nice and simple! Thanks for working through this @pondorasti and @Iron-Ham!

PR #449, opened 2024-08-01
Fix newline at end of generated operation

AnthonyMDev:
The use of the `\(section:)` interpolation method for `DeferredFragmentsMetadataTemplate` was causing the new line at the end of the generated operation files to be deleted. This corrects that.


calvincestari (approved):
Thanks for the fix.

PR #456, opened 2024-08-07
fix: `legacyResponse` property

calvincestari:
When the `legacyResponse` property of `HTTPResponse` was deprecated setting the value was also removed; this was incorrect as it created a hidden breaking change for interceptors that might have been using the value.

The value cannot simply be set though, as it was before, because of Swift's inability to disable the deprecation warning. Instead there is now an internal property that stores the legacy response value and `legacyResponse` becomes a computed property that simply forwards the new internal property value.

The `cacheRecords` property that was added to eliminate the need to re-parse the legacy response has also been made public for interceptors that need cache records and do not want to have to parse the response a second time.

### DIFF
--- a/apollo-ios/Sources/Apollo/IncrementalJSONResponseParsingInterceptor.swift
+++ b/apollo-ios/Sources/Apollo/IncrementalJSONResponseParsingInterceptor.swift
@@ -101,14 +101,15 @@ public struct IncrementalJSONResponseParsingInterceptor: ApolloInterceptor {
           }
         }
 
+        createdResponse._legacyResponse = nil
+
         parsedResult = currentResult
         parsedCacheRecords = currentCacheRecords
 
       } else {
-        let graphQLResponse = GraphQLResponse(
-          operation: request.operation,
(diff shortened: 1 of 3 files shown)


BobaFetters (approved):
LGTM

apollo-ios/Sources/Apollo/IncrementalJSONResponseParsingInterceptor.swift
```diff
@@ -101,14 +101,15 @@ public struct IncrementalJSONResponseParsingInterceptor: ApolloInterceptor {
           }
         }
 
+        createdResponse._legacyResponse = nil
```
calvincestari, line 104: I think this is the correct action to take here. There is nothing to set it to other than `nil` because this code path is for incremental responses. I think leaving the initial (partial) response in `_legacyResponse` is also incorrect because it's not accurate. The 'legacy' part of the response logic could also be argued to mean pre-defer-support in which case `nil` again seems the correct choice.

PR #457, opened 2024-08-08
Fix a leak related to use of function used in a closure

prabhuamol:
There is a leak in fileprivate lazy var loader: DataLoader<CacheKey, Record> because the method implicitly uses self. Making the capture explicit and breaking the retain cycle here. This was also covered in WWDC 2024 analyze heap memory talk https://developer.apple.com/wwdc24/10173?time=1748

Leak detected via instruments
Screenshot 2024-08-07 at 11 41 24 PM


calvincestari (approved):
Thanks for the contribution @prabhuamol.

PR #459, opened 2024-08-13
fix: `ObjectData` type check

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3414

Checking whether `AnyHashable` can be cast to a `Bool` is not enough to infer behaviour from the original type. The only way to do that is to compare against the two possible `Bool` values; `true` and `false`. This is possible because of the way `AnyHashable` does equality comparison.

### DIFF
--- a/apollo-ios/Sources/ApolloAPI/ObjectData.swift
+++ b/apollo-ios/Sources/ApolloAPI/ObjectData.swift
@@ -11,6 +11,9 @@ public struct ObjectData {
   public let _transformer: any _ObjectData_Transformer
   public let _rawData: [String: AnyHashable]
 
+  private static let boolTrue = AnyHashable(true)
+  private static let boolFalse = AnyHashable(false)
+
   public init(
     _transformer: any _ObjectData_Transformer,
     _rawData: [String: AnyHashable]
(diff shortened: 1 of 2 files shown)


calvincestari:
Putting this into draft since I realized it can't use the static constants due to being a public function.

calvincestari:
I can't think of a better non-intrusive way to fix this. The static constants needed to be made public because the `ObjectData` subscript is public. I don't like that they're public but I've underscored them to signal their 'private' intent and having them as static constants negates the performance cost of having to create them for each comparison.

apollo-ios/Sources/ApolloAPI/ObjectData.swift
```diff
@@ -11,6 +11,9 @@ public struct ObjectData {
   public let _transformer: any _ObjectData_Transformer
   public let _rawData: [String: AnyHashable]
 
+  public static let _boolTrue = AnyHashable(true)
+  public static let _boolFalse = AnyHashable(false)
```
AnthonyMDev, line 15: I believe the need for these to be accessible is due to the subscript being `@inlinable`. We should be able to make these `@usableFromInline private` instead of `public`.

AnthonyMDev (approved):
Would like us to try using `@usableFromInline` for those constants. Otherwise this LGTM!

apollo-ios/Sources/ApolloAPI/ObjectData.swift
calvincestari: TIL - that's great to know! They're required to be `@usableFromInline internal` rather than `private` but that achieves what I wanted. Thanks you!

PR #460, opened 2024-08-15
fix: `SelectionSetTemplate` scope comparison

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3424.

Refactored the selection set template scope comparison to account for an edge case in merged sources.


AnthonyMDev (approved):
Thanks for cleaning this up!

PR #463, opened 2024-08-18
Add additional capitalization rules

tonyarnold:
As discussed in apollographql/apollo-ios#3419, I'm proposing changes to add additional, configurable rules around capitalisation of type, property and other names via code-gen, such that I can control the output of `Id` to be `ID`, and similar. 

This is an initial pass to gather feedback on the configuration changes, and do some bike shedding prior to implementing the underlying feature.

### DIFF
--- a/apollo-ios-codegen/Sources/ApolloCodegenLib/Capitalizer.swift
+++ b/apollo-ios-codegen/Sources/ApolloCodegenLib/Capitalizer.swift
@@ -1,0 +1,9 @@
+import Foundation
+
+public enum CapitalizationRule: Codable, Equatable {
+  case uppercase(regex: String)
+  case lowercase(regex: String)
+  case titlecase(regex: String)
+  case camelcase(regex: String)
+  case pascalcase(regex: String)
+}
--- a/apollo-ios-codegen/Sources/ApolloCodegenLib/Capitalizer.swift
+++ b/apollo-ios-codegen/Sources/ApolloCodegenLib/Capitalizer.swift
@@ -1,9 +1,23 @@
 import Foundation
 
-public enum CapitalizationRule: Codable, Equatable {
-  case uppercase(regex: String)
-  case lowercase(regex: String)
-  case titlecase(regex: String)
-  case camelcase(regex: String)
-  case pascalcase(regex: String)
+public struct CapitalizationRule: Codable, Equatable, Sendable {
+  public enum Term: Codable, Equatable, Sendable {
+    case string(String)
+    case regex(String)
+  }
+
+  public enum CaseStrategy: String, Codable, Equatable, Sendable {
+    case upper
+    case lower
+    case camel
+    case pascal
+  }
+
+  public let term: Term
+  public let strategy: CaseStrategy
+
+  public init(term: Term, strategy: CaseStrategy) {
+    self.term = term
+    self.strategy = strategy
+  }
 }
--- a/Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
+++ b/Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
@@ -38,6 +38,9 @@ class ApolloCodegenConfigurationCodableTests: XCTestCase {
           testMocks: .swiftPackage(targetName: "SchemaTestMocks")
         ),
         options: .init(
+          additionalCapitalizationRules: [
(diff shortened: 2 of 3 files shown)


apollo-ios-codegen/Sources/ApolloCodegenLib/Capitalizer.swift
```diff
@@ -0,0 +1,9 @@
+import Foundation
+
+public enum CapitalizationRule: Codable, Equatable {
+  case uppercase(regex: String)
```
BobaFetters, line 4: Wondering if a regex pattern is the right way to handle this vs allowing explicit declaration of the type you want to change capitalization for and what you want it to be through a dictionary similar to how we handle the schema type renaming.

Thoughts @AnthonyMDev @calvincestari ?

AnthonyMDev: I'm not sure. That's a good question. But we aren't just talking about types here, we're also talking about their properties. 

If you used `.uppercase(regex: "id"), that could also result in situations where a property `isHidden` gets changed to `isHIDden`, for example. You would need to make your regex `^id$` if you wanted to just match the exact string `id`, which I guess is alright, but not super easy to figure out if you're not very familiar with regex.

However, I think it's likely that there will be situations where you want to tell the codegen engine "uppercase all instances of the letters 'UUID'. This may show up in multiple different property names, and is less likely to have conflicts where it shows up in the middle of another word. A regex would be helpful in those situations.

There is also situations where you want to capitalize a term in a specific way, like `GraphQL`. I think there should be a way to do that too. This doesn't fall into `camelcase` exactly correctly.

---

I think a more robust API would allow users to use _either a regex or a direct string literal_ and allow them to choose a common casing strategy or a direct replacement string.

```swift
struct CapitalizationRule {

  enum Term {
    case string(String)
    case regex(String)
  }

  enum Strategy {
    case uppercase
    case lowercase
    case titlecase
    case camelcase
    case pascalcase
    case replace(String)
  }

  let term: Term
  let strategy: Strategy
}
```

This would allow for all of the desired capabilities to exist. My only concern is that the `replace` strategy allows for way more use cases outside of just capitalization. It might be used as a hacky workaround for other desired name changes that are not the intended usage of this API.

What do you think @tonyarnold @BobaFetters?

AnthonyMDev (commented):
Thanks for starting this up @tonyarnold!

We need to start by defining the scope of where this applies. This would definitely apply to:

- Field property accessor names
- Generated selection set type names

Should it also apply to schema types? It does seem reasonable that you would expect it to. But we also have the ability to rename schema types on an individual basis, so that already accounts for this functionality on schema types. It doesn't feel great to me to have multiple ways of doing the same thing. On the other hand, I could see it being confusing if you use this, expecting it to capitalize your schema types and it doesn't. I think I lean towards going ahead and applying this to schema types, but would appreciate feedback.

apollo-ios-codegen/Sources/ApolloCodegenLib/Capitalizer.swift
```diff
@@ -0,0 +1,9 @@
...
+  case lowercase(regex: String)
+  case titlecase(regex: String)
+  case camelcase(regex: String)
```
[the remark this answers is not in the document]
tonyarnold: Camelcase capitalizes the first letter of each word, but removes the spaces — titlecase does not remove the spaces — making it wholly unsuitable for this purpose. I'll remove it.

apollo-ios-codegen/Sources/ApolloCodegenLib/Capitalizer.swift
tonyarnold: I quite like this proposed API! I'll update the PR to reflect it. 

We could potentially leave `.replace(…)` out of the strategies?

apollo-ios-codegen/Sources/ApolloCodegenLib/Capitalizer.swift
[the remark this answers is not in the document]
tonyarnold: Done in 0638c5f72212a45928951267494353b3b32f6741.

calvincestari:
I'm closing this issue due to lack of activity.

@tonyarnold - if you feel this is done from your end and you're waiting on us to move things along please comment in this PR and we can take another look at where we're at.

PR #465, opened 2024-08-20
Fix flaky OffsetPagination test

Iron-Ham:

I noticed this test was flaky. This PR fixes that flaky test.

This test is identical to the `AsyncGraphQLQueryPagerTests.test_concatenatesPages_matchingInitialAndPaginated` test, but for use with an offset-based pagination strategy. I updated the test to follow the same logic as that test, which is more reliable.

No.

### Risk Assessment and Mitigation

- [ ] **No risk**

Change does not run in production. No impact on users. (Documentation or some configuration changes)

### Test Plan

CI Passes. 
I ran this test 100 times repeatedly, 5x (i.e., 500 runs) locally and it passed every time.

### DIFF
--- a/Tests/ApolloPaginationTests/OffsetTests.swift
+++ b/Tests/ApolloPaginationTests/OffsetTests.swift
@@ -89,8 +89,13 @@ final class OffsetTests: XCTestCase {
     }
 
     let pager = await createPager()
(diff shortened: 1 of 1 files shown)


AnthonyMDev (approved):
Thanks for the fix @Iron-Ham!

PR #467, opened 2024-08-22
Bring back deprecated CodingKey for localCacheMutation in selectionSetInitializers

AnthonyMDev:
This fixes a decoding error when the deprecated option `localCacheMutations` is present in the JSON config. We should ignore the key when decoding, but not throw an error if it's present.

This issue was surfaced from [this comment on an unrelated issue](https://github.com/apollographql/apollo-ios/issues/3428#issuecomment-2305590461).


calvincestari (approved):
LGTM!

PR #470, opened 2024-08-24
Make required mock fields nonoptional in convenience initializers

dwroth:
Makes the mock initialization parameters corresponding to nonull fields nonoptional. This allows the compiler to detect an invalid mock, rather than allowing an incomplete mock to cause an error at runtime.

Also fixes a bug in which the rendered type of a deeply nested mock field would be typed as required, even though it might be defined as optional.


AnthonyMDev:
Thank you for making a PR @dwroth, but the reason why all fields are optional on the mocks is because, even if a field on a type is non-null, it may not have been selected and fetched for any given operation. You should not be required to set values for fields if they are not selected in the operation that you are trying to mock.

> Also fixes a bug in which the rendered type of a deeply nested mock field would be typed as required, even though it might be defined as optional.

I'm not sure what you are referring to here, and without a unit test as an example, I can't tell if this is expected behavior or not. If you still believe this to be a bug, lets talk about it further.

dwroth:
I see, that makes sense.  Thank you for clearing that up @AnthonyMDev .

As for the bug I was referring to (assuming I'm reading this code correctly):  consider the `customScalarOptionalList` defined in MockObjectTemplateTests.  It is an optional list of optional scalars.  Yet the output mock template has it typed as an optional list of non-optional scalars.

PR #473, opened 2024-08-30
fix: Cherry pick `ObjectData` type check to `ListData`

calvincestari:
Cherry-picking the same type data check from `ObjectData` into `ListData`. See https://github.com/apollographql/apollo-ios-dev/pull/459.


AnthonyMDev (approved):
Thanks for this! LGTM

PR #476, opened 2024-09-11
enhancement: Set URLRequest cache policy on GET requests

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3432.

There is currently no way for a user to opt-out of the URLRequest default caching behaviour. This PR simply sets a comparable cache policy on `URLRequest` for `GET` requests.

* `returnCacheDataElseFetch` == `returnCacheDataElseLoad`: Direct comparison
* `fetchIgnoringCacheData` == `reloadIgnoringLocalCacheData`: The Apollo iOS version of this cache policy means that only the request will ignore cached data but the response will still be cached. It seems correct to ignore any iOS cached responses but still use any remote (CDN) cached responses.
* `fetchIgnoringCacheCompletely` == `reloadIgnoringLocalAndRemoteCacheData`: The Apollo iOS version of this cache policy will ignore cached data for the request but it will also not cache any response data. The equivalent `URLRequest` policy here is to ignore all local (iOS) and remote (CDN) cached data and always go to the server for a response.
* `returnCacheDataDontFetch` == `returnCacheDataDontLoad`: Direct comparison
* `returnCacheDataAndFetch` == `reloadRevalidatingCacheData`: The Apollo iOS request chain will handle the return-cache-data portion of this policy but we extend that to any iOS cached responses, for and-fetch,  by making iOS ensure any cached responses are still valid, otherwise fetch from the server.

### DIFF
--- a/apollo-ios/Sources/Apollo/JSONRequest.swift
+++ b/apollo-ios/Sources/Apollo/JSONRequest.swift
@@ -98,7 +98,8 @@ open class JSONRequest<Operation: GraphQLOperation>: HTTPRequest<Operation> {
       if let urlForGet = transformer.createGetURL() {
         request.url = urlForGet
         request.httpMethod = GraphQLHTTPMethod.GET.rawValue
-        
+        request.cachePolicy = requestCachePolicy
+
         // GET requests shouldn't have a content-type since they do not provide actual content.
         request.allHTTPHeaderFields?.removeValue(forKey: "Content-Type")
       } else {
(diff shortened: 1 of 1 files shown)


BobaFetters (approved):
This LGTM, think the mappings of the cache policies makes sense

calvincestari:
> This LGTM, think the mappings of the cache policies makes sense

The only question I have is whether the following should ignore remote caches or not?
> * `fetchIgnoringCacheData` == `reloadIgnoringLocalCacheData`: The Apollo iOS version of this cache policy means that only the request will ignore cached data but the response will still be cached. It seems correct to ignore any iOS cached responses but still use any remote (CDN) cached responses.

BobaFetters:
> The only question I have is whether the following should ignore remote caches or not?
> 
> > * `fetchIgnoringCacheData` == `reloadIgnoringLocalCacheData`: The Apollo iOS version of this cache policy means that only the request will ignore cached data but the response will still be cached. It seems correct to ignore any iOS cached responses but still use any remote (CDN) cached responses.

Yea I think thats a reasonable assumption that it should ignore anything cached on the iOS client side which is what Apollo iOS would be doing, but could still use remote cached responses.

AnthonyMDev (approved):
Thanks Calvin!

PR #479, opened 2024-09-13
chore: Required swift-tools-version update

calvincestari:
* Adds a comment to the Apollo libraries indicating which version of Xcode provides the required Swift Tools version.
* Updates the SPM package template swift-tools-version to match the Apollo libraries.

PR #484, opened 2024-09-17
fix: Update Codegen generated SPM swift-tools-version check

calvincestari:
The `swift-tools-version` number in generated package manifests was recently updated in #479. We didn't update the test though; _it's debatable whether this test actually has much value to be honest._ Also related to #483.


AnthonyMDev (approved):
Yeah, these tests are kinda silly. I'm not opposed to removing them.

calvincestari:
> Yeah, these tests are kinda silly. I'm not opposed to removing them.

@AnthonyMDev, I've instead removed the two 'boilerplate' tests and regenerated the internal unit test projects in the latest PR commits.

PR #485, opened 2024-09-18
fix: Removes local cache mutation type condition setter

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3443.

Simply removes the setter for mutable inline fragments (type conditions). The correct way to initialize with the type condition is to use `asRootEntityType`.


AnthonyMDev (approved):
Perfect. Nice and easy fix!

jimisaacs:
Is there a document change for this?

jimisaacs:
It is unclear how I can move setting an `self.asSomething = ` to using `asRootEntityType`?

calvincestari:
Hi @jimisaacs. We mention how to use `asRootEntityType` in the selection set initializer docs; there is a section for [Type Conditions](https://www.apollographql.com/docs/ios/development/selection-set-initializers#type-conditions).

jimisaacs:
> Hi @jimisaacs. We mention how to use `asRootEntityType` in the selection set initializer docs; there is a section for [Type Conditions](https://www.apollographql.com/docs/ios/development/selection-set-initializers#type-conditions).

Thanks! I think I figured it out.

PR #490, opened 2024-09-24
Support decoding Decimal as String

jessey2024:
Allows converting `Decimal` to `String` when decoding.
Currently, an error is being thrown.

### DIFF
--- a/Tests/ApolloTests/JSONTests.swift
+++ b/Tests/ApolloTests/JSONTests.swift
@@ -172,4 +172,18 @@ class JSONTests: XCTestCase {
     let converted = JSONConverter.convert(result)
     XCTAssertEqual(converted, expected)
   }
+
+  func testJSONDecodeDecimalAsString() throws {
+    let decimal1 = Decimal(0.0123456789)
+    let decimal2 = Decimal(0.0123456789e-10)
+
+    let jsonValue1 = JSONValue(decimal1)
+    let jsonValue2 = JSONValue(decimal2)
+
+    let string1 = try String(_jsonValue: jsonValue1)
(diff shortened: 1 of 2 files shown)


calvincestari:
Hi @jessey2024, can you give me some more context on how this is needed or where it's useful - thanks.

AnthonyMDev (commented):
Hi @jessey2024, thanks so much for the PR! This doesn't seem to be likely to cause any issues, but I'm unclear of when you would need this? This function is used for converting a JSON value into a String. Using `JSONSerialization`, wouldn't any decimal number be serialized as a `Double`? In what situation is this code actually being called?

AnthonyMDev:
Ha, I didn't see @calvincestari's response before replying here. :)

jessey2024:
> Hi @jessey2024, can you give me some more context on how this is needed or where it's useful - thanks.

Hi @calvincestari 
We currently have values of different number types returning from the API. 
In our project, these values are defined as Float8, which is aliased to Swift.String.
This works fine (with integers and floats) until we receive a Decimal value, and the decoding error is thrown.

calvincestari:
OK, that makes more sense. I suggest reading our documentation about [Custom Scalars](https://www.apollographql.com/docs/ios/custom-scalars) as that would be the correct way to deal with types outside of the [standard GraphQL scalars](https://spec.graphql.org/draft/#sec-Scalars).

calvincestari:
@jessey2024 - I'm going to close this issue as I believe [Custom Scalars](https://www.apollographql.com/docs/ios/custom-scalars) is the correct way to do what you need with those types. If you try that and still have errors then comment in this issue again and we can take another look.

PR #493, opened 2024-09-27
Add async/await to TestCacheProvider

AnthonyMDev:
This adds `async/await` support and improves the teardown logic of this test helper.

PR #496, opened 2024-09-30
Fix MockSchemaMetadata actor isolation

AnthonyMDev:
This is a fix for broken tests due to #493 being accidentally merged too early.

In order to properly support Swift Concurrency, tests that alter the MockSchemaMetadata must be marked as `MainActor`.


calvincestari (approved):
These concurrency changes just end up sprawling everywhere. Thanks for the fix.

PR #497, opened 2024-09-30
Remove server integration tests

AnthonyMDev:
After checking with @calvincestari that all behavior has adequate unit tests, we are removing the outdated and unmaintained server integration tests.


calvincestari:
@AnthonyMDev - added one more commit here; remembered the nightly CI job on main still had the integration tests in.

calvincestari (approved):

PR #498, opened 2024-09-30
Batch writing of records to SQLite store

AnthonyMDev:
This uses `insertMany` to batch the writing of records to the SQLite cache. Previously we were running separate DB transaction/query for each individual record written. This should significantly improve performance of File I/O for cache writes.

We are investigating utilizing the [Benchmarks](https://github.com/ordo-one/package-benchmark/tree/main) package at some point in the future to measure the performance improvement, but that project is blocked for now and we should still release this performance improvement in the mean time.


calvincestari (approved):
Nicely done, thank you!

It's worth noting for anyone reading this later that this isn't batching updates from different operations. The bounds of the update are still for a single operation.

PR #502, opened 2024-10-04
fix: Multipart delimiter boundary parsing

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3453.

The multipart message parsing code was not splitting message chunks at the correct boundary. It would only look for the [dash boundary (`--graphql`)](https://github.com/apollographql/team-clients/blob/9e5914a2eb9e55e1a0882edeab45188acbf76c0e/rfcs/MultipartSubscriptionsTests.md#conventions) instead of the [full delimiter (`\r\n--graphql`)](https://github.com/apollographql/team-clients/blob/9e5914a2eb9e55e1a0882edeab45188acbf76c0e/rfcs/MultipartSubscriptionsTests.md#conventions). This meant that if the dash boundary was present in the message body the chunk would be split causing a parsing error.

* Multipart delimiter changed to match the RFC specifications.
* Refactored property and function names to match multipart RFC terminology.
* Test data updated to represent RFC specification-matching message data.
* `URLSessionClientTests` restyled and new tests added to ensure correct delimiter parsing.
* Tests added to defer and subscription protocol parsers to ensure correct delimiter parsing.
* I've also done manual integration testing with the [`client-router-e2e-tests`](https://github.com/apollographql/client-router-e2e-tests) repo test services and everything works as expected.

### DIFF
--- a/Tests/ApolloTests/Network/URLSessionClientTests.swift
+++ b/Tests/ApolloTests/Network/URLSessionClientTests.swift
@@ -301,10 +301,10 @@ class URLSessionClientTests: XCTestCase {
     let client2 = URLSessionClient(sessionConfiguration: sessionConfiguration,
                                    sessionDescription: expected)
     XCTAssertEqual(expected, client2.session.sessionDescription)
-      
+    
     client2.invalidate()
   }
-    
+
   func testTaskDescription() {
     let url = URL(string: "http://www.test.com/taskDesciption")!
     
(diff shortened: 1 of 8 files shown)


Tests/ApolloTests/Network/URLSessionClientTests.swift
```diff
@@ -337,8 +377,221 @@ class URLSessionClientTests: XCTestCase {
...
+        default:
+          XCTFail("Unexpected data received: \(data)")
+        }
+      }
+    }
+
+    self.wait(for: [expectation], timeout: 1)
+  }
+
+  func test__multipart__givenChunkContainingBoundaryStringWithoutClosingBoundary_shouldNotSplitChunk() throws {
```
asmundg, line 561: This doesn't appear to actually test the critical path.

I was a bit confused about how this managed to work, but when I put a breakpoint on L581, we're actually getting the callback from urlSession dataTask didCompleteWithError and not urlSession dataTask didReceive. Which means we're bypassing the problematic parsing code.

From what I could see earlier, this behavior is actually time sensitive, making it hard to control which parts of URLSessionClient actually get exercised. Which is why I ended up calling its data handler directly.

calvincestari: > I was a bit confused about how this managed to work, but when I put a breakpoint on L581, we're actually getting the callback from urlSession dataTask didCompleteWithError and not urlSession dataTask didReceive. Which means we're bypassing the problematic parsing code.

This is an implementation detail; there is no guarantee the different callbacks nor the logic within them will remain the same in the future. The test message data is what's important and there are multiple tests to cover different combinations of the test data.
- `test__parsing__givenSingleChunk_withDashBoundaryInMessageBody_shouldReturnSuccess`
- `test__multipart__givenChunkContainingBoundaryString_shouldNotSplitChunk`
- `test__multipart__givenChunkContainingBoundaryStringWithoutClosingBoundary_shouldNotSplitChunk`

calvincestari: > From what I could see earlier, this behavior is actually time sensitive, making it hard to control which parts of URLSessionClient actually get exercised. Which is why I ended up calling its data handler directly.

It's not time sensitive but dependent on the type of response (multipart header) and number of chunks. As data is sent in it's checked for complete chunks so they can be processed immediately; this is the callback that needs to check chunk boundaries. Then once the request completes any left over response data is sent on for further parsing.

AnthonyMDev (approved):
This LGTM! Is there any need to test when the body content inadvertently has something that looks like a multipart boundary somewhere in the text of the JSON?

calvincestari:
> This LGTM! Is there any need to test when the body content inadvertently has something that looks like a multipart boundary somewhere in the text of the JSON?

Yes, that's what the tests I mentioned above do. Note that there are no tests for body data that contains the dash-boundary preceded nor followed by a `\r\n` because that is explicitly not allowed in [RFC2046](https://datatracker.ietf.org/doc/html/rfc2046#section-5.1.1). If that combination did appear in a message body it would be a malformed mulitpart chunk.

PR #503, opened 2024-10-09
Call the completions in the given callback queue when `CompletionManager` is getting deinitialized

haldun:
In `CompletionManager.deinit` we directly call each of the completions but instead we should use `.execute` to ensure that the completion closures are being called in the given queue.


AnthonyMDev:
Thanks for this PR! Would you be able to sign our CLA so we can merge this?

@Iron-Ham, this looks right to me, any concerns on your end?

Iron-Ham:
@AnthonyMDev it looks good to me! @haldun committed this to our internal version of `ApolloPagination` and is upstreaming the change.

AnthonyMDev (approved):

PR #506, opened 2024-10-16
fix: Websocket error broadcast for unsubscribed ID

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/2274
Closes https://github.com/apollographql/apollo-ios/issues/2447

There is a sequence of events where the caller unsubscribes which removes their ID from `subscribers` but an incoming message is then received for that subscriber. It doesn't seem correct to broadcast that as an error to all remaining subscribers for unrelated messages.

Changes:
* Only broadcast an error to all subscribers if there was no `id` field in the message
* Do not send any error if the `id` was not found in `subscribers`.

### DIFF
--- a/apollo-ios/Sources/ApolloWebSocket/WebSocketTransport.swift
+++ b/apollo-ios/Sources/ApolloWebSocket/WebSocketTransport.swift
@@ -193,10 +193,21 @@ public class WebSocketTransport {
       }
 
       switch messageType {
-      case .data,
-           .next,
-           .error:
-        if let id = parseHandler.id, let responseHandler = subscribers[id] {
+      case .data, .next, .error:
+        guard let id = parseHandler.id else {
+          let websocketError = WebSocketError(
+            payload: parseHandler.payload,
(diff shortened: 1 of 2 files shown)


apollo-ios/Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -207,11 +218,6 @@ public class WebSocketTransport {
                                                 kind: .neitherErrorNorPayloadReceived)
             responseHandler(.failure(websocketError))
           }
-        } else {
```
calvincestari, line 210: An alternative approach to silencing this error path is to move unsubscribing IDs into another array that can be checked against if the ID is not found in `subscribers`. This would ensure we're discarding a message for an unsubscribing handler but I'm not sold on that being necessary.

_The unsubscribing array should be of `[String]` type since there would be no need to store the handling closure again._

AnthonyMDev (approved):
LGTM! Thanks!

PR #508, opened 2024-10-16
Remove docs publish flow

BlenderDude:
New docs uses webhooks to deploy new versions, so the github action is no longer necessary


calvincestari (approved):
Thank fox fixing this.

PR #509, opened 2024-10-16
Make ApolloStore an open extensible class, and its previously-internal dependencies public

lincolnq:
(This is in reference to https://github.com/apollographql/apollo-ios/issues/3461, which I filed earlier today.)

The goal of this PR is: Make it easier to customize behavior of the Apollo normalized cache.

Currently, ApolloStore is a concrete class depended-on by lots of Apollo internals. It's not marked open, nor is it a protocol, so in practice there is no way without modifying Apollo code for applications to customize the store behavior.

With this PR, it is still necessary to use `@_spi(Execution)` for overriding particularly fiddly bits of ApolloStore behavior. I think this is probably a good thing to help users understand that they're digging deep into the internals, while still allowing us to customize the behavior we want.

Feedback is welcome. In particular I'm not sure whether to do what Android does and make ApolloStore a protocol and rename the current `ApolloStore` to `DefaultApolloStore`. I can make that change too if desired.

---

Feel free to skip the below, but here's some more context about the why:

My specific use-case is one that has been touched-on/requested a number of places (https://github.com/apollographql/apollo-ios/issues/3319, https://github.com/apollographql/apollo-ios/issues/892, https://github.com/apollographql/apollo-ios/issues/3216: make the Apollo cache accept missing values for nullable fields as null, rather than failing to parse them.) While Apollo team has resisted adding a flag, if ApolloStore were extensible I could implement the behavior I want myself without forking all of apollo-ios.

This example is what I was trying to write, impossible (as far as I can tell) without making ApolloStore extensible. It's just an ApolloStore that passes `.allowForOptionalFields` to the `GraphQLSelectionSetMapper` during load:

```swift
import Foundation
import ApolloAPI
@_spi(Execution) import Apollo

public class MyApolloStore: ApolloStore {
    override open func load<Operation: GraphQLOperation>(
        _ operation: Operation,
        callbackQueue: DispatchQueue? = nil,
        resultHandler: @escaping GraphQLResultHandler<Operation.Data>
    ) {
        withinReadTransaction({ transaction in
            let (data, dependentKeys) = try transaction.readObject(
                ofType: Operation.Data.self,
                withKey: CacheReference.rootCacheReference(for: Operation.operationType).key,
                variables: operation.__variables,
                accumulator: zip(GraphQLSelectionSetMapper<Operation.Data>(handleMissingValues: .allowForOptionalFields),
                                 GraphQLDependencyTracker())
            )

            return GraphQLResult(
                data: data,
                extensions: nil,
                errors: nil,
                source: .cache,
                dependentKeys: dependentKeys
            )
        }, callbackQueue: nil, completion: resultHandler)
    }
}

```

### DIFF
--- a/apollo-ios/Sources/Apollo/ApolloStore.swift
+++ b/apollo-ios/Sources/Apollo/ApolloStore.swift
@@ -21,11 +21,11 @@ public protocol ApolloStoreSubscriber: AnyObject {
 }
 
 /// The `ApolloStore` class acts as a local cache for normalized GraphQL results.
-public class ApolloStore {
-  private let cache: any NormalizedCache
-  private let queue: DispatchQueue
+open class ApolloStore {
+  public let cache: any NormalizedCache
+  public let queue: DispatchQueue
 
-  internal var subscribers: [any ApolloStoreSubscriber] = []
+  open var subscribers: [any ApolloStoreSubscriber] = []
 
   /// Designated initializer
   /// - Parameters:
@@ -36,7 +36,7 @@ public class ApolloStore {
     self.queue = DispatchQueue(label: "com.apollographql.ApolloStore", attributes: .concurrent)
   }
 
-  fileprivate func didChangeKeys(_ changedKeys: Set<CacheKey>, identifier: UUID?) {
+  open func didChangeKeys(_ changedKeys: Set<CacheKey>, identifier: UUID?) {
     for subscriber in self.subscribers {
       subscriber.store(self, didChangeKeys: changedKeys, contextIdentifier: identifier)
     }
@@ -47,7 +47,7 @@ public class ApolloStore {
   /// - Parameters:
   ///   - callbackQueue: The queue to call the completion block on. Defaults to `DispatchQueue.main`.
   ///   - completion: [optional] A completion block to be called after records are merged into the cache.
-  public func clearCache(callbackQueue: DispatchQueue = .main, completion: ((Result<Void, any Swift.Error>) -> Void)? = nil) {
+  open func clearCache(callbackQueue: DispatchQueue = .main, completion: ((Result<Void, any Swift.Error>) -> Void)? = nil) {
     queue.async(flags: .barrier) {
       let result = Result { try self.cache.clear() }
       DispatchQueue.returnResultAsyncIfNeeded(
@@ -65,7 +65,7 @@ public class ApolloStore {
   ///                 to assist in de-duping cache hits for watchers.
   ///   - callbackQueue: The queue to call the completion block on. Defaults to `DispatchQueue.main`.
   ///   - completion: [optional] A completion block to be called after records are merged into the cache.
-  public func publish(records: RecordSet, identifier: UUID? = nil, callbackQueue: DispatchQueue = .main, completion: ((Result<Void, any Swift.Error>) -> Void)? = nil) {
+  open func publish(records: RecordSet, identifier: UUID? = nil, callbackQueue: DispatchQueue = .main, completion: ((Result<Void, any Swift.Error>) -> Void)? = nil) {
     queue.async(flags: .barrier) {
       do {
         let changedKeys = try self.cache.merge(records: records)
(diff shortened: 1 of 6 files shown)


AnthonyMDev:
Thanks for this PR! I haven't had time to review this in detail yet, but I wanted to let you know that I'm working on Apollo 2.0 to support the Swift 6 Concurrency Model, and one of the big features I'm currently planning on designing is this.

Once I have a better idea of the design, I'll be updating the [2.0 RFC](https://github.com/apollographql/apollo-ios/issues/3411) with more information. I will definitely be looking at this PR in greater detail for inspiration as I iterate on the design for this in 2.0.

AnthonyMDev:
Thanks for following up @lincolnq. I just took a loot at the PR.

I'd like to allow my colleagues @calvincestari and @BobaFetters the chance to comment on this, but my inclination is towards rejecting this PR. While I do understand your needs are valid here, this is exposing a _lot_ of APIs publicly that are not really intended for public consumption.

We do want to expose a lot more extensibility in the caching mechanisms, but to be honest, a lot of these APIs are not really well designed, especially not for extensibility. We take great care to design our public APIs well and document them thoroughly.  All of these internals are very difficult to use and expose lots of possibilities for bugs if not used correctly.

A lot of these internal pieces are not set up for extensibility. They are interacting with other internals that expect them to work _exactly_ how they currently work. If you are able to manage your customizations in a way that doesn't break anything, that's great, but unless we do a lot of work to account for possible changes in behaviors people could create, this isn't appropriate to actually be public. 

From experience, every time we expose a new feature or customization point, we inevitably get a number of new requests from users who want to use it in different ways we didn't account for. These APIs are already very sensitive to change and weren't designed with end users in mind. I'm concerned this PR would just open a big can of worms for us.

This also makes it a lot harder for us to make improvements or changes in the future without making breaking changes. Some of the things you are exposing have changed internally quite a lot, and I expect will continue to change in ways that would be breaking if they were public APIs. Keeping them private allows us to safely iterate on them.

> Right now I'm maintaining a fork to add this feature, but I don't really want to keep doing that 

I do understand the frustrations that causes with staying up to date with other changes to the repo, but unfortunately, I'd recommend you keep using your own fork for these changes. Part of the value of open source is that you are welcome to do anything you want with the code, but if we make this part of the public API, we are then beholden to support users who are using them in ways that we are not currently prepared to support.

The right approach to this is to design a better, more extensible public API surface that we actually feel good about exposing to users and are prepared to support going forward. And that is something we're working on for 2.0 and beyond.

lincolnq:
Ok, sad to hear it but glad we're having the discussion.

Apollo Android uses an [interface for `ApolloStore`](https://github.com/apollographql/apollo-kotlin/blob/5736f7cbf9f7802d8bae8675fcc8f5195a61db33/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/ApolloStore.kt) and then [an internal impl of `DefaultApolloStore`](https://github.com/apollographql/apollo-kotlin/blob/5736f7cbf9f7802d8bae8675fcc8f5195a61db33/libraries/apollo-normalized-cache/src/commonMain/kotlin/com/apollographql/apollo/cache/normalized/internal/DefaultApolloStore.kt).

It's a lot less useful for my purposes, but I would still have less 'surface area' to copy, if ApolloStore was similarly a protocol instead of a concrete class. I haven't looked into how difficult it would be, but it would at least be possible, if iOS made the same change. Would you be more open to a PR that made ApolloStore into an interface?

calvincestari:
I agree with @AnthonyMDev. I'm also not comfortable with the number of internal APIs that need to be exposed, particularly around the executor. I would prefer taking a look at redesigning this with 2.0 or shortly thereafter with the idea of extensibility from the start. Making the proposed changes might open what we have now but it's not what the design was intended for which increases the maintenance cost.

We do appreciate the contribution @lincolnq but for now I'm going to close this PR.

ediazest:
>The right approach to this is to design a better, more extensible public API surface that we actually feel good about exposing to users and are prepared to support going forward. And that is something we're working on for 2.0 and beyond.

Hello @AnthonyMDev ! I would like to resurface this conversation as we are planning the update to Apollo 2.0. I can see in the new implementation that [ApolloStore](https://github.com/apollographql/apollo-ios-dev/blob/0b2eece8e08ba1e577df8976c41c1216fcb926a3/apollo-ios/Sources/Apollo/Caching/ApolloStore.swift#L152) still loads from the cache without allowing missing values and I don't immediately see a way of extending the current behaviour. 

Using the new API design, is there a way of customising the behaviour of the Apollo normalised cache?

PR #514, opened 2024-10-18
feat(apollo-ios-codegen): Stable sort schema types for SchemaMetadata

asmundg:
In some cases, the order of parsed types come out differently on different systems (unclear why). This creates problems in some situations where we're checking in the generated SchemaMetadata files, as we continuously flip back and forth between two orderings, generating git diffs.

By sorting the parsed types, we guarantee that their order is stable and no unnecessary diffs are introduced.

### DIFF
--- a/Tests/ApolloCodegenTests/CodeGenIR/IRInputObjectTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGenIR/IRInputObjectTests.swift
@@ -25,7 +25,7 @@ class IRInputObjectTests: XCTestCase {
 
   func buildSubject() async throws {
     let ir: IRBuilder = try await .mock(schema: schemaSDL, document: document)
-    subject = ir.schema.referencedTypes.inputObjects.first!
+    subject = ir.schema.referencedTypes.inputObjects[1]
   }
 
   // MARK: - Tests
--- a/apollo-ios-codegen/Sources/GraphQLCompiler/GraphQLSchema.swift
+++ b/apollo-ios-codegen/Sources/GraphQLCompiler/GraphQLSchema.swift
@@ -19,7 +19,8 @@ protocol GraphQLSchemaType: JavaScriptReferencedObject {
 
 }
 
-public class GraphQLNamedType: JavaScriptReferencedObject, @unchecked Sendable, Hashable, CustomDebugStringConvertible, GraphQLNamedItem {
+public class GraphQLNamedType: 
+  JavaScriptReferencedObject, @unchecked Sendable, Hashable, CustomDebugStringConvertible, GraphQLNamedItem {
   public let name: GraphQLName
 
   public let documentation: String?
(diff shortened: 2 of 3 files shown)


asmundg:
I'm not sure what's going on here, and I'm struggling to produce a minimal repo. But I have two machines that consistently produce a different ordering of the types in the schema loaded by the frontend (compilationResult.referencedTypes) in IRBuilder.

Tests/ApolloCodegenTests/CodeGenIR/IRInputObjectTests.swift
```diff
@@ -25,7 +25,7 @@ class IRInputObjectTests: XCTestCase {
 
   func buildSubject() async throws {
     let ir: IRBuilder = try await .mock(schema: schemaSDL, document: document)
-    subject = ir.schema.referencedTypes.inputObjects.first!
+    subject = ir.schema.referencedTypes.inputObjects[1]
```
BobaFetters, line 28: Why would we change this to the second object of the array instead of the first?

BobaFetters:
I'm surprised this issue hasn't been raised before, can you provide more info about the 2 machines where you see the differences? Such as processor type, os versions, maybe xcode versions between the machines etc? Would be useful to help understand the underlying cause of the issue.

I'm not opposed to having this be sorted though, would make that file more readable, not sure if there are any potential complications of this you can think of @AnthonyMDev ? I supposed this would cause a change for diff purposes for all users the first time they ran it.

apollo-ios-codegen/Sources/GraphQLCompiler/GraphQLSchema.swift
```diff
@@ -59,6 +59,10 @@ public class GraphQLNamedType:
     return lhs.name == rhs.name
   }
 
+  public static func < (lhs: GraphQLNamedType, rhs: GraphQLNamedType) -> Bool {
```
AnthonyMDev, line 62: I don't think adding `Comparable` to this type is appropriate. I'd rather we just a sort function at the call site in the template.

AnthonyMDev:
Yeah, it's very odd that this is happening, but I'm not opposed to sorting the types before generation. Yes, there will be a diff for existing users, but it won't be a breaking change.

Tests/ApolloCodegenTests/CodeGenIR/IRInputObjectTests.swift
asmundg: Because when the types are sorted, `Input` is the second type in referencedTypes, instead of first.

asmundg:
> I'm surprised this issue hasn't been raised before, can you provide more info about the 2 machines where you see the differences? Such as processor type, os versions, maybe xcode versions between the machines etc? Would be useful to help understand the underlying cause of the issue.

Unclear, I mainly observed a diff between my Macbook and our internal CI system where I don't have full clarity on the environment. But I found some other devs with slightly different hardware that get the alternate type order. I can see if I can do a more thorough comparison between the systems.

BobaFetters (approved):
I think this looks good, thanks for the contribution!

PR #517, opened 2024-10-23
fix: Bug with AnyHashable coercion for non-iOS platforms

VMLe:
We encountered a similar [issue](https://github.com/apollographql/apollo-ios/issues/3247) on the macOS 11..<11.3 versions, which has been [fixed](https://github.com/apollographql/apollo-ios-dev/pull/61) for iOS. 

Although the issue was observed on macOS, I suggest including all platforms that could potentially face the same problem.

---

This PR extends the `_AnyHashableCanBeCoerced` check to include macOS, watchOS, and tvOS with their respective minimum versions.

### DIFF
--- a/apollo-ios/Sources/ApolloAPI/DataDict.swift
+++ b/apollo-ios/Sources/ApolloAPI/DataDict.swift
@@ -174,7 +174,7 @@ extension DataDict {
   /// we need to do some additional unwrapping and casting of the values to avoid crashes and other
   /// run time bugs.
   public static let _AnyHashableCanBeCoerced: Bool = {
-    if #available(iOS 14.5, *) {
+    if #available(iOS 14.5, macOS 11.3, tvOS 14.5, watchOS 7.4, *) {
       return true
     } else {
       return false


AnthonyMDev:
Thank you for this! I'm happy to accept this PR. Just want to make sure we have the correct versions for each platform. Where did you get the versions that you added to the check for each platform from? This behavior of AnyHashable isn't well documented, and I want to make sure we are getting it right.

VMLe:
I can confirm that for macOS, AnyHashable coercion starts working correctly with macOS 11.3, based on the data we get from Sentry and our observations.

Also, all the versions I mentioned, including iOS 14.5, are related to [Xcode 12.5](https://developer.apple.com/documentation/xcode-release-notes/xcode-12_5-release-notes) and have similar release dates and release notes. ( [macOS](https://developer.apple.com/documentation/macos-release-notes/macos-big-sur-11_3-release-notes), [iOS](https://developer.apple.com/documentation/ios-ipados-release-notes/ios-ipados-14_5-release-notes), [watchOS](https://developer.apple.com/documentation/watchos-release-notes/watchos-7_4-release-notes), [tvOS](https://developer.apple.com/documentation/tvos-release-notes/tvos-14_5-release-notes) )

However, you are correct — Apple does not provide any information about this issue. Consequently, I do not have confirmation for tvOS and watchOS due to the lack of available data.

I believe that all these versions are closely related. However, if you feel that the connection between these versions is insufficient, I can remove watchOS and tvOS from this pull request.

AnthonyMDev (approved):

PR #519, opened 2024-10-25
Don't add a raw type to Enums with no generated cases

dwroth:
When an enum has no cases, or when all of the existing cases have been deprecated, the generated code will not be compilable since Xcode/swift cannot automatically synthesize conformance to RawRepresentable.

This fix simply omits the `String` raw type when the enum has no cases, which allows the build to proceed.

### DIFF
--- a/Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
@@ -491,7 +491,7 @@ class EnumTemplateTests: XCTestCase {
 
   // MARK: - No Enum Cases Tests
 
-  func test_render_givenEmptyEnum_shouldRenderWithoutAssociatedType() {
+  func test_render_givenEmptyEnum_shouldRenderWithoutRawType() {
     // given
     buildSubject(values: [], config: .mock(.other))
 
@@ -506,7 +506,7 @@ class EnumTemplateTests: XCTestCase {
     expect(actual).to(equalLineByLine(expected, ignoringExtraLines: true))
   }
 
-  func test_render_givenEnumWithAllDeprectedValues_shouldRenderWithoutAssociatedType() {
+  func test_render_givenEnumWithAllDeprectedValues_shouldRenderWithoutRawType() {
     // given
     buildSubject(
       values: [
--- a/Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
@@ -506,7 +506,7 @@ class EnumTemplateTests: XCTestCase {
     expect(actual).to(equalLineByLine(expected, ignoringExtraLines: true))
   }
 
-  func test_render_givenEnumWithAllDeprectedValues_shouldRenderWithoutRawType() {
+  func test_render_givenEnumWithAllDeprecatedValues_shouldRenderWithoutRawType() {
     // given
     buildSubject(
       values: [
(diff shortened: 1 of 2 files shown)


Tests/ApolloCodegenTests/CodeGeneration/Templates/EnumTemplateTests.swift
```diff
@@ -488,7 +488,48 @@ class EnumTemplateTests: XCTestCase {
...
+    """
+
+    // when
+    let actual = renderSubject()
+
+    // then
+    expect(actual).to(equalLineByLine(expected, ignoringExtraLines: true))
+  }
+
+  func test_render_givenEnumWithAllDeprectedValues_shouldRenderWithoutRawType() {
```
tt-pkaminski, line 509: nit - typo `deprected`

dwroth: Thank you!

tt-pkaminski:
Hey @AnthonyMDev, any thoughts on this PR? 

This addresses a bug we occasionally run into at our organization where a backend developer deprecates an enum type in the schema, and graphql codegen temporarily breaks by trying to create an enum type conforming to String without any cases. 

E.g 
```
# deprecated.graphql
enum ExperimentSignal {
    A
    B
}

# ExperimentSignal.graphql.swift

// Errors as we configure codegen not to generate deprecated enum cases
// and an enum without cases cannot conform to `String`
public extension GraphQL {
  enum ExperimentSignal: String, EnumType {
  }
}
```

AnthonyMDev:
I'm actually surprised that this builds for you. The `EnumType` protocol requires that the type be `RawRepresentable where RawValue == String`. I guess with no cases in the enum, this requirement is still fulfilled?

If it's building properly, I don't have an issue with the fix though! We had some CI tests fail due to Xcode 16.1 moving out of beta, which should be fixed now. Let us make sure this passes CI first!

AnthonyMDev:
@calvincestari, I thought this would be working when I re-run it after #522 was merged. It's still failing. Do we need to rebase this on `main` or is something else wrong?

calvincestari:
> @calvincestari, I thought this would be working when I re-run it after #522 was merged. It's still failing. Do we need to rebase this on `main` or is something else wrong?

It will need to be rebased yes. I can take care of that..

tt-pkaminski:
That's a good callout. I tried this in a playground and it does throw an error. 

```
protocol EnumType: RawRepresentable where RawValue == String {}
enum TestEnum: EnumType {}
```

I'm not too familiar with the repository set up, but do you know if generated code will build here @dwroth ?

dwroth:
That's a good point.  Let me test it against our repo to be sure.

calvincestari:
@AnthonyMDev - this is now successfully building against our tests but you make a good point about whether the change is actually valid Swift code. Which seems to be confirmed as not by @tt-pkaminski's [comment](https://github.com/apollographql/apollo-ios-dev/pull/519#issuecomment-2448237119).

dwroth:
You're right, it's not valid swift code.  🤦‍♂️ 

I'm not sure that there's actually a way around this.  Even when all of the enum cases are deprecated, there may still be references to the enum type in the project, so we probably shouldn't just remove the type entirely.  But leaving it in creates  a "does not conform to RawRepresentable" error.

dwroth:
If the enum has no cases, does it need to have an associated type of String?  Dropping that associated type would solve the problem.

calvincestari:
I'm going to put this PR into draft so it doesn't get merged accidentally. When we figure out if there is a solution we can move forward with it again.

dwroth:
Thanks, @calvincestari.

calvincestari:
I'm closing this issue due to lack of activity.

@dwroth - I don't think we managed to figure out another solution for this but if you feel this is done from your end and you're waiting on us to move things along please comment in this PR and we can take another look at where we're at.

PR #523, opened 2024-10-28
ci: Main branch health failure logs

calvincestari:
I noticed that we've been having some failures on the Main Branch Health job but the test logs aren't being saved.
- Adds the `failure()` condition so that the save steps are still run when the job fails
- Adds pagination tests (not sure why these weren't in here, maybe just forgot when the pagination lib was added)
- Some clean up


calvincestari:
Merging without review since this change only affects a nightly build action.

PR #524, opened 2024-10-29
ci: Fix `URLSessionClientTest` tests

calvincestari:
We've been getting frequent failures of the `test__request__cancellingTaskThroughClient_shouldNotCallCompletion` test in the [Main Branch Health](https://github.com/apollographql/apollo-ios-dev/actions/workflows/release-check.yml) nightly build. I [fixed the build logs](https://github.com/apollographql/apollo-ios-dev/pull/523) not being archived which lead to the following:
- We shouldn't be using `fatalError` for a missing request handler. I think it's a bad practice and the behaviour obscures the actual test failure by crashing instead.
- Improves the `test__request__cancellingTaskThroughClient_shouldNotCallCompletion` test by invoking the same delegate callback that happens when a task is cancelled. This means we don't need to wait an arbitrary amount of time for the request callback to be called and can know for certain that the test succeeded.
- Improves the `test__request__taskDescription` test by not having to wait an arbitrary amount of time for the test to exit. The test has nothing to do with the completion handler and the task description can be checked and the test satisfied immediately.

### DIFF
--- a/Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
+++ b/Tests/ApolloInternalTestHelpers/MockURLProtocol.swift
@@ -11,10 +11,10 @@ public class MockURLProtocol<RequestProvider: MockRequestProvider>: URLProtocol 
   }
   
   override public func startLoading() {
-    guard let url = self.request.url,
-          let handler = RequestProvider.requestHandlers[url] else {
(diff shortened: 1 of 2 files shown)


AnthonyMDev (approved):
Thanks so much for these improvements!

PR #529, opened 2024-10-31
fix: Assign websocket callback queue before connecting

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3466.

Assign the callback queue before connecting to prevent potential data race.


AnthonyMDev (approved):

AnthonyMDev:
This seems really difficult to test, so I don't think we need to add a test. But it might be a good idea to add a comment to ensure that nobody moves that code again.

AnthonyMDev (approved):

PR #530, opened 2024-10-31
fix: `GraphQLOperation` hash uniqueness

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3467.

Adding uniqueness to `GraphQLOperation` hashing.


apollo-ios/Sources/ApolloAPI/GraphQLOperation.swift
```diff
@@ -121,6 +121,8 @@ public extension GraphQLOperation {
   }
 
   func hash(into hasher: inout Hasher) {
+    hasher.combine(Self.operationType)
```
calvincestari, line 124: We probably don't need `operationType` since operation names must be unique across all documents. This is something we enforce during codegen validation.

AnthonyMDev (approved):

PR #534, opened 2024-11-06
docs: remove SummitCallouts

Meschreiber:
Removes callouts for iOS Summit sessions.


Meschreiber:
!docs preview

calvincestari (approved):

calvincestari:
@Meschreiber - we can ignore the failed "Check PR Metadata" check but there are also failed Netlify checks; do we need to be concerned about those? I did check the modified pages ([index](https://www.apollographql.com/docs/deploy-preview/7e75fd375c7bc1af3182/ios) and [get-started](https://www.apollographql.com/docs/deploy-preview/7e75fd375c7bc1af3182/ios/get-started)) in the deploy preview and they both render correctly so not sure what the failed checks are about.

Meschreiber:
Hmm, I think those checks should go away with future PRs, but will look into if they don't.

In the meanwhile, would you mind merging when you get a chance please? I don't have write access for the repo. TIA!

Meschreiber:
!docs preview

PR #539, opened 2024-11-15
feature: Adding codegen config support for spm module type versions

BobaFetters:
Adding new codegen config options for supplying a version to the SPM module type to allow for pointing to specific branches or local versions.

Closes apollographql/apollo-ios#3017


calvincestari (commented):
Minor comment on the case naming but otherwise I think this looks good so far. Still the swift package manager module template changes to implement right?

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -321,13 +321,67 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
         case .other:
           self = .other
         }
       }
     }
+    
+    public enum ApolloPackageVersion: Codable, Equatable {
+      case `default`
+      case branch(name: String)
+      case commitHash(hash: String)
```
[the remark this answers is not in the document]
BobaFetters: Ah yea good call thats alot better.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
```diff
@@ -862,7 +862,7 @@ class ApolloCodegenTests: XCTestCase {
       input: .init(schemaPath: schemaPath, operationSearchPaths: [operationsPath]),
       output: .init(
         schemaTypes: .init(path: directoryURL.path,
-                           moduleType: .swiftPackageManager),
+                           moduleType: .swiftPackageManager()),
```
AnthonyMDev, line 865: Oh, looks like this is going to be a slight breaking change for existing users... Even with a default value, you have to add the `()`. Not ideal, but if it has to be, then we can live with this.

I don't think Swift will let overload enum case names with different associated values, but can you try it just in case? If we could have the old `swiftPackageManager` case (and mark it deprecated), along with the new `swiftPackageManager(version:) case, that would be better for backwards compatibility.

Otherwise, the only way to make this backwards compatible I can think of would be to deprecate the old case and create the new case with a different name. Something like `swiftPackage(version:)` might make sense? What do you two think?

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -283,7 +283,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       case embeddedInTarget(name: String, accessModifier: AccessModifier = .internal)
       /// Generates a `Package.swift` file that is suitable for linking the generated schema types
       /// files to your project using Swift Package Manager.
-      case swiftPackageManager
+      case swiftPackageManager(version: ApolloPackageVersion = .default)
```
AnthonyMDev, line 286: When calling this in config, you're writing `moduleType: .swiftPackageManager(version:)` and there isn't great context on what "version" means here. I would assume I was setting the version number I want my schema module to have.

Possible other names for this parameter: 
- `apolloDependencyVersion`
- `apolloSDKVersion`
- `apolloVersion`

@calvincestari @BobaFetters, opinions?

apollo-ios-codegen/Sources/ApolloCodegenLib/Constants.swift
```diff
@@ -0,0 +1,5 @@
+import Foundation
+
+public enum Constants {
+  public static let CodegenVersion: String = "1.15.3"
```
AnthonyMDev, line 4: I'm not clear on why we need this new constants file. The version of Apollo-ios and codegen should always be the same. 

Do we just need to add this because they are going to be split into separate repos and the codegen won't have access to the constant from the Apollo library?

But wasn't the codegen already pointing to the version of Apollo though? In order to do the version mismatch validation? Why is this different than that?

apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
```diff
@@ -10,6 +10,8 @@ struct SwiftPackageManagerModuleTemplate: TemplateRenderer {
   let target: TemplateTarget = .moduleFile
 
   let config: ApolloCodegen.ConfigurationContext
+  
+  let version: ApolloCodegenConfiguration.SchemaTypesFileOutput.ApolloPackageVersion
```
AnthonyMDev, line 14: Can't this template get the version from the `config` internally? By adding this parameter, you're adding it to the synthesized initializer. Which means the call site has to access and pass in the `version`.

If you want this to be a storied property, I'd write a custom initializer that doesn't take in the `version`, just gets it from the `config` in the `init`.

Or just get the `version` from the `config` inside of the rendering function and kill the stored property altogether.

AnthonyMDev (changes requested):
Thanks for doing this Zach! A couple things I'd like us to consider here before merging this.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
BobaFetters: I can try the overload, I'm not against just deprecating the existing case and adding a new one, wouldn't take much to do and like you said avoids the minor breaking change.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
BobaFetters: Hmmm yea i guess its not even necessarily a "version" its almost a dependency type, like which type do you want? Branch name? Commit hash? etc maybe `apolloDependencyType`?

apollo-ios-codegen/Sources/ApolloCodegenLib/Constants.swift
BobaFetters: The existing constants files arent available to the `ApolloCodegenLib` so this seemed like the appropriate way to provide the version number for that library to use with the new `default` case for this work. Currently the generated `Package.swift` is not using exact version its a `from: 1.0.0`

apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/SwiftPackageManagerModuleTemplate.swift
BobaFetters: We can do it that way yea, did it this way currently because the file generator is already handling the `switch` on the module type and getting the associated value, can move this logic into the template if you feel like thats the more appropriate place to do it.

Tests/ApolloCodegenTests/ApolloCodegenTests.swift
BobaFetters: Confirmed the overload doesn't work, going to go the deprecation route with new case.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
BobaFetters: Went with `dependencyType` for now but can adjust further if need be.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -268,7 +268,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       moduleType: ModuleType
     ) {
       self.path = path
-      self.moduleType = moduleType
+      self.moduleType = moduleType == .swiftPackageManager ? .swiftPackage(dependencyType: .default) : moduleType
```
BobaFetters, line 271: Did this to ensure we always convert deprecated `.swiftPackageManager` values to the new `.swiftPackage()` values so the rest of the codebase can just use that (also decoding the deprecated version to the new one from JSON), also reduces the amount of warnings we are generating from using a deprecated case in the code.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -283,7 +283,12 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       case embeddedInTarget(name: String, accessModifier: AccessModifier = .internal)
       /// Generates a `Package.swift` file that is suitable for linking the generated schema types
       /// files to your project using Swift Package Manager.
+      @available(*, deprecated, message: "Use .swiftPackage(dependencyType:) case instead.")
       case swiftPackageManager
+      /// Generates a `Package.swift` file that is suitable for linking then generated schema types
+      /// files to your project using Swift Package Manager. Uses the `dependencyType` associated
+      /// value to determine how to setup the dependency on `apollo-ios`.
+      case swiftPackage(dependencyType: ApolloDependencyType = .default)
```
AnthonyMDev, line 291: It still isn't clear to me that this is the dependency on `apollo-ios` without reading the documentation. I think something like `apolloSDKVersion` makes it clear that we are not talking about the "dependency type" of your schema module.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
AnthonyMDev: Oooh, yeah, we can't keep a deprecation warning in the code like this. It will show up in all user's projects. Swift needs better support for suppressing deprecation warnings. It's so frustrating that this still works this way.

There are also issues with deprecating enum cases at all, since switch statements on enums need to be exhaustive, so you always have to switch on the deprecated cases and get warnings.

I almost think it would be better to just add a comment stating the case is deprecated and not using the `@available` to deprecate here for that reason.

apollo-ios-codegen/Sources/ApolloCodegenLib/Constants.swift
AnthonyMDev: Gotcha. What I was thinking of is only in the CLI, not in the `ApolloCodegenLib`.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
BobaFetters: Yea I could just switch it to doc comment, under the hood we are going to be converting everything to the new value anyways and we can just remove it with the next major release.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
calvincestari: I'm catching up on the changes now and I came here to make the same comment. @AnthonyMDev's naming suggestion certainly makes it clearer to the user which version they're configuring.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -283,7 +283,12 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       case embeddedInTarget(name: String, accessModifier: AccessModifier = .internal)
       /// Generates a `Package.swift` file that is suitable for linking the generated schema types
       /// files to your project using Swift Package Manager.
+      /// Attention: This case has been deprecated, use .swiftPackage(apolloSDKVersion:) case instead.
       case swiftPackageManager
+      /// Generates a `Package.swift` file that is suitable for linking then generated schema types
+      /// files to your project using Swift Package Manager. Uses the `dependencyType` associated
```
AnthonyMDev, line 289:
```suggestion
      /// files to your project using Swift Package Manager. Uses the `apolloSDKVersion` associated
```

AnthonyMDev (commented):
Reviewing this again got me thinking. Is it worth adding an option for users to point their Apollo dependency to their own fork? Right now you can point to a branch or commit hash, but it's still only on the url `https://github.com/apollographql/apollo-ios.git`.

@BobaFetters @calvincestari do you think that's worth doing at this time?

calvincestari:
I just had the thought that maybe allowing something other than `https://github.com/apollographql/apollo-ios.git` makes the CLI version checking less useful BUT is this whole PR change doing that? How do we now ensure generated code compatibility with the runtime used by that generated code?

AnthonyMDev:
> I just had the thought that maybe allowing something other than https://github.com/apollographql/apollo-ios.git makes the CLI version checking less useful BUT is this whole PR change doing that? How do we now ensure generated code compatibility with the runtime used by that generated code?

Likely you're just making some changes in your fork that don't affect codegen compatibility, but yeah any manual dependency here means you may generate incompatible code. That's something the user needs to accept and deal with on their own. If you're using anything but the default package version, we shouldn't even be using the version checker at all actually.

The version checker isn't foolproof right now anyways. It only runs on the CLI, not Swift Scripts, and if it can't find a package.resolved file, it won't run at all. But worse case, you generate code that doesn't compile properly and you know it when you try to build. This is just to attempt to detect that early and give you a clear error message.

But I'm actually wondering if we should move this into `ApolloCodegenLib` now. We are currently only running it when using CLI, but could probably run it while using Swift Scripts also. Doesn't need to be in this PR if we decide to do so.

calvincestari:
> But I'm actually wondering if we should move this into ApolloCodegenLib now. We are currently only running it when using CLI, but could probably run it while using Swift Scripts also. Doesn't need to be in this PR if we decide to do so.

Yes we probably should, and in a separate PR.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -321,12 +326,69 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
+          let apolloSDKVersion = try nestedContainer.decodeIfPresent(ApolloSDKVersion.self, forKey: .apolloSDKVersion) ?? .default
+          self = .swiftPackage(apolloSDKVersion: apolloSDKVersion)
 
         case .other:
           self = .other
         }
       }
+      
+      public enum ApolloSDKVersion: Codable, Equatable {
+        case `default`
```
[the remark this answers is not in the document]
BobaFetters: Planning to do all documentation (code and docs) once everything is finalized.

BobaFetters:
> Reviewing this again got me thinking. Is it worth adding an option for users to point their Apollo dependency to their own fork? Right now you can point to a branch or commit hash, but it's still only on the url `https://github.com/apollographql/apollo-ios.git`.
> 
> @BobaFetters @calvincestari do you think that's worth doing at this time?

I don't see a reason not to add that, I guess maybe add a url associated value that defaults to `apollo-ios` to the existing cases? Although we dont currently have a case that supports version numbers since for Apollo we want it to match the codegen version, so may need a new case that can support the `from` and `exact` versions.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -321,12 +326,87 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
...
+        case branch(name: String)
+        case commit(hash: String)
+        case custom(url: String, dependencyType: DependencyType, value: String)
```
BobaFetters, line 349: Also considered calling this case `fork` with same associated values

calvincestari: This can only come from a forked repo right? The `url` parameter is pretty clear on that deviating from `github.com/apollographql/apollo-ios`. In that case yes, I think `fork` is the better case name.

BobaFetters: Yea this is really meant for use specifically with forked repos, so yea I guess the `fork` name makes more sense.

AnthonyMDev (commented):
Love this API now. I think the changes we made are great and this is the right way forward now. Great work @BobaFetters!

I've got a couple minor suggestions, and we're still waiting on documentation. But the core of this is looking good!

Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
```diff
@@ -1699,10 +1705,13 @@ class ApolloCodegenConfigurationCodableTests: XCTestCase {
           "schemaTypes" : {
             "moduleType" : {
               "swiftPackage" : {
-                "apolloSDKVersion" : {
-                  "default" : {
+                "apolloSDKDependency" : {
+                  "sdkVersion" : {
+                    "default" : {
```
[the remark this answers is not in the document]
BobaFetters: Hadn't tried, can look into it, looks like this is how some other enums with associated values are being handled as well.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -268,7 +268,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       moduleType: ModuleType
     ) {
       self.path = path
-      self.moduleType = moduleType == .swiftPackageManager ? .swiftPackage(apolloSDKVersion: .default) : moduleType
+      self.moduleType = moduleType == .swiftPackageManager ? .swiftPackage(apolloSDKDependency: ModuleType.ApolloSDKDependency()) : moduleType
```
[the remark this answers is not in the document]
BobaFetters: Yea I can add that

Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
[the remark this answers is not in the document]
calvincestari: > looks like this is how some other enums with associated values are being handled as well.

Ideally we should get to making those others easier to use too.

docs/source/code-generation/codegen-configuration.mdx
```diff
@@ -106,7 +106,7 @@ The properties to configure `output` are:
 "output": {
   "schemaTypes": {
     "moduleType": {
-      "swiftPackageManager": {}
+      "swiftPackage": {}
```
calvincestari, line 109: It's long bugged me that we had `swiftPackageManager` as a [`ModuleType` option](https://www.apollographql.com/docs/ios/code-generation/codegen-configuration#module-type) and `swiftPackage` as a [`TestMocks` option](https://www.apollographql.com/docs/ios/code-generation/codegen-configuration#test-mocks); I'm happy these are more similar now. 😄

docs/source/code-generation/codegen-configuration.mdx
```diff
@@ -257,11 +258,59 @@ let configuration = ApolloCodegenConfiguration(
...
+
+This option automates the creation of an SPM package for your schema types. This generates a `Package.swift` file that is suitable for linking the generated schema types files to your project using Swift Package Manager.
+
+The `apolloSDKDependency` paramter allows you to customize the `apollo-ios` dependency of your SPM package by providing a url and SDK version. The `url` parameter allows you to customize the location of the `apollo-ios` dependency used by the generated SPM package, and has a default value that points to the main [`apollo-ios`](https://github.com/apollographql/apollo-ios) repository. The `sdkVersion` parameter allows you to set the version of the `apollo-ios` dependency to use. The default value of this is the `default` case, which uses an exact version of `apollo-ios` matching the code generation version in use which is the recommended way of handling this. The other provided options allow flexibility for doing things such as testing experimental branches, or using local versions for testing.
```
calvincestari, line 267: _I don't know why I can't make an inline edit (suggestion) for this change but here's the change._

From:
```
The default value of this is the `default` case, which uses an exact version of `apollo-ios` matching the code generation version in use which is the recommended way of handling this.
```
To: 
```
The default value is the `default` case, which uses an exact version of `apollo-ios` matching the code generation version in use. This is the recommended configuration.
```

calvincestari (approved):
Looks good to me, nice work!

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -283,7 +283,12 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       case embeddedInTarget(name: String, accessModifier: AccessModifier = .internal)
       /// Generates a `Package.swift` file that is suitable for linking the generated schema types
       /// files to your project using Swift Package Manager.
+      /// Attention: This case has been deprecated, use .swiftPackage(apolloSDKVersion:) case instead.
       case swiftPackageManager
```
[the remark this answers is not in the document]
BobaFetters: Think it just has to be removed in a major version and noted in release notes. Not much else to do without adding the actual deprecation warning.

PR #543, opened 2024-11-20
Add documentation about using GraphQLQueryWatcher with mutations

damian0815:
Ref https://github.com/apollographql/apollo-ios/issues/3472

### DIFF
--- a/apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift
+++ b/apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift
@@ -3,11 +3,14 @@ import Foundation
 import ApolloAPI
 #endif
 
-/// A `GraphQLQueryWatcher` is responsible for watching the store, and calling the result handler with a new result whenever any of the data the previous result depends on changes.
+/// A `GraphQLQueryWatcher` is responsible for watching the store, and calling the result handler with a new result
+/// whenever any of the data the previous result depends on changes. If your query shares response objects with other
+/// operations you should read about normalizing objects by cache key in our documentation. Normalized objects will
+/// update all query watchers that had received that object before.
 ///
-/// NOTE: In order for the `resultHandler` to be called when mutations are executed, you will need to implement Custom Cache Keys - see the Caching section of the Apollo iOS documentation for more details.
-///
-/// NOTE: The store retains the watcher while subscribed. You must call `cancel()` on your query watcher when you no longer need results. Failure to call `cancel()` before releasing your reference to the returned watcher will result in a memory leak.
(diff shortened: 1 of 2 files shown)


apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift
```diff
@@ -5,6 +5,8 @@ import ApolloAPI
 
 /// A `GraphQLQueryWatcher` is responsible for watching the store, and calling the result handler with a new result whenever any of the data the previous result depends on changes.
 ///
+/// NOTE: In order for the `resultHandler` to be called when mutations are executed, you will need to implement Custom Cache Keys - see the Caching section of the Apollo iOS documentation for more details.
```
calvincestari, line 8: Thanks @damian0815, some suggestions:
* This is not specific, nor limited, to mutations though. If you had another query that used the same fragment you would have the same problem as you experienced before.
* The direction to use custom cache keys is not about having the result handler called but rather about normalizing the response objects that could be shared between operations. The `resultHandler` callback being called is simply the Swift mechanism for getting the result back to your code.

What I'm saying above is that the direction to read about custom cache keys should be about why normalized objects are important for query watchers to function correctly.

calvincestari (changes requested):

apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift
damian0815: okay. i am a newbie to apollo, perhaps you could suggest the exact text and i will update the PR accordingly

calvincestari: I've pushed two commits with the relevant doc changes.

calvincestari (approved):

calvincestari:
Going to merge this despite the Docs Preview check failing; I believe that is because it's on a fork. Once merged I'll double-check the published change.

damian0815:
oh @calvincestari I see you merged it already. I'm not sure this is sufficient - the documentation about "normalized cache" in the link you've added is ambiguous about whether Apollo iOS already implements cache normalization. My reading of this document that led to my error with the cache keys not overlapping was, indeed, that Apollo iOS is already normalizing cache keys for me, so there's nothing further for me to do - which is why I was stumped when it didn't work the way I was expecting. 

On closer inspection there's a sentence "if properly configured" in the introduction, but without a link to `SchemaConfiguration.swift` the reader can't really make much use of that statement. Should I open another PR with my suggestion for improvement in that document directly?

calvincestari:
https://github.com/apollographql/apollo-ios-dev/pull/547

damian0815:
thank you that's better yes!

PR #545, opened 2024-11-22
Fix getting started button link

Meschreiber:
Just noticed these lead to 404s. I believe an absolute link needs to be used for ButtonLinks (didn't notice any others in the ios docs)


Meschreiber:
!docs preview

calvincestari (approved):

PR #548, opened 2024-11-26
feature: Emit Identifiable conformance on SelectionSets

x-sheep:
This PR changes the code generator to emit Identifiable conformance on a SelectionSet where possible.

See also https://github.com/apollographql/apollo-ios/issues/710#issuecomment-1204419597

### DIFF
--- a/Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGeneration/Templates/SelectionSet/SelectionSetTemplateTests.swift
@@ -837,6 +837,49 @@ class SelectionSetTemplateTests: XCTestCase {
     // then
     expect(actual).to(equalLineByLine(expected, atLine: 7, ignoringExtraLines: true))
   }
+  
+  func test__render_selections__givenFieldWithNonScalarID_doesNotAddIdentifiableConformance() async throws {
+    // given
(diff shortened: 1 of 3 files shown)


calvincestari:
Thanks for the PR @x-sheep. I appreciate your effort here but I don't think we can accept it into the library.

The main reason for this is the assumption this change makes that `id` is unique. While that may be true for your data it might not be true for everyone else that uses an `id` field. Take for instance a response object that is an interface named `Animal` and it returns `Dog` and `Cat` objects that conform to `Animal`. There is no guarantee that `id` is unique across both `Dog` and `Cat` instances while this change would impose the requirement that they _must_ be unique.

Our team has discussed this before and our preference is to use a [custom client directive](https://www.apollographql.com/docs/ios/client-directives) that is declaratively applied to a schema type. This eliminates assumptions and it's then the responsibility of the developer to ensure the specified field adheres to being unique. If you want to make another attempt using that approach we can certainly provide design and review guidance.

x-sheep:
@calvincestari According to Apple's documentation, [the Identifiable protocol](https://developer.apple.com/documentation/swift/identifiable) leaves the duration and scope of the identity unspecified. It is entirely valid to reuse IDs between different types of resources.

calvincestari:
I think a declarative mechanism is still a better choice for the library; it's explicit and can be applied to a wider set of use cases, such as objects that have identity but not through a field named `id`.

PR #549, opened 2024-11-27
rfc: Conformance to Identifiable protocol

x-sheep:
Here is a document for how to add [Identifiable](https://developer.apple.com/documentation/swift/identifiable) conformance to generated models using a new GraphQL directive.

This is a followup to #548 and https://github.com/apollographql/apollo-ios/issues/710#issuecomment-1204419597.

I'm currently unsure on which cases should lead to a validation error, and what should be handled silently by the codegen library. I'd love to hear your feedback.

### DIFF
--- a/apollo-ios/Design/swift-identifiable.md
+++ b/apollo-ios/Design/swift-identifiable.md
@@ -123,4 +123,4 @@ Apollo Kotlin has [custom directives](https://www.apollographql.com/docs/kotlin/
 extend type Book @typePolicy(keyFields: "id")
 ```
 
-This could be used for Identifiable conformance if a single field is provided, and caching behavior could also be matched with Apollo Kotlin. A disadvantage is that this directive is only used inside type extensions, and can't be used to mark fields directly in a query.
+This could be used for Identifiable conformance by generating a getter which returns a tuple of all listed key fields, and caching behavior could also be matched with Apollo Kotlin. A disadvantage is that this directive is only used inside type extensions, and can't be used to mark fields directly in a query.
--- a/apollo-ios/Design/swift-identifiable.md
+++ b/apollo-ios/Design/swift-identifiable.md
@@ -1,0 +1,94 @@
+# Summary
+
+This document proposes a new GraphQL directive named `@apollo_client_ios_identity`, to be used to mark a field which can uniquely identify an object.
+
+# Introduction
+
+SwiftUI makes heavy use of the [Identifiable](https://developer.apple.com/documentation/swift/identifiable) protocol, which is used to track the identity of entities across data changes. If an object with a List is replaced with a different object with the same identity, SwiftUI will animate the item changing instead of animating an insertion. Objects that do not conform to the Identifiable protocol require additional boilerplate to be usable inside SwiftUI.
+
+Apollo Client iOS could assist the developer by adding conformance to the Identifiable protocol to its generated models. Selecting the field to be used as an identity is done by adding a new GraphQL directive.
+
+# Definition
+
+The directive is defined as:
+```graphql
+directive @apollo_client_ios_identity on FIELD | FIELD_DEFINITION
+```
+
+This directive MUST be used on a field with a non-nullable scalar type. 
+
+## Usage in types
+
+A type MAY have a field annotated with the directive.
+
+```graphql
+type Animal {
+	id: ID! @apollo_client_ios_identity
+	name: String
+}
+```
+
+Within a type, the directive MUST NOT be used on more than one field.
+
+## Usage in interfaces
+
+It's possible for interfaces to use the directive on a field.
+
+```graphql
+interface Identifiable {
+	id: ID! @apollo_client_ios_identity
+}
+```
+
+Implementations of this interface MUST copy the directive to the same field.
+
+## Usage in operations
+
+It's likely that an external schema will not use this directive. In this case, an identity MAY be chosen when writing a query. The directive does not need to be included in the query sent to the GraphQL server.
+
+```graphql
+query GetAllAnimals {
+	allAnimals {
+		id @apollo_client_ios_identity
+		string
+	}
+}
+```
+
+It is allowed for a query and a type to both use the directive to describe the same field.
+
+If the server defines a field as an identity, a query SHOULD NOT choose another field. A query MUST NOT use the directive on more than one field in the same selection (unless they are in differently nested objects).
+
+## Generated code
+
+If an operation contains a field marked with the `@apollo_client_ios_identity` directive (by either the schema or the operation itself), the generated SelectionSet will have a conformance to the Identifiable protocol.
+
+```swift
+// Inline selection
+public struct Data: AnimalKingdomAPI.SelectionSet, Identifiable { /* ... */ }
+
+// Fragment selection
+public struct PetDetails: AnimalKingdomAPI.SelectionSet, Fragment, Identifiable { /* ... */ }
+```
+
+The protocol requires that the identity is accessible through a public field named exactly `id`. If the annotated field has a different name, an additional getter will be generated:
+
+```swift
+public var id: String { self.uuid }
+```
+
+## Naming conflicts
+
+If the identity field is not called `id`, but another field called `id` is present in the selection, a custom getter cannot be added. Swift does not support using another field to handle the conformance.
+
+In this case, a conformance to Identifiable SHOULD NOT be generated.
+
+# Alternatives
+
+## Automatic conformance to Identifiable
+
+The code generator could be updated to always emit a conformance to Identifiable if a scalar `id` field is present in the selection set, removing the need for a custom directive. This was suggested in Pull Request [#548](https://github.com/apollographql/apollo-ios-dev/pull/548).
+
+## Declaring identity scope
+
+The directive could be expanded to declare the scope of an identity (e.g. unique across the entire API or only one specific Database table). This makes the directive beneficial outside of Swift code generation, but it doesn't change the behavior of the Identifiable protocol. The [Swift documentation](https://developer.apple.com/documentation/swift/identifiable) states that the scope and duration of an identity is unspecified, so modifying the scope would not lead to a difference in the generated code.
--- a/apollo-ios/Design/swift-identifiable.md
+++ b/apollo-ios/Design/swift-identifiable.md
@@ -14,10 +14,9 @@ A concept of identity is required to allow response objects to be cached. The va
 
 The directive is defined as:
 ```graphql
-directive @identity(scope: IdentityScope = SELECTION) on FIELD | FIELD_DEFINITION
+directive @identity(scope: IdentityScope = TYPE) on FIELD | FIELD_DEFINITION
 
 enum IdentityScope {
-  SELECTION
   TYPE
   SERVICE
   GLOBAL
@@ -30,10 +29,9 @@ This directive MUST be used on a field with a non-nullable scalar type. Other ty
 
 The directive can have a parameter indicating the scope of the identity. They are ordered from _narrowest_ to _widest_:
 
-1. `SELECTION`: The identifier is only unique within the current list. Identifiers may be reused between different objects with the same typename.
-2. `TYPE`: The identifier is unique for the type, e.g. an auto-incrementing column from a database table.
-3. `SERVICE`: The identifier is unique across the current GraphQL Service.
-4. `GLOBAL`: The identifier is unique across all GraphQL services. This can be used by identifiers generated according to [RFC 4122](https://datatracker.ietf.org/doc/html/rfc4122) or [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) (also known as Universally Unique IDentifiers or UUIDs).
+1. `TYPE`: The identifier is unique for the type, e.g. an auto-incrementing column from a database table.
+2. `SERVICE`: The identifier is unique across the current GraphQL Service.
+3. `GLOBAL`: The identifier is unique across all GraphQL services. This can be used by identifiers generated according to [RFC 4122](https://datatracker.ietf.org/doc/html/rfc4122) or [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) (also known as Universally Unique IDentifiers or UUIDs).
 
 ## Usage in types
 
@@ -73,7 +71,7 @@ query GetAllAnimals {
 }
 ```
 
-It is allowed for a query and a type to both use the directive to describe the same field. If the schema and client define different scopes for the same field, the widest option is used. This allows client authors to widen the scope if required.
+It is allowed for a query and a type to both use the directive to describe the same field. If the client and schema use different scopes for the same field, the scope defined by the schema is used. Clients SHOULD NOT define a wider scope than declared by the schema. Codegen should emit a warning in this case.
 
 If the server defines a field as an identity, a query SHOULD NOT choose another field. A query MUST NOT use the directive on more than one field in the same selection (unless they are in differently nested objects).
 
@@ -103,7 +101,7 @@ In this case, a conformance to Identifiable SHOULD NOT be generated. Codegen sho
 
 ## Caching behavior
 
-A scope of `SELECTION` does not allow the identifier to be used as a caching key. Clients MAY use other mechanisms to determine if and how to cache the object.
+If no `@identity` directive is present, clients MAY use other mechanisms to determine if and how to cache the object.
 
 An identifier with scope `TYPE` can be combined with the `__typename` field to generate a caching key that's unique for the GraphQL Service.
 
@@ -117,7 +115,7 @@ The code generator could be updated to always emit a conformance to Identifiable
 
 ## Apollo Kotlin's @typePolicy directive
 
-Apollo Kotlin has [custom directives](https://www.apollographql.com/docs/kotlin/caching/declarative-ids) that allows for client authors to specify the caching key though pure GraphQL:
+Apollo Kotlin has [custom directives](https://www.apollographql.com/docs/kotlin/caching/declarative-ids) that allows for client authors to specify the caching key through pure GraphQL:
 
 ```graphql
 extend type Book @typePolicy(keyFields: "id")
--- a/apollo-ios/Design/swift-identifiable.md
+++ b/apollo-ios/Design/swift-identifiable.md
@@ -1,6 +1,6 @@
 # Summary
 
-This document proposes a new GraphQL directive named `@apollo_client_ios_identity`, to be used to mark a field which can uniquely identify an object.
+This document proposes a new GraphQL directive named `@identity`, to be used to mark a field which can uniquely identify an object.
 
 # Introduction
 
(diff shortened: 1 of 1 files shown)


apollo-ios/Design/swift-identifiable.md
```diff
@@ -0,0 +1,94 @@
...
+
+```swift
+public var id: String { self.uuid }
+```
+
+## Naming conflicts
+
+If the identity field is not called `id`, but another field called `id` is present in the selection, a custom getter cannot be added. Swift does not support using another field to handle the conformance.
+
+In this case, a conformance to Identifiable SHOULD NOT be generated.
```
BobaFetters, line 84: In this case we should definitely be throwing an error to notify the user of the issue, whether or not the error should actually cause code generation to fail would be the other question. I think it's acceptable for the error to be thrown and logged without failing code generation but could be persuaded either way.

apollo-ios/Design/swift-identifiable.md
```diff
@@ -0,0 +1,94 @@
...
+interface Identifiable {
+	id: ID! @apollo_client_ios_identity
+}
+```
+
+Implementations of this interface MUST copy the directive to the same field.
+
+## Usage in operations
+
+It's likely that an external schema will not use this directive. In this case, an identity MAY be chosen when writing a query. The directive does not need to be included in the query sent to the GraphQL server.
```
BobaFetters, line 47: Are we thinking that applying the directive at the query level would bubble up so that all selections of that type across all queries would receive the `Identifiable`? Wondering if it may result in a confusing result for users if one selection of a given type is Identifiable, while another use of the same type in a different selection set is not?

BobaFetters: Yea that would be my general sentiment as well, figured this scenario was at least worth discussion because of the behavior being provided. Although since this directive would really leave it up to the user to define how and when to use `Identifiable` then they have the flexibility to apply it to the type itself if they want.

calvincestari: > It doesn't seem like a good idea to let queries change the behavior of other queries.

This sentiment might be true if the only application of the directive is `Identifable` conformance; I believe it's larger than that.

apollo-ios/Design/swift-identifiable.md
calvincestari: +1 to validation.

Not adding conformance to `Identifiable` will not prevent the generated code from building. It will certainly be confusing to the user but not result in a build error so I prefer logging over failing.

calvincestari:
This is a good start, thanks @x-sheep.

As I said above, this RFC should focus on uncovering all possible uses of object identity. `Identifiable` conformance is only one but knowing how else we want to use this might influence the design. We don't need to plan to implement everything right away but we do need to think about all applications of it.

x-sheep:
> IMO the more valuable benefit is automatic cache key configuration, i.e.: it becomes schema based instead of having to declare it in code.

The final section regarding 'Scope' briefly touches on this. This document could be expanded to mention caching explicitly.

apollo-ios/Design/swift-identifiable.md
x-sheep: Can you give an example of how it would be beneficial to look at only the queries, instead of using existing GraphQL constructs like fragments or type extensions? I think it would be surprising behavior if the existence of certain queries affected how other queries are handled.

calvincestari: I'm not advocating for operation-based directives to supersede type-based directives - I agree that's bad behaviour.

This RFC should focus on object identity, not `Identifiable` conformance. Once we've expanded it to include all desired applications of the directive there may be a case where its use does influence other selection sets within the same operation. I don't know if those exist, none of us do until we expand the scope of the RFC.

x-sheep:
On the topic of caching, Apollo Kotlin uses its own mechanism that Apollo iOS currently doesn't. It adds a directive that should be used in a type extension, instead of as part of the query.
https://www.apollographql.com/docs/kotlin/caching/declarative-ids

While this could also be used for Identifiable conformance, it's not exactly the same as this directive alliows for a combination of fields to be used as the cache key.

x-sheep:
I've added an idea of a scope parameter, and made mention of how it can be used for caching purposes. WDYT?

apollo-ios/Design/swift-identifiable.md
```diff
@@ -1,29 +1,47 @@
...
+This document proposes a new GraphQL directive named `@identity`, to be used to mark a field which can uniquely identify an object.
 
 # Introduction
 
 SwiftUI makes heavy use of the [Identifiable](https://developer.apple.com/documentation/swift/identifiable) protocol, which is used to track the identity of entities across data changes. If an object with a List is replaced with a different object with the same identity, SwiftUI will animate the item changing instead of animating an insertion. Objects that do not conform to the Identifiable protocol require additional boilerplate to be usable inside SwiftUI.
 
 Apollo Client iOS could assist the developer by adding conformance to the Identifiable protocol to its generated models. Selecting the field to be used as an identity is done by adding a new GraphQL directive.
 
+A concept of identity is required to allow response objects to be cached. The various Apollo Client projects have mechanisms that allow for identifiers to be selected through additional code. The new directive could allow schema authors to assist client authors with caching.
```
calvincestari, line 11: > The new directive could allow schema authors to assist client authors with caching.

This is a nice ideal but not likely, as you noted below in [Usage in operations](https://github.com/apollographql/apollo-ios-dev/pull/549/commits/6430f37b1cc2931dd5236c4087a78e31a3ecffb8#diff-acc259b506e96644229b6d8785740a9248ed3849714b716e6b354258fe712004R63).
* Federated schema authors already have the [`@key` directive](https://www.apollographql.com/docs/graphos/reference/federation/directives#key) but I don't think those are passed through in the API schema SDL, nor introspection.
* You'll end up with a string of different directives if there isn't uniformity across all the clients.

apollo-ios/Design/swift-identifiable.md
```diff
@@ -1,29 +1,47 @@
...
+
 # Definition
 
 The directive is defined as:
 ```graphql
-directive @apollo_client_ios_identity on FIELD | FIELD_DEFINITION
+directive @identity(scope: IdentityScope = SELECTION) on FIELD | FIELD_DEFINITION
+
+enum IdentityScope {
```
calvincestari, line 19: This is a novel idea but you can probably derive the scope from the context in which the directive is used; interface field vs. type field vs. selection field.

This may also introduce a bunch of validation required to make sure it isn't used with excessive scope, i.e.: should a selection field be able to define global scope? _Probably not._

apollo-ios/Design/swift-identifiable.md
```diff
@@ -77,18 +95,32 @@ The protocol requires that the identity is accessible through a public field nam
...
+## Apollo Kotlin's @typePolicy directive
+
+Apollo Kotlin has [custom directives](https://www.apollographql.com/docs/kotlin/caching/declarative-ids) that allows for client authors to specify the caching key though pure GraphQL:
+
+```graphql
+extend type Book @typePolicy(keyFields: "id")
+```
 
-The directive could be expanded to declare the scope of an identity (e.g. unique across the entire API or only one specific Database table). This makes the directive beneficial outside of Swift code generation, but it doesn't change the behavior of the Identifiable protocol. The [Swift documentation](https://developer.apple.com/documentation/swift/identifiable) states that the scope and duration of an identity is unspecified, so modifying the scope would not lead to a difference in the generated code.
+This could be used for Identifiable conformance if a single field is provided, and caching behavior could also be matched with Apollo Kotlin. A disadvantage is that this directive is only used inside type extensions, and can't be used to mark fields directly in a query.
```
calvincestari, line 126: If `@typePolicy` was used could the Identifiable `id` value be a hash value of all specified key fields? I don't have an answer for not being able to use `@typePolicy` in operations though.

calvincestari: If this were possible then maybe `Identifiable` conformance could just be a codegen configuration option (`generateIdentifiableConformance: Bool`) and the hashed type policy key values are used for `id` value?

Not great that there would be two places for configuration but it does highlight that `Identifiable` is a side-effect of object identity. Maybe that's not a bad requirement, plus it wouldn't need to be used in operations then?

apollo-ios/Design/swift-identifiable.md
x-sheep: I agree this document is coming close to a [xkcd 927 situation](https://xkcd.com/927/). I wasn't aware of the `@key` directive, but it not being revealed to the clients is a deal-breaker for my use case.

apollo-ios/Design/swift-identifiable.md
x-sheep: > could the Identifiable id value be a hash value of all specified key fields?

You're right, I forgot that a tuple of `Hashable` objects is also `Hashable` itself, which makes it valid for use with `Identifiable`.

apollo-ios/Design/swift-identifiable.md
x-sheep: I changed my mind regarding the SELECTION scope. Its application is too limited (protocol conformance only), so I took it out. I think removing this case also removes the reason for clients to widen the scope themselves.

apollo-ios/Design/swift-identifiable.md
x-sheep: > If this were possible then maybe Identifiable conformance could just be a codegen configuration option (generateIdentifiableConformance: Bool)

When the identity is established as a single field named `id`, there is no drawback to always adding Identifiable conformance.

If a custom getter for `id` has to be added for conformance, then following a configuration option seems like a good idea. This would also make it less surprising when Codegen suddenly emits a warning about a getter being unable to be generated because of naming conflicts.

calvincestari: > When the identity is established as a single field named id, there is no drawback to always adding Identifiable conformance.

Even if it is a single field named `id` I think we'd still prefer configuration over convention. That configuration option can have a default of `true` if not supplied explicitly = less auto-magic behaviour.

calvincestari:
@AnthonyMDev - it would be a good idea for you to catch up on this conversation. Would appreciate your opinion here as I know you have thoughts on type policy, caching, etc.

apollo-ios/Design/swift-identifiable.md
AnthonyMDev: We have not previously defined any directives that are exposed as part of the public schema and are intended for consumption by the Apollo client's specifically. While this is certainly possible, we need to consider all of the issues this can cause. These directives would need to be namespaced accordingly. It is very possible that some schema's out there already have an `@identity` directive that they are using for some other use case. We can't just assume that directive is what we are looking for and consume it on the client for this usage.

Namespacing of directives doesn't really have great support in the GraphQL spec currently. We'd need to use some ugly directive name like `@apollo_client_identity`. I know that an [`@core` directive ](https://github.com/apollographql/specs/blob/main/core/v0.2/core-v0.2.md) was discussed by us at some point to help with this, but it's not an accepted and used feature currently. @martinbonnin knows more about that than I do.

apollo-ios/Design/swift-identifiable.md
```diff
@@ -0,0 +1,124 @@
...
+
+This directive MUST be used on a field with a non-nullable scalar type. Other types are not supported.
+
+## Scope parameter
+
+The directive can have a parameter indicating the scope of the identity. They are ordered from _narrowest_ to _widest_:
+
+1. `TYPE`: The identifier is unique for the type, e.g. an auto-incrementing column from a database table.
+2. `SERVICE`: The identifier is unique across the current GraphQL Service.
+3. `GLOBAL`: The identifier is unique across all GraphQL services. This can be used by identifiers generated according to [RFC 4122](https://datatracker.ietf.org/doc/html/rfc4122) or [RFC 9562](https://datatracker.ietf.org/doc/html/rfc9562) (also known as Universally Unique IDentifiers or UUIDs).
```
AnthonyMDev, line 34: How can you guarantee that this is unique across all GraphQL services globally? Are we just assuming you won't likely ever hit a conflicting ID if you are using UUIDs? In the future of a global graph using the [`@link` directive](https://github.com/apollographql/specs/blob/main/link/v1.0/link-v1.0.md) (which is only a concept for one possible future right now), there will likely need to be a way to ensure global uniqueness, but I have a suspicion that it will end up being much more complex than this.

apollo-ios/Design/swift-identifiable.md
```diff
@@ -0,0 +1,124 @@
...
+
+It's possible for interfaces to use the directive on a field.
+
+```graphql
+interface Identifiable {
+	id: ID! @identity(scope: TYPE)
+}
+```
+
+Implementations of this interface MUST copy the directive to the same field. The scope argument in the implementation MUST NOT be narrower than the scope in the interface, but it may be wider.
```
AnthonyMDev, line 59: I'm not sure of how else to handle this, but I'm concerned this may be too constraining for many real life applications. What if you have a field of an interface type and you want to conform to that interface by multiple different types that have different fields used for identity?

Maybe that isn't too much of a concern? But it seems to me that what we really want is some way of saying "all types that implement this interface _must have_ a field marked `@identity`. Not "all types that implement this interface must have a field with _this specific name and type_ marked `@identity`.

I'm also not clear how this would work with fields of a union type. There was once an [RFC to allow unions to require an interface implementation](https://github.com/graphql/graphql-spec/pull/939), but it's stalled out as of now.

apollo-ios/Design/swift-identifiable.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Usage in operations
+
+It's likely that an external schema will not use this directive. In this case, an identity MAY be chosen when writing a query. The directive SHOULD NOT be included in the query sent to the GraphQL server.
+
+```graphql
+query GetAllAnimals {
+	allAnimals {
+		id @identity
```
AnthonyMDev, line 68: In operation usage, do you not need to provide a `scope`? If so, then I think there needs to be two different directives defined for `FIELD` and `FIELD_DEFINITION`.

apollo-ios/Design/swift-identifiable.md
```diff
@@ -0,0 +1,124 @@
...
+		string
+	}
+}
+```
+
+It is allowed for a query and a type to both use the directive to describe the same field. If the client and schema use different scopes for the same field, the scope defined by the schema is used. Clients SHOULD NOT define a wider scope than declared by the schema. Codegen should emit a warning in this case.
```
AnthonyMDev, line 74: I'm unclear why the client would ever need to define any scope here. When using `@identity` in a operation, you aren't going to be using these keys for cache normalization, since other selections sets may not provide an `@identity` field, or could even provide a _different_ field.

apollo-ios/Design/swift-identifiable.md
```diff
@@ -0,0 +1,124 @@
...
+
+If the server defines a field as an identity, a query SHOULD NOT choose another field. A query MUST NOT use the directive on more than one field in the same selection (unless they are in differently nested objects).
```
AnthonyMDev, line 76: While a directive provided by schema authors is likely to be helpful in many cases. I also have a concern that it won't be 100% correct for all client use cases. There would need to be a mechanism for client developers to easily override that behavior.

apollo-ios/Design/swift-identifiable.md
```diff
@@ -0,0 +1,124 @@
...
+
+```swift
+public var id: String { self.uuid }
+```
+
+### Naming conflicts
+
+If the identity field is not called `id`, but another field called `id` is present in the selection, a custom getter cannot be added. Swift does not support using another field to handle the conformance.
+
+In this case, a conformance to Identifiable SHOULD NOT be generated. Codegen should emit a warning without stopping the generation process.
```
AnthonyMDev, line 100: This is a definite, unavoidable edge case. It's not ideal, but I suppose just emitting a warning and not generating the conformance is the proper behavior here.

apollo-ios/Design/swift-identifiable.md
```diff
@@ -0,0 +1,124 @@
...
+
+## Apollo Kotlin's @typePolicy directive
+
+Apollo Kotlin has [custom directives](https://www.apollographql.com/docs/kotlin/caching/declarative-ids) that allows for client authors to specify the caching key through pure GraphQL:
+
+```graphql
+extend type Book @typePolicy(keyFields: "id")
+```
+
+This could be used for Identifiable conformance by generating a getter which returns a tuple of all listed key fields, and caching behavior could also be matched with Apollo Kotlin. A disadvantage is that this directive is only used inside type extensions, and can't be used to mark fields directly in a query.
```
AnthonyMDev, line 124: I actually think that is a more principled approach, and leads to far less awkward and confusing interactions. In a GraphQL client's normalized cache, you really want identity to be consistent across your operations for proper normalization.

I think the use case where you want to use a field for identity in only a single operation is pretty narrow and probably an anti-pattern.

apollo-ios/Design/swift-identifiable.md
```diff
@@ -0,0 +1,124 @@
...
+
+In this case, a conformance to Identifiable SHOULD NOT be generated. Codegen should emit a warning without stopping the generation process.
+
+## Caching behavior
+
+If no `@identity` directive is present, clients MAY use other mechanisms to determine if and how to cache the object.
+
+An identifier with scope `TYPE` can be combined with the `__typename` field to generate a caching key that's unique for the GraphQL Service.
+
+Identifiers with a scope of `SERVICE` or `GLOBAL` can be directly used as a caching key.
```
AnthonyMDev, line 108: This makes sense, but honestly, we may continue to combine with the `__typename` even when the scope is `SERVICE` or `GLOBAL`, just for consistency sake.

AnthonyMDev (commented):
Thank you so much @x-sheep for putting together this detailed and well-formed RFC. I definitely have a lot of questions and concerns about this approach, but I appreciate the time and thought that has gone into it thus far!

I do think that the `@typePolicy` alternative from Apollo Kotlin is a better approach. It allows for multiple fields to be used together to create identity. And as I commented inline, I think that it's better to not be able to configure identity on a per-selection-set basis, but leave it to types only.

`@typePolicy` implementation has been on our Roadmap for quite some time, but other things have taken priority. However, using the `@typePolicy` to conform to `Identifiable` in `Swift` was not on our radar. You've convinced me that this should be part of that work.

apollo-ios/Design/swift-identifiable.md
x-sheep: Omitting the parameter defaults to the narrowest scope.

apollo-ios/Design/swift-identifiable.md
x-sheep: There's no guarantee beyond what the RFC provides (where the chance of collision is near-zero). I've specifically worded it as "generated according to the spec" instead of only _looking_ like 128 bits.

It's still allowed to append more text to an identifier of this type (as long as the UUID portion itself is unchanged) to make the risk of collisions even smaller.

x-sheep:
Thanks for the feedback, everyone.

I think the document as I wrote it is somewhere between "high level directive that should maybe be part of the GraphQL spec itself" and "client-specific details". If this were to become an RFC for GraphQL itself, I would remove the entire portion that allows the directive to be used on queries or fragments and focus only on schema authoring. This would make it a replacement for Apollo Federation's `@key`, except it would be accessible through introspection similar to the `@deprecated` directive.

If schemas were not changed, I now think that `@typePolicy` is the best option for Apollo iOS. While I'm not a fan of specifying a string of field names, it's the only way to declare it through a type extension (which is the best for the DRY principe).

My current idea is to emit an `id` field when Identifiable conformance is desired, and have the return type be a single field or a tuple of fields. This field can be called a lot of times per second by SwiftUI, so string concatenation in every call isn't desirable. For caching, a new field `__cacheKey` could be added to all types, which is either nil or a `CacheKeyInfo` object. If this object was stored inside the SelectionSet struct instead of calculated every time, it could also be used as the `id` instead of a tuple.

calvincestari:
> I think the document as I wrote it is somewhere between "high level directive that should maybe be part of the GraphQL spec itself" and "client-specific details". If this were to become an RFC for GraphQL itself, I would remove the entire portion that allows the directive to be used on queries or fragments and focus only on schema authoring. This would make it a replacement for Apollo Federation's `@key`, except it would be accessible through introspection similar to the `@deprecated` directive.

If you look at the wider GraphQL ecosystem these things are already in-play to make them part of 'GraphQL':
1. The [Composite Schemas Working Group](https://github.com/graphql/composite-schemas-wg?tab=readme-ov-file) (aka. GraphQL Federation) is working to turn Apollo Federation into a formalized spec for all of GraphQL. It carries over the [`@key` (and other) directives](https://github.com/graphql/composite-schemas-spec/blob/main/spec/Section%202%20--%20Source%20Schema.md#key). _I'm unsure whether this makes them able to be specified in subgraphs though._
2. The _very_ long standing [issue #300](https://github.com/graphql/graphql-spec/issues/300) in the GraphQL spec to make user-defined meta-data directives available through introspection.

> If schemas were not changed, I now think that `@typePolicy` is the best option for Apollo iOS. While I'm not a fan of specifying a string of field names, it's the only way to declare it through a type extension (which is the best for the DRY principle).

I certainly think this is the correct way forward.

AnthonyMDev:
Closing this, as we are merging and releasing `@typePolicy` shortly!

PR #557, opened 2024-12-16
fix: `DataDict` initialization of `deferredFragments`

calvincestari:
While working on selection set initializers I realized that the current generated initializers do not set the selection set types in the `deferredFragments` property.

Some of the named fragment initializers do not look correct but I believe this is due to https://github.com/apollographql/apollo-ios/issues/3486 and https://github.com/apollographql/apollo-ios/issues/3487. Adding the values to `deferredFragment` will not have any detrimental effect and those will get fixed on top of this work.

### DIFF
--- a/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
+++ b/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
@@ -604,6 +604,9 @@ struct SelectionSetTemplate {
   private func InitializerTemplate(
     _ selectionSet: ComputedSelectionSet
   ) -> TemplateString {
+    let containsDeferredFragment = (selectionSet.direct?.inlineFragments.containsDeferredFragment ?? false) ||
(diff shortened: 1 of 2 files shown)


apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/SelectionSetTemplate.swift
```diff
@@ -708,9 +718,41 @@ struct SelectionSetTemplate {
...
+  }
+
+  private func InitializerDeferredFragments(
+    _ selectionSet: ComputedSelectionSet
+  ) -> TemplateString? {
+    guard let directSelections = selectionSet.direct else { return nil }
+
+    var deferredFragments: OrderedSet<String> = []
+
+    let nameGenerator: (_ typeInfo: SelectionSet.TypeInfo) -> String = { typeInfo in
```
AnthonyMDev, line 732: Love this pattern. Very clean!

AnthonyMDev (approved):
LGTM. Great fix!

PR #558, opened 2024-12-16
feature: Adds support for GraphQL over HTTP media type

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3485.

1. Make sure we include the new type in the 'accept' request header.
2. Update tests for new content type.
3. There were no changes needed for error handling because [we do not throw when the response status code falls outside the 200..<300 range](https://github.com/apollographql/apollo-ios-dev/blob/main/apollo-ios/Sources/Apollo/ResponseCodeInterceptor.swift#L60).
4. Our non-multipart response parser does not require that the content type actually be `application/json`. Instead it simply attempts to [deserialize anything as if it is JSON](https://github.com/apollographql/apollo-ios-dev/blob/main/apollo-ios/Sources/Apollo/JSONResponseParsingInterceptor.swift#L54). It's only in the multipart parsers that we do a content type check for the chunks. @AnthonyMDev - this hasn't been an issue for now but do you think we should add this check? I'm inclined to leave the standard parser as it is.

### DIFF
--- a/Tests/ApolloTests/RequestChainTests.swift
+++ b/Tests/ApolloTests/RequestChainTests.swift
@@ -1215,4 +1215,193 @@ class RequestChainTests: XCTestCase {
     // then
     wait(for: expectations, timeout: 1, enforceOrder: true)
   }
+
+  // MARK: Response Tests
+
+  func test__response__givenUnsuccessfulStatusCode_shouldFail() throws {
+    // given
+    let client = MockURLSessionClient(
+      response: .mock(
+        url: TestURL.mockServer.url,
+        statusCode: 500,
+        httpVersion: nil,
+        headerFields: nil
+      ),
+      data: """
+      {
+        "data": {
+          "__typename": "Hero",
+          "name": "R2-D2"
+        }
+      }
+      """.data(using: .utf8)
+    )
+
+    let provider = DefaultInterceptorProvider(
+      client: client,
+      store: ApolloStore()
+    )
+
+    let transport = RequestChainNetworkTransport(
+      interceptorProvider: provider,
+      endpointURL: TestURL.mockServer.url
+    )
+
+    let expectation = expectation(description: "Response received")
+
+    _ = transport.send(operation: MockQuery<Hero>()) { result in
+      switch result {
+      case .success:
+        XCTFail("Unexpected response: \(result)")
+
+      case .failure:
+        expectation.fulfill()
+      }
+    }
+
+    wait(for: [expectation], timeout: 1)
+  }
+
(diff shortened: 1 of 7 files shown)


calvincestari:
I'm going to add a couple tests of our request chain to ensure that the assumption we make on content type holds true.

Tests/ApolloTests/RequestChainTests.swift
```diff
@@ -1215,4 +1215,193 @@ class RequestChainTests: XCTestCase {
...
+        expectation.fulfill()
+      }
+    }
+
+    wait(for: [expectation], timeout: 1)
+  }
+
+  // This test is odd because you might assume it would fail but there is no content-type checking on standard
+  // GraphQL response parsing. So this test is here to ensure that existing behaviour does not change.
+  func test__response__givenUnknownContentType_shouldNotFail() throws {
```
calvincestari, line 1267: I've added tests to keep the existing behaviour of standard GraphQL response parsing including this odd one.

apollo-ios/Sources/Apollo/MultipartResponseDeferParser.swift
```diff
@@ -59,7 +59,7 @@ struct MultipartResponseDeferParser: MultipartResponseSpecificationParser {
     for dataLine in chunk.components(separatedBy: Self.dataLineSeparator.description) {
       switch DataLine(dataLine.trimmingCharacters(in: .newlines)) {
       case let .contentHeader(type):
-        guard type == "application/json" else {
+        guard type == "application/graphql-response+json" || type == "application/json" else {
```
AnthonyMDev, line 62: We should create a helper method for this. We're doing string comparisons against strings that are really constants. And we're doing the same comparison in two places. This is prone to future bugs.

 We should store the constants as `StaticString`s and have a helper method to check the content header.

AnthonyMDev (commented):
Tests all look good. I just thing that we should clean the repeated string constants up with a helper method.

apollo-ios/Sources/Apollo/MultipartResponseDeferParser.swift
calvincestari: I'll put this PR back into draft and make that change when I get back.

calvincestari: This is done in 179d0043f690ae8c215e9b66d55306bfa68c127a.

calvincestari:
@AnthonyMDev - this PR is ready for another review, I've addressed the previous comment.

AnthonyMDev (approved):

PR #562, opened 2024-12-29
[Codegen]: Partial fragments

igor-florescu-ck:
We are currently migrating to the new Apollo codegen.

The new codegen tool runs for over 30 minutes and uses a lot of memory to generate the latest Swift types, and our CI is failing to complete. I suspect this thread might be related to https://github.com/apollographql/apollo-ios/issues/3434.

I am opening this pull request to propose changes and gather feedback from the repo authors.

**The current flow:**
Whenever the [RootFieldBuilder](https://github.com/apollographql/apollo-ios-codegen/blob/6db35551c7bc0879b61539c08338253a7b4a146b/Sources/IR/IR%2BRootFieldBuilder.swift#L6) builds named fragments via [buildNamedFragmentSpread](https://github.com/apollographql/apollo-ios-codegen/blob/6db35551c7bc0879b61539c08338253a7b4a146b/Sources/IR/IR%2BRootFieldBuilder.swift#L401). The codegen flow will eventually reuse/create a previously built fragment from [BuiltFragmentStorage](https://github.com/apollographql/apollo-ios-codegen/blob/6db35551c7bc0879b61539c08338253a7b4a146b/Sources/IR/IRBuilder.swift#L53). Once resolved, a [NamedFragment](https://github.com/apollographql/apollo-ios-codegen/blob/6db35551c7bc0879b61539c08338253a7b4a146b/Sources/IR/IR%2BNamedFragment.swift#L4) type would be returned by the codegen flow and will merge the [Entity](https://github.com/apollographql/apollo-ios-codegen/blob/6db35551c7bc0879b61539c08338253a7b4a146b/Sources/IR/IR%2BEntity.swift#L8) fields to build up the [EntitySelectionTree](https://github.com/apollographql/apollo-ios-codegen/blob/6db35551c7bc0879b61539c08338253a7b4a146b/Sources/IR/IR%2BEntitySelectionTree.swift#L15) in the Selection Set which references the named fragment.

Blank diagram (7)

The fragments resolved from [BuiltFragmentStorage](https://github.com/apollographql/apollo-ios-codegen/blob/6db35551c7bc0879b61539c08338253a7b4a146b/Sources/IR/IRBuilder.swift#L53) are eventually referenced to it's previously created swift file.

Some fragments can contain very large amounts of [Entity](https://github.com/apollographql/apollo-ios-codegen/blob/6db35551c7bc0879b61539c08338253a7b4a146b/Sources/IR/IR%2BEntity.swift#L8) fields which might not be relevant for referencing intents.

With this pull request, I suggest adding support for partial [NamedFragment](https://github.com/apollographql/apollo-ios-codegen/blob/6db35551c7bc0879b61539c08338253a7b4a146b/Sources/IR/IR%2BNamedFragment.swift#L4) to **reduce the amount of data merged into the containing EntitySelectionTree**.

The intention of partial fragments is to return a [NamedFragment](https://github.com/apollographql/apollo-ios-codegen/blob/6db35551c7bc0879b61539c08338253a7b4a146b/Sources/IR/IR%2BNamedFragment.swift#L4) which contain entities necessary for **referencing** intents. (ex: field for inlineFragments, references for **referencedFragments** etc.)

Copy of Blank diagram

In this example, I am naively reducing the number of entities that have FieldPaths deeper than 3 (?)
I was also considering an experimental flag along to the config, however I've limited the amount of changes to gather initial feedback.

**Why 3**? Assumption: entities containing the fields to be spread into the containing inline fragment + underlying fragments and their fields. (maybe rather than naively filtering a depth of 3 a better approach would be to filter by definition/GraphQLType)

Blank diagram (8) (1)
I'm not however familiar with all edge cases and would appreciate any suggestions and reviews.

With this change, the same output is generated against our schema, but the duration is reduced from 30 minutes to 3 minutes, and memory usage has decreased to 400 MB.

### DIFF
--- a/apollo-ios-codegen/Sources/IR/IR+DefinitionEntityStorage.swift
+++ b/apollo-ios-codegen/Sources/IR/IR+DefinitionEntityStorage.swift
@@ -11,6 +11,14 @@ public class DefinitionEntityStorage {
     self.entitiesForFields[rootEntity.location] = rootEntity
   }
 
+  private init(
+    sourceDefinition: Entity.Location.SourceDefinition,
+    entitiesForFields: [Entity.Location: Entity]
+  ) {
+    self.sourceDefinition = sourceDefinition
+    self.entitiesForFields = entitiesForFields
+  }
+
   func entity(
     for field: CompilationResult.Field,
     on enclosingEntity: Entity
@@ -66,5 +74,37 @@ public class DefinitionEntityStorage {
     entitiesForFields[location] = entity
     return entity
   }
-
 }
+
+extension DefinitionEntityStorage {
+  func partial() -> DefinitionEntityStorage {
+    return DefinitionEntityStorage(
+      sourceDefinition: sourceDefinition,
+      entitiesForFields: makePartialEntitiesForFields()
+    )
+  }
+
+  private func makePartialEntitiesForFields() -> [Entity.Location: Entity] {
+    var partialEntitiesForFields: [Entity.Location: Entity] = [:]
+    for (location, entity) in entitiesForFields {
+      /*
+       Based on my understanding:
+       Fragment rendering points to these entities.
+       Whenever we try to render a top-level fragment spread, it resolves the fragment from the entity storage.
+       In theory, the generation flow needs the types and entities to spread, fragments, and their underlying entities,
+       hence the hardcoded 'count < 4'.
+
+       Possibly, could iterate through the field path and exclude irrelevant entities rather than a hardcoded value?
+
+       */
+      if let count = location.fieldPath?.count, count < 4 {
+        partialEntitiesForFields[location] = entity
+      }
+      if location.fieldPath == nil {
+        partialEntitiesForFields[location] = entity
+      }
+    }
+    return partialEntitiesForFields
+  }
+}
+
--- a/apollo-ios-codegen/Sources/IR/IR+NamedFragment.swift
+++ b/apollo-ios-codegen/Sources/IR/IR+NamedFragment.swift
@@ -51,3 +51,15 @@ public class NamedFragment: Definition, Hashable, CustomDebugStringConvertible {
     definition.debugDescription
   }
 }
+
+public extension NamedFragment {
+  func partialFragment() -> NamedFragment {
+    NamedFragment(
+      definition: definition,
+      rootField: rootField,
+      referencedFragments: referencedFragments,
+      entityStorage: entityStorage.partial(),
+      containsDeferredFragment: containsDeferredFragment
+    )
+  }
+}
--- a/apollo-ios-codegen/Sources/IR/IR+RootFieldBuilder.swift
+++ b/apollo-ios-codegen/Sources/IR/IR+RootFieldBuilder.swift
@@ -398,7 +398,7 @@ class RootFieldBuilder {
     spreadIntoParentWithTypePath parentTypeInfo: SelectionSet.TypeInfo,
     deferCondition: CompilationResult.DeferCondition? = nil
   ) async -> NamedFragmentSpread {
-    let fragment = await ir.build(fragment: fragmentSpread.fragment)
+    let fragment = await ir.build(fragment: fragmentSpread.fragment, shouldGeneratePartialFragment: true)
     referencedFragments.append(fragment)
     referencedFragments.append(contentsOf: fragment.referencedFragments)
 
@@ -410,7 +410,7 @@ class RootFieldBuilder {
 
     self.containsDeferredFragment = fragment.containsDeferredFragment || scope.deferCondition != nil
 
-    let scopePath = scope.isEmpty 
+    let scopePath = scope.isEmpty
       ? parentTypeInfo.scopePath
       : parentTypeInfo.scopePath.mutatingLast { $0.appending(scope) }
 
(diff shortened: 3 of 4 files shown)


AnthonyMDev:
I'm looking at this now, but I'm having a hard time wrapping my head around how this is even passing tests right now. I would assume that cutting off the depth at 3 would make some merged fields be missing. I also haven't worked in this code in a while, and it's quite complex, so I need to regain some context. I'll keep looking into it.

Can you expand on what this means:

> In theory, the generation flow needs the types and entities to spread, fragments, and their underlying entities, hence the hardcoded 'count < 4'.

I feel like I'm missing something about why this is adequate. My hunch is that we just don't have any unit tests that actually exceed a depth of 3.

AnthonyMDev:
I've added a new unit test in PR #570 that merges entities from a fragment with a depth of 4. If you rebase this PR on `main` now, that test will fail. This is the simplest case in which the partial fragments with a depth of 3 do not work.

While I agree that for almost all cases, having a depth of 3 (or 4 for that matter) will create models that serve their purpose. The issue is that because they are incomplete, selection set initializers will be incorrect. Missing the property accessors on the merged models, while not completely and exhaustively correct, is not a huge issue. Users could access the fragment models to access those fields. 

But we use the calculated merged selections to generate the initializers for the models. This means those initializers will be missing fields. If you initialize the models with missing fields and then access those fields by accessing the named fragments that do contain them, you will get a crash, additionally, trying to write the incomplete models to the cache will cause a crash.

Any shortcuts like this that actually make the models not exhaustively correct is going to have that problem. It's actually why the experimental feature for turning off fragment field merging is incompatible with selection set initializers. If you try to enable both features, you get an error.

---

The fact that you report getting the same output with this partial fragment workaround implemented makes me assume that you are already using the config option to disable fragment field merging and thus, are not using selection set initializers. Is that assumption correct?

Currently, disabling field merging only shortcuts the traversal of the merged selection set trees when calculating the merged selections. I wonder if we could add some logic to prevent the `mergeAllSelectionsIntoEntitySelectionTrees` from being called in the first place if named fragment field merging is disabled. That may be able to give you the performance improvements you're looking for. I'm not sure if that would cause other edge cases to arise.

AnthonyMDev:
I've implemented the idea that I had in my last comment, and I believe it works!!! Please check out this PR: https://github.com/apollographql/apollo-ios-dev/pull/571 and tell me if it runs for you.

This only works when fragment field merging is disabled, but as far as I can tell it's working. I'm concerned there might be some edge cases, but I can't come up with any reason why there would be, so I'm hoping this solves it!

igor-florescu-ck:
Hi @AnthonyMDev, 

Thanks for getting back on this. I indeed had initializers disabled, but field merging was enabled. This workaround carried the Entities for those fragments to be merged.

I first started looking at mergeAllSelectionsIntoEntitySelectionTrees; however, this partial approach seemed easier for me to move forward at the time.

I pulled your changed from fragment-field-merging-disabled-performance and it's significantly faster and memory usage manageble.

I will try run it against our project tomorrow and let you know.

igor-florescu-ck:
> I'm looking at this now, but I'm having a hard time wrapping my head around how this is even passing tests right now. I would assume that cutting off the depth at 3 would make some merged fields be missing. I also haven't worked in this code in a while, and it's quite complex, so I need to regain some context. I'll keep looking into it.
> 
> Can you expand on what this means:
> 
> > In theory, the generation flow needs the types and entities to spread, fragments, and their underlying entities, hence the hardcoded 'count < 4'.
> 
> I feel like I'm missing something about why this is adequate. My hunch is that we just don't have any unit tests that actually exceed a depth of 3.

@AnthonyMDev sorry for the confusion... this was never meant to be hardcoded value (the hardcoded value was for review purposes)

The idea here was, to iterate through the Field Path and leave the necessary Entities for:
1. Field merging
2. Initializers
3. Type references
4. Nested fragments

As opposed to https://github.com/apollographql/apollo-ios-dev/pull/571 I was hoping to leave field merging enabled.

AnthonyMDev:
Ah, I see. I'm not sure what "unnecessary entities" you expect to filter out though? If you have fragment field merging off, then you would filter out the same things that PR #571 prevents from being merged into the trees in the first place. But if you have field merging enabled, I don't think there are any selections you can filter out and still have safe and fully correct models.

Either way, if you are filtering anything out at all, initializers are going to be a no-go. Selection set initializers require that _all_ fields are calculated and merged to ensure that the initializer has all of the parameters needed to successfully create the model. If you filter a single selection out, you now have initializers that will crash on access or throw errors when written to the normalized cache.

igor-florescu-ck:
> Hi @AnthonyMDev,
> 
> Thanks for getting back on this. I indeed had initializers disabled, but field merging was enabled. This workaround carried the Entities for those fragments to be merged.
> 
> I first started looking at **mergeAllSelectionsIntoEntitySelectionTrees**; however, this partial approach seemed easier for me to move forward at the time.
> 
> I pulled your changed from **fragment-field-merging-disabled-performance** and it's significantly faster and memory usage manageble.
> 
> I will try run it against our project and let you know.

Hi @AnthonyMDev,
I've tried https://github.com/apollographql/apollo-ios-dev/pull/571 and looks good so far with our project with field merging disabled.

Thanks for the follow up.
It prevents merging entities and speeds up the tool considerably.
Closing this PR in favour of: https://github.com/apollographql/apollo-ios-dev/pull/571

PR #570, opened 2025-01-06
Add test for merging a fragment with a depth of 4

AnthonyMDev:
PR #562 highlighted that we are not testing fragment field merging past a depth of 3 entities. While we cannot test for every depth to infinity, this test confirms merging works to a depth of 4 and proves the edge case of that PR.

### DIFF
--- a/Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
+++ b/Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
@@ -4090,6 +4090,151 @@ class IRRootFieldBuilderTests: XCTestCase {
       .to(shallowlyMatch(predator_asWarmBlooded_height_expected))
   }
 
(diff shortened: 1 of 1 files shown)


Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
```diff
@@ -4090,6 +4090,151 @@ class IRRootFieldBuilderTests: XCTestCase {
...
+          }
+        }
+      }
+    }    
+    """
+
+    // when
+    try await buildSubjectRootField()
+
+    let Object_Root = try XCTUnwrap(schema[object: "Root"])
```
calvincestari, line 4155: There is a warning that this variable is never used - is that intentional or highlighting something missed in the test?

calvincestari (commented):
One question about the var assignment warning.

Tests/ApolloCodegenTests/CodeGenIR/IRRootFieldBuilderTests.swift
AnthonyMDev: Oh you're right. For some reason the warning did not show up for me in Xcode? Fixed!

calvincestari (approved):

PR #571, opened 2025-01-06
Prevent merging of fragment fields into selection tree if fragment field merging is disabled

AnthonyMDev:
Currently, when fragment field merging is disabled we are preventing the fields from being merged into the `MergedSelections` while traversing the built `EntitySelectionTree`. However, the performance issues described in https://github.com/apollographql/apollo-ios/issues/3434 are actually caused during the merging of the fragment fields _into_ the `EntitySelectionTree`s during the building of the operations.

This PR prevents the fragment selection trees from being merged into the `EntitySelectionTree` while building the operations **if fragment field merging is disabled**.

This means that the entity trees are actually incomplete, but they still work for the purposes of Codegen in this scenario. Incomplete `EntitySelectionTree`s would cause selection set initializers to be invalid. But this is not a problem because disabling fragment field merging is already incompatible with selection set initializers (and we have a validation step to ensure those options are not used together).

### DIFF
--- a/Tests/ApolloCodegenInternalTestHelpers/IRBuilderTestWrapper.swift
+++ b/Tests/ApolloCodegenInternalTestHelpers/IRBuilderTestWrapper.swift
@@ -24,7 +24,10 @@ public class IRBuilderTestWrapper {
     operation operationDefinition: CompilationResult.OperationDefinition,
     mergingStrategy: MergedSelections.MergingStrategy = .all
   ) async -> IRTestWrapper<IR.Operation> {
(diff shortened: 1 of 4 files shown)


pm-dev:
Although this does allow codegen to complete for me, I'm not sure if this is a good idea:
https://github.com/apollographql/apollo-ios/issues/3496

AnthonyMDev:
I'm not sure that issue is related to this feature. It could be, but I'd like to get more information from you in that thread @pm-dev.

BobaFetters (approved):

PR #572, opened 2025-01-08
fix: Multipart chunk content type

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3495.

Refactored the parsing of `content-type` to handle multiple directives, which resolved the bug but also removed duplicated code. and allows for all directives to be used.

### DIFF
--- a/Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
+++ b/Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
@@ -438,6 +438,47 @@ final class MultipartResponseSubscriptionParserTests: XCTestCase {
     let network = buildNetworkTransport(responseData: """
       
       --graphql
+      content-type: application/json
+
+      {
+        "payload": {
+          "data": {
+            "__typename": "Time",
+            "ticker": 1
+          }
+        }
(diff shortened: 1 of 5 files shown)


Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
```diff
@@ -438,7 +438,7 @@ final class MultipartResponseSubscriptionParserTests: XCTestCase {
     let network = buildNetworkTransport(responseData: """
       
       --graphql
-      content-type: application/json
```
AnthonyMDev, line 441: Shouldn't we keep a test that it works when you just have `application/json` and not the additional directive?

AnthonyMDev (approved):

Tests/ApolloTests/Interceptors/MultipartResponseSubscriptionParserTests.swift
calvincestari: We do have those. They're further down in other tests within this file and `MultipartResponseDeferParserTests.swift`. They aren't specifically labelled as content type tests though; that felt superfluous. If you want I can add them though.

PR #578, opened 2025-01-14
fix: WebSocket data race crash

BobaFetters:
Fixing a crash in websockets that can occur because of a data race when multiple successive disconnect/connect calls are made.

Closes apollographql/apollo-ios#3390


apollo-ios/Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
```diff
@@ -10,6 +10,7 @@
 
 import Foundation
 import CommonCrypto
+import Apollo
```
calvincestari, line 13: I don't think we need this anymore; it's a remnant of `@Atomic` approaches in this file too.

calvincestari (approved):
Just that one import statement to remove otherwise looks good. Nice work, thanks @BobaFetters!

apollo-ios/Sources/ApolloWebSocket/DefaultImplementation/WebSocket.swift
BobaFetters: Good catch

PR #580, opened 2025-01-16
feature: Add suffix to schema type filenames

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/2598.

Users now have the codegen configuration option (`appendSchemaTypeFilenameSuffix`) to have a suffix added to their schema type filenames. This will help resolve problems in Xcode where the name of a schema type and operation type are the same causing a build error.

This new config option has a default value of `false` which means this should have no effect on users without the filename conflict error.


Sources/AnimalKingdomAPI/AnimalKingdomAPI/Sources/Schema/CustomScalars/CustomDate.scalar.swift
```diff
@@ -0,0 +1,10 @@
+// @generated
```
AnthonyMDev, line 1: The old versions of these scalar files need to be deleted. But this actually surfaces a bigger issue here.

If a user had this new config `false` and then sets it to true, previously generated custom scalars (and any other  files that are marked to not be overwritten) will not be handled properly. We need to preserve the changes to the code the user may have made. So we actually need to detect the old file name and _rename_ the existing file, rather than just writing a new file.

Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
AnthonyMDev, line 1: I think we need tests that validates that when this is set to `true` that is reads it and writes it to the config properly.

AnthonyMDev (changes requested):
Just one issue to be resolved with the way the editable generated files are handled.

We also need to add this new option to the documentation (not just the generated docs).

calvincestari:
This PR already [updated](https://github.com/apollographql/apollo-ios-dev/pull/580/files#diff-8ee8c63314d42c4a78ba9586041015b89525cb0b7ac57d6c65a70d9174acfed4) the codegen configuration documentation - [Here it is in the docs preview for this PR](https://www.apollographql.com/docs/deploy-preview/21c7d158a4c6ca9eb90ba8ff/ios/code-generation/codegen-configuration#output-options).

I'll take a look at the other two review comments now..

Tests/ApolloCodegenTests/ApolloCodegenConfigurationCodableTests.swift
calvincestari: Done in https://github.com/apollographql/apollo-ios-dev/pull/580/commits/34fea8b8a4b199999bbc1a09958805c7eefffe00.

Sources/AnimalKingdomAPI/AnimalKingdomAPI/Sources/Schema/CustomScalars/CustomDate.scalar.swift
calvincestari: OK, this is now done too. Logic in https://github.com/apollographql/apollo-ios-dev/pull/580/commits/6a22d3767ea74d77b426d3d55366bd3fc68f6444 and tests in https://github.com/apollographql/apollo-ios-dev/pull/580/commits/5743c59c1c77de0200c08776cdb045ab0b2ff5ee.

BobaFetters (approved):

AnthonyMDev (approved):
Thanks for making the changes! LGTM

PR #581, opened 2025-01-21
fix: Defer metadata extensions

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3503

This implements solution 2 in the issue description for the following reasons:
* This diff is minimal and resolves the issue for all module types. We could keep the extension but we're going to end up with multiple templates that need to handle the metadata insertion = IMO less optimal.
* I don't see any benefit to having the defer metadata within the same file but in an extension. The metadata is partitioned with a `MARK` comment line.
* Operation files also do not support custom edits so there is no need to preserve anything about previously generated operation files.


calvincestari:
Only thing I'm not clear on is whether the `deferredFragments` property should still be declared as `public` when the module type is `embeddedInTarget(accessModifier: .internal)`. Will the executor still be able to access it if not.

calvincestari:
We should probably also get `@defer` into an operation in one of our internal codegen reference projects like AnimalKingdom.

AnthonyMDev (approved):
Thanks for this fix!

PR #584, opened 2025-01-23
feature: Emit Identifiable conformance on SelectionSets

x-sheep:
This is a rework of #548, except this time it will only emit a conformance if the `keyFields` is set to exactly `["id"]` (instead of assuming the presence of an `id` field is enough).

If a type has a field named `id` it is actually impossible to override the behavior of `Identifiable`, so I believe there is no situation in which an automatic conformance to Identifiable isn't desirable.

This PR will not generate a getter named `id` (as was proposed earlier), so there won't be an unexpected field named `id` in any case.

### DIFF
--- a/apollo-ios-codegen/Sources/IR/IR+ComputedSelectionSet.swift
+++ b/apollo-ios-codegen/Sources/IR/IR+ComputedSelectionSet.swift
@@ -1,4 +1,5 @@
 import Foundation
+import GraphQLCompiler
 import OrderedCollections
 import Utilities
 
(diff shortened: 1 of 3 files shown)


apollo-ios-codegen/Sources/IR/IR+ComputedSelectionSet.swift
```diff
@@ -18,6 +19,24 @@ public struct ComputedSelectionSet {
 
   /// The `TypeInfo` for the selection set of the computed selections
   public let typeInfo: IR.SelectionSet.TypeInfo
+  
+  /// Indicates if the parent type has a single keyField named `id`.
+  public var isIdentifiable: Bool {
+    guard direct?.fields ["id"] ?? merged.fields["id"] != nil else {
```
calvincestari, line 25:
```suggestion
    guard direct?.fields["id"] ?? merged.fields["id"] != nil else {
```

calvincestari:
I think it would be beneficial to include a test that has `id` as the key field but it is not included in the selection set; test should not generate `Identifiable` conformance.

```
// given
schemaSDL = """
type Query {
  allAnimals: [Animal!]
}
type Animal @typePolicy(keyFields: "id") {
  id: ID!
  species: String!
}
"""

document = """
query TestOperation {
  allAnimals {
    species
  }
}
"""
```

calvincestari (commented):
One formatting suggestion.

x-sheep:
I've added another test.

calvincestari (approved):
Thanks for following through on this @x-sheep.

AnthonyMDev (approved):
I'm happy with this approach! Thanks so much @x-sheep!

PR #585, opened 2025-01-23
chore: test codegen configuration project cleanup

calvincestari:
Deleting some files that don't need to be committed because they're regenerated on each test run.


calvincestari:
Merging since the 'Codegen Test Configurations' job passed.

PR #587, opened 2025-01-24
fix: `DataDict` initialization of `deferredFragments` for named fragments

calvincestari:
_Discovered this bug when working on the selection set initializers for defer._

When deferred fragments are named fragments the deferred type is the fragments generated definition name.


AnthonyMDev (approved):
Good catch. Thanks for the quick fix!

calvincestari:
Merging since the only failed CI task if the docs deploy preview and this does not change any docs.

PR #588, opened 2025-01-26
Propagate likely unrecoverable watch errors to result handler

dfed:
Addresses [#3507](https://github.com/apollographql/apollo-ios/issues/3507)

I know creating a PR on a repo before maintainers have had a chance to review or comment on the linked issue is not a great practice. Please review this PR in the spirit in which it is given: as a starting discussion point and an attempt to further elucidate the points made in [#3507](https://github.com/apollographql/apollo-ios/issues/3507).

Getting this change to pass CI is going to take some doing – it looks like `AsyncResultObserver` expects that _any_ call to `resultHandler` triggers a refetch, which doesn't seem to true to me. I'd love some guidance before I try to shave that yak.


calvincestari:
@dfed - I'm going to close this PR since we've been trying to figure out the root issue via email. If we reach a point where this is the right direction for a solution we can reopen.

PR #586, opened 2025-01-24
Added documentation for @typePolicy

AnthonyMDev:
This PR adds documentation for the new `@typePolicy` feature.

PR #591, opened 2025-01-27
Switch sidebar

Meschreiber:


Meschreiber:
Not sure why the `Check PR Metadata` action is failing, but this should work fine. See DP: https://deploy-preview-436--apollo-docs-rewrite.netlify.app/docs/ios

calvincestari:
> Not sure why the `Check PR Metadata` action is failing, but this should work fine. See DP: https://deploy-preview-436--apollo-docs-rewrite.netlify.app/docs/ios

@Meschreiber - I saw that too and have PR https://github.com/apollographql/apollo-ios-dev/pull/592 up to fix it. I wouldn't worry about the failure here; it's just to ensure that PRs aren't created with an empty description.

calvincestari:
@Meschreiber - this PR looks good, thank you.

I noticed the URL for Programmatic Cache Keys leads to a 404 because that page isn't in `main` yet; it's only on @AnthonyMDev's documentation branch. It might be better if we change the base branch of this PR to Anthony's branch so that we can verify it all works there? Then both sets of changes go in together.

calvincestari:
> Ah, thanks @calvincestari ! Meant to just move the _sidebar.yaml version without the new page since @AnthonyMDev mentioned he still had some changes to make. If it's all good with you, I'll just remove that page from the _sidebar.yaml, change the base back, and then we can merge this in first?

I spoke to @AnthonyMDev about it now and he still has some changes to make to the docs so he's going to handle getting this commit into his when he does those changes.

AnthonyMDev:
I've merged this into #586 instead.

PR #594, opened 2025-01-28
Fix documentation for @typePolicy directive

x-sheep:
The document has `module` instead of `keyFields` as the parameter name.


calvincestari (approved):
Thank you @x-sheep, much appreciated.

PR #599, opened 2025-02-04
Fix data race for subscribers property

tahirmt:
Fixes https://github.com/apollographql/apollo-ios/issues/3512

This data race can happen if is a start/stop subscription at the same time as a message received on the websocket getting processed. The `subscribers` property is accessed from two threads at the same time causing a crash. 

I am not really sure how I can add unit tests for this change. The data race type tests are really hard to write.

### DIFF
--- a/apollo-ios/Sources/ApolloWebSocket/WebSocketTransport.swift
+++ b/apollo-ios/Sources/ApolloWebSocket/WebSocketTransport.swift
@@ -49,6 +49,7 @@ public class WebSocketTransport {
 
   private var queue: [Int: String] = [:]
 
+  @Atomic
   private var subscribers = [String: (Result<JSONObject, any Error>) -> Void]()
   private var subscriptions : [String: String] = [:]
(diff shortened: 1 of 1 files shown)


BobaFetters:
Thanks for putting up a PR for this fix! Is there more work you are planning for this PR since its in a draft state still?

tahirmt:
@BobaFetters no I just left it in draft because I wasn't sure of the feedback I would get regarding unit tests.

BobaFetters:
@tahirmt Gotcha, yea like you said this type of data race is hard to really write a test for so I don't think thats necessary here

BobaFetters (approved):

tahirmt:
@BobaFetters any idea when this will be released?

BobaFetters:
@tahirmt Not sure the exact day yet but we will likely do a release later this week once some other pending work is merged in as well.

BobaFetters:
@tahirmt We got the other pending work merged so will be going ahead and pushing a new release today

PR #601, opened 2025-02-12
feature: Reduce Generated Schema Types

BobaFetters:
Adds a new codegen configuration option to reduce the number of Object types that are generated so that only types that are referenced in an operation document or have a @typePolicy will be generated, instead of generating all types that implement a referenced Interface.

Closes apollographql/apollo-ios#3505


apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -703,6 +705,7 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
       public static let fieldMerging: FieldMerging = [.all]
       public static let operationDocumentFormat: OperationDocumentFormat = .definition
       public static let schemaCustomization: SchemaCustomization = .init()
+      public static let reduceGeneratedSchemaTypes: Bool = false
```
AnthonyMDev, line 708: Do we have a way for the user to force a specific type to be generated with this setting enabled? I know we had talked about this at one point, but it looks like we never followed through on it. I'm wondering if that is necessary or not?

One reason I could think it would be needed is for configuring programmatic cache keys for a specific object type. Though now with `@typePolicy` that is not as relevant. I'm not coming up with a specific use case right now that requires it, but I'm concerned that one may arise.

I suppose we could just wait and see if anyone asks for that with a valid use case. @calvincestari do you see any reason why we would need this feature?

BobaFetters: We talked about doing that for the JSON only response model work, but not for this since we have the type policies now.

calvincestari: > One reason I could think it would be needed is for configuring programmatic cache keys for a specific object type. Though now with `@typePolicy` that is not as relevant. I'm not coming up with a specific use case right now that requires it, but I'm concerned that one may arise.

Would it be possible to have a returned type be something that wasn't generated? If a type is referenced in an operation it would be generated no? Or would an underlying interface be 'discarded' if it didn't have a type policy and the field return type was expressly of that interface type?

BobaFetters: So this work is specifically focused on the objects, so interfaces will function the same as always. For the objects, if it is referenced in an operation document it will be generated, if its not the only way it will be generated is if it has a type policy.

calvincestari: OK, so in reference to @AnthonyMDev question here, I can't think of any need for forcing a specific type to be generated if it wasn't referenced. We had [this issue](https://github.com/apollographql/apollo-ios/issues/3508#issuecomment-2632279201) come in recently where a user wanted to generate all types regardless of usage but I don't believe that is a widespread need.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -721,6 +724,8 @@ public struct ApolloCodegenConfiguration: Codable, Equatable {
     ///   - selectionSetInitializers: Which generated selection sets should include
     ///     generated initializers.
     ///   - operationDocumentFormat: How to generate the operation documents for your generated operations.
+    ///   - schemaCustomization:Customization options to be applied to the schema during code generation.
+    ///   - reduceGeneratedSchemaTypes:Whether to reduce the number of schema types that are generated to only those that are referenced in an operation.
```
AnthonyMDev, line 728: Both of these lines probably need a space after the `:` in order for documentation to display properly.

AnthonyMDev (commented):
Just a couple of minor comments. This looks really good!

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
AnthonyMDev: > Would it be possible to have a returned type be something that wasn't generated?

@calvincestari Yes. For a field with an interface type this is possible. But I'm not sure if that even matters. If the type is only ever returned abstractly based on an interface field, I can't think of a situation in which it would need to referenced in client code other than if you want to give it a `@typePolicy`.

I think we're fine to move forward without this now. If someone in the future comes to us with a use case for that, we can add it.

apollo-ios-codegen/Sources/ApolloCodegenLib/CodegenConfiguration/ApolloCodegenConfiguration.swift
```diff
@@ -1697,6 +1711,58 @@ extension ApolloCodegenConfiguration.FileOutput {
 }
 
 extension ApolloCodegenConfiguration.OutputOptions {
+  /// Designated initializer.
```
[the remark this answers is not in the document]
BobaFetters: Forgot to mark that as deprecated, its the initializer prior to adding the new option so moved it down with the other deprecated initializers.

apollo-ios/Sources/ApolloAPI/SchemaTypes/Object.swift
```diff
@@ -44,6 +44,6 @@ public struct Object: Hashable, Sendable {
   /// - Parameter interface: An ``Interface`` Type
   /// - Returns: A `Bool` indicating if the receiver implements the given ``Interface`` Type.
   public func implements(_ interface: Interface) -> Bool {
-    implementedInterfaces.contains(where: { $0 == interface })
```
calvincestari, line 47: Shouldn't we deprecate `implementedInterfaces` if it's no longer being used? As much as I can find it's only the generated `Object` initializers that use this; it really shouldn't even be a public property.

So adding the deprecation attribute will then throw warnings in previously generated projects and they can regenerate to have those object initializers 'fixed'.

calvincestari (commented):
Just a question about a property on `Object`, looks good otherwise.

apollo-ios/Sources/ApolloAPI/SchemaTypes/Object.swift
BobaFetters: I think the concern here when @AnthonyMDev and I talked about it was it being potentially a breaking change if someone is using that property to determine cache keys if we removed it. I think we could mark this as deprecated to remove in 2.0, would we want to actually use the `@` attribute to give warning in users code or do like we do with some other stuff and signify deprecated with a comment to avoid the warning for users?

calvincestari: I personally think people will ignore deprecations-as-comments and then be surprised when it's eventually removed; we should use deprecation annotations for this case. We didn't in other places where the generated code could not be changed to eliminate the warning. In this case though a regeneration is all that would be needed to get rid of the deprecation warnings.

BobaFetters: But currently, the `implementedInterface` is still being generated as part of the generated objects, so none of that will change with regeneration. We just changed how we are calculating if an object implements and interface to prepare for eventually get rid of `implementedInterfaces`.

calvincestari: You're right, I'm getting ahead of myself. I was thinking we'd do both in one commit; deprecate the property and remove it from the initializer. But that's not actually deprecating it, it's effectively removing the property because it'll never hold a value without being initialized. I still think we should correctly deprecate it with `@available` so that users stop depending on it.

BobaFetters: Yea I agree with that, I'll add the attribute.

BobaFetters: Actually looks like this also causes test failures due to the deprecation, so will probably need to adjust this like we have in other instances to a comment deprecation.

x-sheep: The correct way is to rename the existing property and mark it `private`, and then add a deprecated getter with the same name.

calvincestari (approved):

PR #605, opened 2025-02-19
ci: Adds new workflow for Swift JS bundle updates

calvincestari:
Some of the secops PRs require that the Swift JS bundle be updated so that there is no diff detected with the updated dependencies. It has so far been a manual process, this changes that so CI can do it for us.

### DIFF
--- a/.github/workflows/update-swift-js-bundle.yml
+++ b/.github/workflows/update-swift-js-bundle.yml
@@ -1,0 +1,39 @@
+name: "Update Swift JS Bundle"
+
+on:
+  workflow_dispatch:
+    inputs:
+      branch:
+        description: 'Branch'
+        required: true
+        type: string
+
+env:
+  BRANCH_NAME: ${{ github.event.inputs.branch }}
+
+jobs:
+  update-swift-js-bundle:
+    runs-on: macos-latest
+    timeout-minutes: 5
+    name: "Update Swift JS Bundle"
+    steps:
+    - name: Checkout Repo
+      uses: actions/checkout@v4
+    - name: Setup SSH Keys
+      uses: webfactory/ssh-agent@v0.7.0
+      with:
+        ssh-private-key: |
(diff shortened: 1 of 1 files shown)


.github/workflows/update-swift-js-bundle.yml
```diff
@@ -0,0 +1,39 @@
...
+  BRANCH_NAME: ${{ github.event.inputs.branch }}
+
+jobs:
+  update-swift-js-bundle:
+    runs-on: macos-latest
+    timeout-minutes: 5
+    name: "Update Swift JS Bundle"
+    steps:
+    - name: Checkout Repo
+      uses: actions/checkout@v4
```
BobaFetters, line 21: We want this to run on the PR branch right? I think this is going to checkout the `main` branch by default since this is setup to trigger manually and not being triggered by a PR where it would automatically use that ref. Think we need to add `ref: ${{ env.BRANCH_NAME }}`

calvincestari: Yes, good catch! Thank you.

BobaFetters (approved):
Thanks for doing this!

.github/workflows/update-swift-js-bundle.yml
```diff
@@ -0,0 +1,41 @@
...
+    - name: Configure Git
+      uses: ./.github/actions/configure-git
+    - name: Build JS Bundle
+      shell: bash
+      working-directory: apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript
+      run: ./auto_rollup.sh
+    - name: Commit and Push Branch
+      shell: bash
+      run: |
+        git checkout -b ${{ env.BRANCH_NAME }}
```
calvincestari, line 38: @BobaFetters - with the change to use `ref` above we don't need this `git checkout` command right? Should be on the right branch already at this point.

BobaFetters: Yea good catch!

PR #607, opened 2025-02-20
Updating package config for test apis

BobaFetters:
Updating the spm config for testing apis to use local `apollo-ios` pathing


calvincestari (approved):
LGTM!

PR #608, opened 2025-03-03
Add mutateIfFulfilled functions

AnthonyMDev:
This addresses the limitations described in https://github.com/apollographql/apollo-ios/issues/3443. This is solved by adding a new `mutateIfFulfilled` function that only allows for the properties of an inline fragment to be mutated if the fragment already is fulfilled, while still preventing a fragment from being added/removed from an existing model.

### DIFF
--- a/Tests/ApolloTests/MutatingSelectionSetTests.swift
+++ b/Tests/ApolloTests/MutatingSelectionSetTests.swift
@@ -86,4 +86,153 @@ class MutatingSelectionSetTests: XCTestCase {
     expect(data2.hero.name).to(equal("Leia"))
   }
 
+  func test__mutateIfFulfilled_typeCaseProperties__typeCaseIsFulfilled_mutatesProperties() throws {
+    // given
+    var model = TypeCaseSelectionSet.AsHero(__typename: "Hero", name: "A", age: 1).asRootEntityType
+
+    // when
+    let didMutate = model.mutateIfFulfilled(\.asHero) { hero in
+      hero.name = "B"
+      hero.age = 2
+    }
+
+    // then
+    expect(didMutate).to(beTrue())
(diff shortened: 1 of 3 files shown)


Tests/ApolloTests/MutatingSelectionSetTests.swift
```diff
@@ -86,4 +86,153 @@ class MutatingSelectionSetTests: XCTestCase {
...
+
+    struct AsHero: MockMutableInlineFragment {
+      typealias RootEntityType = TypeCaseSelectionSet
+
+      static var __parentType: any ApolloAPI.ParentType { Object.mock }
+
+      public var __data: DataDict = .empty()
+      init(_dataDict: DataDict) { __data = _dataDict }
+
+      static var __selections: [Selection] { [
```
calvincestari, line 166: I don't think it matters to this particular test suite but it looks like we're missing the `age` selection.

calvincestari (commented):
Will these extensions cover named fragments too?

calvincestari:
Also, the test failures look like proper errors.

```
Error: cannot find type 'ApolloAPI' in scope
  mutating func mutateIfFulfilled<T: ApolloAPI.InlineFragment>(
                                     ^~~~~~~~~
Error: 'RootEntityType' is not a member type of type 'T'
  ) -> Bool where T.RootEntityType == Self.RootEntityType {
                  ~ ^
Error: cannot find type 'ApolloAPI' in scope
  mutating func mutateIfFulfilled<T: ApolloAPI.InlineFragment>(
                                     ^~~~~~~~~
Error: 'RootEntityType' is not a member type of type 'T'
  ) -> Bool where T.RootEntityType == Self {
                  ~ ^
```

calvincestari (approved):

Iron-Ham:
🥳

PR #615, opened 2025-03-14
Initial version of Caching Rewrite RFC

BobaFetters:
Read the formatted RFC [here](https://github.com/apollographql/apollo-ios-dev/blob/main/apollo-ios/Design/rfc-caching-rewrite.md)

### DIFF
--- a/apollo-ios/Design/rfc-caching-rewrite.md
+++ b/apollo-ios/Design/rfc-caching-rewrite.md
@@ -1,0 +1,171 @@
+###  **This RFC is a work in progress. Additions and changes will be made throughout the design process. Changes will be accompanied by a comment indicating what sections have changed.**
+
+ # Background
+
+ In order to provide a better caching experience and more functionality, the existing caching APIs need a significant overhaul. We believe everything from cache configuration and management, to the structure of the SQLite database can be improved to provide users with a better more powerful caching experience. This overhaul will involve breaking changes and will be released as a major version bump when complete.
+
+ The features outlined in this RFC are considered core features for the initial release of the new caching system. While there are other features on the roadmap for caching they will come as future updates after the initial release of the core features. Some of those features include:
+
+ - Object and Field watchers
+ - Faceted searching support
+
+ # Proposal
+
+ While there are other features we want to eventually provide as well, the initial release of the new caching system will focus on the following features (described in more detail below). Restructuring the SQLite database will open up the ability to query and operate on individual fields. Configuring size limits on your caches to handle auto cache eviction, as well as configuring _what_ should/shouldn't be evicted. Along with being able to set Time to Live (TTL) on a per object or per field basis, enabled by the new SQLite structure.
+
+ ## SQLite Structure
+
+ The current SQLite structure stores your data as a cache key and a JSON blob of the response data from your query. This makes it difficult to operate on the data without having to pull the JSON out of the SQLite and then deserialize it. With the new caching system we will be storing data in the SQLite so that each row represents a single field of an object. This will allow us to be able to implement many features which currently either wouldn't be possible, or would be very complex to implement and have poor performance implications.
+
+ ## Time to Live (TTL)
+
+ To help control when data becomes "stale" we will introduce the ability to configure a TTL for an entire object or an individual field. This means that when accessing data from the cache, if any part of an object has surpassed its TTL it will cause a cache miss and the object will need to be re-fetched from the server. This will be configured through your schema and operations using a local directive, and allow the flexibility to set overall TTL's at the schema level but override them at an operation level if needed.
+
+ ### Directive
+
+ We will create a new local directive `@cacheControl(maxAge: Int)`, where the `maxAge` is measured in seconds, which can be applied to objects and fields in your schema or operations:
+
+ ```graphql
+ type Song @cacheControl(maxAge: 3600) {
+  id: ID!
+  name: String!
+  description: String! @cacheControl(maxAge: 900)
+  artist: Artist!
+ }
+
+ type Artist @cacheControl(maxAge: 3600) {
+  id: ID!
+  name: String!
+  genre: String!
+ }
+ ```
+
+ ```graphql
+ query GetSong($id: ID!) {
+  getSong(id: $id) {
+    id
+    name
+    description @cacheControl(maxAge: 300)
+    artist @cacheControl(maxAge: 900) {
+      id
+      name 
+    }
+  }
+ }
+ ```
+
+ Using the above example schema/operation, you can see how you would apply TTL to objects and fields in the schema. However, you can also override TTL for objects/fields in individual operations if a particular operation has a different requirement for how fresh its data should be. When generating the code TTL will be tied to individual fields, so applying to an object applies to all of its fields. However, the most specific TTL will take precedence, with specificity in order of least to most being `Schema Type > Schema Field > Operation Type > Operation Field`, so in the example above for the song `description` field, it would get its TTL applied as follows:
+
+- `Song` schema type TTL of 60 minutes applies
+- `description` schema type field overrides the previous TTL of 60 and sets it to 15 minutes
+- `description` query field overrides the schema field TTL and sets it to 5 minutes
+
+So for the `GetSong` query, the `description` field would have a TTL of 5 minutes, but any other query using the `description` field would have a TTL of 15 minutes (unless you override it there as well). The same logic would apply to nested objects/fields such as the artist in the `GetSong` query above. In the generated code, fields at the `SelectionSet` level would have a static metadata TTL property so that the final value is available for use when querying the cache for data.
+
+ ## Cache Configuration
+
+ We want to provide more control over the caches to allow for better cache management, this includes the ability to set size limits for both the in-memory and SQLite caches which can be used to trigger automatic cache eviction. As well as being able to configure what get evicted, and specifying whether object deletion should cascade to child objects. Providing configuration to a `NormalizedCache` will be done through providing it with a `NormalizedCacheConfiguration` struct that looks like this:
+
+ ```swift
+ public struct NormalizedCacheConfiguration {
+  let sizeLimit: Int
+  let autoEvictionSize: Int
+  let evictionFieldsIgnoreList: [Field]
+  let delegate: NormalizedCacheConfigurationDelegate 
+ }
+ ```
+
+ These properties and how they function will be described in more detail below.
+
+ ### Cache Size Limits
+
+ Configuring cache size limits will be done using the `sizeLimit` property of the `NormalizedCacheConfiguration` and will represent the maximum size in kilobytes (KB) you wish the cache to be before evicting some data to free up space.
+
+ For the Apollo provided `InMemoryNormalizedCache` the size will be actively monitored with every new write to the cache, and if the size limit has been exceeded an overflow will be triggered.
+
+ For the Apollo provided `SQLiteNormalizedCache` the size will be monitored more passively at set time intervals, and the same as the in memory cache if the size limit is exceeded and overflow will be triggered.
+
+ ### Eviction Configuration
+
+ By default the automatic eviction done from the caches will be handled as a least recently used (LRU) style cache. However, by using the `evictionFieldsIgnoreList` property on the `NormalizedCacheConfiguration` you will be able to provide an array of `Field` objects representing fields from any of your types you wish to not have evicted for any reason, such as it being long lived data that is unlikely to change.
+
+ ### Cache Overflow Handling
+
+ When an overflow of a cache is detected because it has exceeded its size limit, before any eviction takes place a call will be made to the `delegate` of the `NormalizedCacheConfiguration` to provide an opportunity to do or complete any work you feel is necessary before cache eviction takes place. You can also choose to skip the eviction, until the next check when an overflow is detected and triggered again. The `NormalizedCacheConfigurationDelegate` will look something like:
+
+ ```swift
+ public protocol NormalizedCacheConfigurationDelegate {
+  func willRunCacheEviction() -> Bool
+ }
+ ```
+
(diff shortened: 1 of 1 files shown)


apollo-ios/Design/rfc-caching-rewrite.md
```diff
@@ -0,0 +1,171 @@
...
+
+ ## Time to Live (TTL)
+
+ To help control when data becomes "stale" we will introduce the ability to configure a TTL for an entire object or an individual field. This means that when accessing data from the cache, if any part of an object has surpassed its TTL it will cause a cache miss and the object will need to be re-fetched from the server. This will be configured through your schema and operations using a local directive, and allow the flexibility to set overall TTL's at the schema level but override them at an operation level if needed.
+
+ ### Directive
+
+ We will create a new local directive `@cacheControl(maxAge: Int)`, where the `maxAge` is measured in seconds, which can be applied to objects and fields in your schema or operations:
```
AnthonyMDev, line 26: I don’t love having to handle `maxAge` in seconds always. It’s one thing to do the math once when you’re setting it up, but every time you are going back to refer to it, it’s frustrating to have the mental overhead of trying to figure out that, for example 604800 means one week. This could be a good place to use a @oneOf input object here that had fields for seconds, minutes, and days.

I know we are trying to match the `cacheControl` directive used by Apollo Kotlin. Is there a chance that both teams can align on a new version of this directive API?

apollo-ios/Design/rfc-caching-rewrite.md
```diff
@@ -0,0 +1,171 @@
...
+
+ ## Cache Configuration
+
+ We want to provide more control over the caches to allow for better cache management, this includes the ability to set size limits for both the in-memory and SQLite caches which can be used to trigger automatic cache eviction. As well as being able to configure what get evicted, and specifying whether object deletion should cascade to child objects. Providing configuration to a `NormalizedCache` will be done through providing it with a `NormalizedCacheConfiguration` struct that looks like this:
+
+ ```swift
+ public struct NormalizedCacheConfiguration {
+  let sizeLimit: Int
+  let autoEvictionSize: Int
+  let evictionFieldsIgnoreList: [Field]
```
AnthonyMDev, line 73: What is the `API` of this `Field` type? How is this going to reference a specific field on a specific entity type? What about referencing a field on an interface type?

apollo-ios/Design/rfc-caching-rewrite.md
```diff
@@ -0,0 +1,171 @@
...
+  let delegate: NormalizedCacheConfigurationDelegate 
+ }
+ ```
+
+ These properties and how they function will be described in more detail below.
+
+ ### Cache Size Limits
+
+ Configuring cache size limits will be done using the `sizeLimit` property of the `NormalizedCacheConfiguration` and will represent the maximum size in kilobytes (KB) you wish the cache to be before evicting some data to free up space.
```
AnthonyMDev, line 82: I know I had originally said we should change this from MBs to KBs, but I'm thinking that the easiest API to use might be MBs as a decimal number. You don't want MBs as an `Int` b/c then you can only configure it in increments of a MB. But with a decimal number you could configure it more granularly (eg. 3.25 MBs).

apollo-ios/Design/rfc-caching-rewrite.md
```diff
@@ -0,0 +1,171 @@
...
+
+ For the Apollo provided `InMemoryNormalizedCache` the size will be actively monitored with every new write to the cache, and if the size limit has been exceeded an overflow will be triggered.
+
+ For the Apollo provided `SQLiteNormalizedCache` the size will be monitored more passively at set time intervals, and the same as the in memory cache if the size limit is exceeded and overflow will be triggered.
+
+ ### Eviction Configuration
+
+ By default the automatic eviction done from the caches will be handled as a least recently used (LRU) style cache. However, by using the `evictionFieldsIgnoreList` property on the `NormalizedCacheConfiguration` you will be able to provide an array of `Field` objects representing fields from any of your types you wish to not have evicted for any reason, such as it being long lived data that is unlikely to change.
+
+ ### Cache Overflow Handling
```
AnthonyMDev, line 92: I think we need to put more thought into the Cache Overflow handling. Nothing here is wrong, it just feels incomplete to me. The complexity of how users can do manual cache eviction are unclear to me right now and what common use cases for that are. I’m also not sure if 20% of the cache size limit is always going to be desirable for auto eviction. Might need more configuration of that.

apollo-ios/Design/rfc-caching-rewrite.md
```diff
@@ -0,0 +1,171 @@
...
+ type Artist {
+  id: ID!
+  name: String!
+  genre: String!
+ }
+ ```
+
+ ### Other options
+
+ We have considered an implementation of cascading deletions where a heuristic is used to determine relationships between objects, which would detect parent/child object relationships and having the default behavior be to always delete child objects with parent objects. Currently we don't plan to move forward with this implementation.
```
AnthonyMDev, line 146: Thanks for adding the mention of this discussion. After consideration, I agree that this isn't probably something worth implementing, but I wanted a note here so that others can give us feedback on this.

apollo-ios/Design/rfc-caching-rewrite.md
```diff
@@ -0,0 +1,171 @@
...
+
+ Cache chaining will handle writing to an `InMemoryNormalizedCache` and then subequently writing to the `SQLiteNormalizedCache` automatically for you. In order to handle this there will be a new `NormalizedCache` implementation `ChainedNormalizedCache` which will use both the in memory and SQLite cache and handle chaining for you. By default all data will chain to both caches, however you will be able to exclude object types from either the in memory or SQLite cache so they are only stored in one or the other if desired. That configuration will look like this:
+
+ ```swift
+ public final class ChainedNormalizedCache: NormalizedCache {
+  private let inMemoryCache: InMemoryNormalizedCache
+  private let sqliteCache: SQLiteNormalizedCache
+
+  private let inMemoryExcludeList: [Object]
+  private let sqliteExcludeList: [Object]
```
AnthonyMDev, line 158: The exclude lists could easily put the cache in a state where you are inadvertently getting a lot of cache misses that you don’t want. Maybe there is a better way to handle this that would be more user-friendly. I think we need to first answer the question of what use-cases would anyone actually want to use this for, then we can design the right APIs to enable the use cases without making it a massive foot-gun.

I have a feeling that it would make sense to have the ability to exclude short-lived data from being cached in the persistent SQLite cache, but I can't come up with a situation in which you would want to exclude something from the in-memory cache but still have it in the SQLite cache. That would just lead to a lot of repeated cache misses.

apollo-ios/Design/rfc-caching-rewrite.md
```diff
@@ -0,0 +1,171 @@
...
+ We have considered an implementation of cascading deletions where a heuristic is used to determine relationships between objects, which would detect parent/child object relationships and having the default behavior be to always delete child objects with parent objects. Currently we don't plan to move forward with this implementation.
+
+ ## Cache Chaining
+
+ Cache chaining will handle writing to an `InMemoryNormalizedCache` and then subequently writing to the `SQLiteNormalizedCache` automatically for you. In order to handle this there will be a new `NormalizedCache` implementation `ChainedNormalizedCache` which will use both the in memory and SQLite cache and handle chaining for you. By default all data will chain to both caches, however you will be able to exclude object types from either the in memory or SQLite cache so they are only stored in one or the other if desired. That configuration will look like this:
+
+ ```swift
+ public final class ChainedNormalizedCache: NormalizedCache {
+  private let inMemoryCache: InMemoryNormalizedCache
+  private let sqliteCache: SQLiteNormalizedCache
```
AnthonyMDev, line 155: I know we had discussed having the ability to provide an array of `[any NormalizedCache]` objects to be chained together, but that does seem like it's adding complexity for the sake of flexibility that I don't think is actually ever needed. So I am on board with only allowing you to provide a single in-memory cache and a single SQLite cache. 

However, users currently have the ability to provide their own `NormalizedCache` conforming types, and I think that it makes sense that they would want to provide those to this as well. Maybe there should be some way of denoting the semantic difference between an in-memory and persistent cache, but I still think we should allow users to use those with chained caching.

AnthonyMDev (commented):
This is a great RFC to kick-off the discussion! Thanks @BobaFetters!

One thing that I think is missing from this is a discussion of changes to the `NormalizedCache` protocol's API itself. I know that we have gotten requests from users to give the cache protocol functions access to more context of the operation that is being run during a cache read (looking at you @jimisaacs 😉).

I'd like to get some more feedback from users on this RFC about what would be needed there and see if we want to include changes to that protocol in this update as well.

BobaFetters:
> One thing that I think is missing from this is a discussion of changes to the `NormalizedCache` protocol's API itself. I know that we have gotten requests from users to give the cache protocol functions access to more context of the operation that is being run during a cache read (looking at you @jimisaacs 😉).

Waiting to add anything about the `NormalizedCache` protocol until your work for Swift Concurrency is further along since there will be changes to the protocol to go along with that work.

apollo-ios/Design/rfc-caching-rewrite.md
BobaFetters: I'm open to making this MB again, technically if we want to more closely match other caching apis such as `NSCache` etc they typically work off of bytes which may be more familiar to people when working with this. Would love to hear thoughts from the community though.

apollo-ios/Design/rfc-caching-rewrite.md
BobaFetters: Can add more detail for manual cache eviction, should support either single object deletion, transaction for group deletions of multiple objects, or full cache clear as a base.

Also the amount that gets clear from auto eviction is already detailed as being configurable [here](https://github.com/apollographql/apollo-ios-dev/pull/615/files#diff-bcd9fa477694ac1140e3c3157d6a30abcf5002bdcdd8841ab43db867692207d9R102).

apollo-ios/Design/rfc-caching-rewrite.md
BobaFetters: Main use case I can think of is preventing short lived data between app runs by excluding from the sqlite cache. No specific use case for excluding from in-memory cache, although I guess it would allow an in-between of no caching and only caching some data in memory.

I don't think this would put the cache in a state where you get inadvertent cache misses because users would be explicitly saying what they dont want in the caches therefor it _should_ result in a cache miss. Imagine feedback from the community here about how they may use this feature would go a long way.

apollo-ios/Design/rfc-caching-rewrite.md
BobaFetters: With us providing the chained cache as an implementation of `NormalizedCache` I think it makes sense that it accepts the 2 cache implementations we provide. Still leaves open the possibility for users to create their own chained caches to go along with custom caches. If we want to support this we could added protocols for in-memory vs persistent caches that inherit from normalize cache and go that route. Would need to consider if theres any added functionality the protocols would need to properly support the chaining or if it would mostly just be plug and play. Curious if thats some the community would find useful to have.

apollo-ios/Design/rfc-caching-rewrite.md
BobaFetters: Initially thinking it would be [Field](https://github.com/apollographql/apollo-ios/blob/main/Sources/ApolloAPI/Selection.swift#L13) although looking through this again, makes more sense to probably have an `Object`  or `Interface` and `Field` together as a tuple so you could specify fields on an object or interface and the field type has the extra context from the object/interface as well.

PatrickDanino:
Thanks for the work and proposal on this, it's nice to see progress being made.

It would be helpful for me to understand how this would work in practice. For example, what happens if you have two queries, each including the same type and/or field, but with a different `maxAge` value? How would you know what TTL to use for the eviction process?

Separately, there's no mention in the RFC of types or fields you may not want to cache at all. Is that implied by using `maxAge: 0`? Or would it make sense to have a `@cacheControl(noCache: true)`?

BobaFetters:
> Thanks for the work and proposal on this, it's nice to see progress being made.
> 
> It would be helpful for me to understand how this would work in practice. For example, what happens if you have two queries, each including the same type and/or field, but with a different `maxAge` value? How would you know what TTL to use for the eviction process?
> 
> Separately, there's no mention in the RFC of types or fields you may not want to cache at all. Is that implied by using `maxAge: 0`? Or would it make sense to have a `@cacheControl(noCache: true)`?

@PatrickDanino So the cached data will have a `createdTime` and `lastUpdated` timestamp, but the cache data itself doesnt know if it is stale or not, that is up to the operation/query that is being run to decide.

So in your example of having 2 different queries which have overlapping data, those 2 queries may be used in separate parts of an application that have different requirements for how "fresh" the data needs to be. So query A may say the TTL is 5 min while query B says the TTL is 15 min. In that scenario, assuming the cached data is currently 10 min old, if you run query B it will pull the data from the cache, if you run query A it will treat the data as stale and the data would then be fetched from the server and updated into the cache. So if query B was run again, it would still pull from the cache since the data is now brand new, and so would query A.

As for eviction, by default it will be LRU (least recently used) based while having some [configuration available](https://github.com/apollographql/apollo-ios-dev/blob/design/rfc-caching/apollo-ios/Design/rfc-caching-rewrite.md#eviction-configuration), but since the cached data doesn't inherently know if it is stale based that isnt factored in.

For ignoring objects from the cache yea that currently isn't called out, I will be updating the RFC soon with more details, specifically around the `@cacheControl` directive. But to answer your question yes currently the idea would be that having a `maxAge` of 0 would mean the object is ignored from the cache.

BobaFetters:
@pixelmatrix It sounds like what you are running into should be resolved with this new caching work. The changes to the SQLite structure to store 1 field per row to facilitate some more advanced functionality, vs storing the data based on the query that was run as a blob. Also the Time-to-live (max-age) support will let you control how long that data is used from the cache as well.

finebel:
Thanks for the RFC! Are there any updates regarding the timeline when these changes are expected to be available for general usage (i.e. production ready)?

PR #618, opened 2025-03-18
feature: Configure `URLRequest` timeout interval

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3522.

The _desired_ solution mentioned in the issue was to introduce an additional parameter on the `fetch` and `send` functions. That solution would have introduced a breaking change due to those functions being part of the protocols `ApolloClientProtocol` and `NetworkTransport` respectively.

The solution implemented in this PR is to leverage the `RequestContext` marker protocol and introduce a configuration specific sub-protocol. The idea is that this can be extended in the future to introduce more request configuration parameters as needed. It also is not a breaking change.

### DIFF
--- a/apollo-ios/Sources/Apollo/RequestContext.swift
+++ b/apollo-ios/Sources/Apollo/RequestContext.swift
@@ -11,8 +11,8 @@ import ApolloAPI
 /// that they cannot get just from the existing operation. It can be anything that conforms to this protocol.
 public protocol RequestContext {}
 
-/// A specialized request context that specifies configuration details for the URLRequest.
-public protocol RequestConfigurationContext: RequestContext {
+/// A specialized request context that specifies configuration details for the URLRequest timing.
+public protocol RequestContextTimeoutConfigurable: RequestContext {
   /// The timeout interval specifies the limit on the idle interval allotted to a request in the process of
   /// loading. This timeout interval is measured in seconds.
   var requestTimeout: TimeInterval { get }
--- a/apollo-ios/Sources/Apollo/RequestContext.swift
+++ b/apollo-ios/Sources/Apollo/RequestContext.swift
@@ -11,7 +11,10 @@ import ApolloAPI
 /// that they cannot get just from the existing operation. It can be anything that conforms to this protocol.
 public protocol RequestContext {}
 
-/// A specialized request context that specifies configuration details for the URLRequest timing.
+/// A request context specialization protocol that specifies options for configuring the timeout of a `URLRequest`.
+///
+/// A `RequestContext` object can conform to this protocol to provide a custom `requestTimeout` for an individual request.
+/// If the `RequestContext` for a request does not conform to this protocol, the default request timeout of `URLRequest` will be used.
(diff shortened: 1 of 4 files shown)


apollo-ios/Sources/Apollo/RequestContext.swift
```diff
@@ -10,3 +10,10 @@ import ApolloAPI
 /// This allows the various interceptors to make modifications, or perform actions, with information
 /// that they cannot get just from the existing operation. It can be anything that conforms to this protocol.
 public protocol RequestContext {}
+
+/// A specialized request context that specifies configuration details for the URLRequest.
+public protocol RequestConfigurationContext: RequestContext {
```
AnthonyMDev, line 15: I think this should be a specific protocol for this feature named something like `RequestContextTimeoutConfigurable`. Any additional features we want to add onto the `RequestContext` in the future should be their own individual protocols.

That way people can opt-in to the configuration options they want and not be forced to implement configuration for other options they don't need to get the ones they do. This is also more likely to prevent us from needing breaking changes in the future when we want to make changes to these things (or at least limit the surface area effected by changes).

AnthonyMDev (changes requested):
Implementation looks good. Just think we should change the naming to reflect a more specialized approach to this and future configuration options.

apollo-ios/Sources/Apollo/RequestContext.swift
```diff
@@ -10,3 +10,10 @@ import ApolloAPI
 /// This allows the various interceptors to make modifications, or perform actions, with information
 /// that they cannot get just from the existing operation. It can be anything that conforms to this protocol.
 public protocol RequestContext {}
+
+/// A specialized request context that specifies configuration details for the URLRequest timing.
```
AnthonyMDev, line 14:
```suggestion
/// A request context specialization protocol that specifies options for configuring the timeout of a `URLRequest`.
///
/// A `RequestContext` object can conform to this protocol to provide a custom `requestTimeout` for an individual request.
/// If the `RequestContext` for a request does not conform to this protocol, the default request timeout of `URLRequest` will be used.
```

apollo-ios/Sources/Apollo/RequestContext.swift
```diff
@@ -10,3 +10,10 @@ import ApolloAPI
...
+public protocol RequestContextTimeoutConfigurable: RequestContext {
+  /// The timeout interval specifies the limit on the idle interval allotted to a request in the process of
+  /// loading. This timeout interval is measured in seconds.
```
AnthonyMDev, line 17:
```suggestion
  /// The timeout interval specifies the limit on the idle interval allotted to a request in the process of
  /// loading. This timeout interval is measured in seconds.
  ///
  /// The value of this property will be set as the `timeoutInterval` on the `URLRequest` created for this GraphQL request.
```

AnthonyMDev (approved):
Just added some more detail to the documentation, but this LGTM! Thanks @calvincestari!

PR #619, opened 2025-03-19
feature: Enhanced Client Awareness request extension

calvincestari:
Closes https://github.com/apollographql/apollo-ios/issues/3523.

Adding the ability for the Apollo iOS to send the library name (`apollo-ios`) and version (eg: `1.18.0`) metadata in the `extensions` key of each request. This is different from the existing [client info metadata](https://www.apollographql.com/docs/graphos/routing/observability/client-id-enforcement) that can be configured by user-code.


apollo-ios/Sources/Apollo/Constants.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 
 public enum Constants {
-    public static let ApolloVersion: String = "1.18.0"
+  public static let ApolloName = "apollo-ios"
```
AnthonyMDev, line 4: Can we add a comment of how this is being used. And maybe change it to `ApolloClientName`

AnthonyMDev (commented):
I'm not sure this is the appropriate way to handle Client Awareness. The `RequestContext` should be used for request specific configuration. But it seems to me that this is likely something you would want configured for your entire Apollo Client to be used across all requests.

With this design, you would need to ensure that you correctly pass in a request context to every call site where you are making any requests to the client. It would be easy to miss some of them.

apollo-ios/Sources/Apollo/Constants.swift
calvincestari: I'll rename the version constant too; they should both reflect their _client_ intent.

apollo-ios/Sources/Apollo/ApolloClient.swift
```diff
@@ -50,21 +52,36 @@ public class ApolloClient {
   /// - Parameters:
   ///   - networkTransport: A network transport used to send operations to a server.
   ///   - store: A store used as a local cache. Note that if the `NetworkTransport` or any of its dependencies takes a store, you should make sure the same store is passed here so that it can be cleared properly.
```
AnthonyMDev, line 54: Just need to add documentation of this parameter here.

apollo-ios/Sources/Apollo/RequestChainNetworkTransport.swift
```diff
@@ -52,13 +52,16 @@ open class RequestChainNetworkTransport: NetworkTransport {
   ///   - requestBodyCreator: The `RequestBodyCreator` object to use to build your `URLRequest`. Defaults to the provided `ApolloRequestBodyCreator` implementation.
   ///   - useGETForQueries: Pass `true` if you want to use `GET` instead of `POST` for queries, for example to take advantage of a CDN. Defaults to `false`.
   ///   - useGETForPersistedQueryRetry: Pass `true` to use `GET` instead of `POST` for a retry of a persisted query. Defaults to `false`.
```
AnthonyMDev, line 54: Docs here too!

AnthonyMDev (approved):

Tests/ApolloTests/AutomaticPersistedQueriesTests.swift
```diff
@@ -145,9 +145,6 @@ class AutomaticPersistedQueriesTests: XCTestCase {
       expect(file: file, line: line, version).to(equal(1))
 
       expect(file: file, line: line, sha256Hash).to(equal(O.operationIdentifier))
-
-    } else {
-      expect(file: file, line: line, ext).to(beNil())
```
calvincestari, line 150: We're needlessly testing that `extensions` is empty; we shouldn't be enforcing that outside of what this test is concerned with, which is the persisted query keys in `extensions`.

Tests/ApolloTests/AutomaticPersistedQueriesTests.swift
```diff
@@ -238,9 +235,6 @@ class AutomaticPersistedQueriesTests: XCTestCase {
       expect(file: file, line: line, version).to(equal(1))
 
       expect(file: file, line: line, sha256Hash).to(equal(MockHeroNameQuery.operationIdentifier))
-
-    } else {
-      expect(file: file, line: line, ext).to(beNil())
```
calvincestari, line 243: We're needlessly testing that `extensions` is empty; we shouldn't be enforcing that outside of what this test is concerned with, which is the persisted query keys in `extensions`.

apollo-ios/scripts/get-version.sh
```diff
@@ -4,5 +4,5 @@ directory=$(dirname "$0")
 source "$directory/version-constants.sh"
 
 constantsFile=$(cat "$directory/../$APOLLO_CONSTANTS_FILE")
-currentVersion=$(echo $constantsFile | sed 's/^.*ApolloVersion: String = "\([^"]*\).*/\1/')
+currentVersion=$(echo $constantsFile | sed 's/^.*ApolloClientVersion: String = "\([^"]*\).*/\1/')
```
calvincestari, line 7: This script is used indirectly in many other tests and was cascading failures.

apollo-ios/Sources/Apollo/ApolloClient.swift
```diff
@@ -49,22 +51,40 @@ public class ApolloClient {
   ///
   /// - Parameters:
   ///   - networkTransport: A network transport used to send operations to a server.
-  ///   - store: A store used as a local cache. Note that if the `NetworkTransport` or any of its dependencies takes a store, you should make sure the same store is passed here so that it can be cleared properly.
-  public init(networkTransport: any NetworkTransport, store: ApolloStore) {
+  ///   - store: A store used as a local cache. Note that if the `NetworkTransport` or any of its dependencies takes
+  ///   a store, you should make sure the same store is passed here so that it can be cleared properly.
+  ///   - sendClientMetadataExtension: Specifies whether client library metadata is sent in each request `extensions`
+  ///   key. Client library metadata is the Apollo iOS library name and version. Defaults to `true`.
```
martinbonnin, line 57: Sorry I'm late to the party. I'm guessing we went with `sendClientMetadataExtension` instead of just `enhanced client awareness` because we are keeping the door open to adding more things in those metadatas?

If that is the case, can we have a link to a public specification of what the `metadata` contains? This would help users understand what exactly is being sent and what tradeoff they are doing by disabling the feature.

calvincestari: I'm not opposed to renaming it but was there a discussion somewhere about the naming?

martinbonnin: I don't think so.

PR #623, opened 2025-03-27
fix: Xcode 16.3 CLI installation

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3518.

#### Cause
Investigating this issue revealed that Xcode 16.2 has the plugins execute from within the `SourcePackages` directory; that made sense why we could [simply backtrack two folders](https://github.com/apollographql/apollo-ios-dev/blob/1.19.0/apollo-ios/Plugins/InstallCLI/InstallCLIPluginCommand.swift#L31) (`../../`) and get access to `checkouts`. Xcode 16.3 now has a separate directory that it gives the plugin complete control over, which is outside of `SourcePackages`. The result is that we need to backtrack further up the directory tree to get to `checkouts`.

#### Solution
* This changeset gets the active version of Xcode by using `xcrun`. This will allow for different Xcode versions being installed in `/Applications`, or wherever else, and only get the version for the 'active' application. Once we know the version of Xcode being used then the correct known script location can be used.
* I've tested this with Xcode 16.2, all the Xcode 16.3 beta versions, and the latest Xcode 16.3 RC 2. The version-based solution works for all of them.
* This bug and subsequent fix is targeted at Xcode projects vs. SPM projects. SPM projects appear unaffected by the change in plugin execution context.

#### Discarded
These are the other solutions I discarded:
1. `if #available(..)` - I don't think this can dynamically get the installed Swift version and instead targets a specific OS version. I believe Xcode 16.2 and 16.3 can both be installed on macOS Sequoia (15.3) so that's not helpful.
2. `#if swift(..)` - This is conditional compilation code for build time; we need the solution to operate on the users' machine. Also not helpful.
3. Crawl up the directory tree from the plugin context work directory until `SourcePackages` is found. This is still a viable alternative and if the Xcode version approach causes more problems in the future we should try this.

### DIFF
--- a/apollo-ios/Plugins/InstallCLI/InstallCLIPluginCommand.swift
+++ b/apollo-ios/Plugins/InstallCLI/InstallCLIPluginCommand.swift
@@ -1,8 +1,13 @@
 import Foundation
 import PackagePlugin
+import os
 
 @main
 struct InstallCLIPluginCommand: CommandPlugin {
+
+  enum Error: Swift.Error {
+    case CannotDetermineXcodeVersion
+  }
 
   func performCommand(context: PackagePlugin.PluginContext, arguments: [String]) async throws {
     let dependencies = context.package.dependencies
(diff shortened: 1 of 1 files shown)


AnthonyMDev (approved):
Thanks for resolving this Calvin. I know this must have been a frustrating one to figure out.

calvincestari:
Two more notes I forgot to add yesterday:
1. This does not work for local dependency configurations. I tested this on 1.19.0 too with both Xcode 16.2 and 16.3 and that didn't work either so there is no change in behaviour. I just don't think we note this in our docs though.
2. We are still using some to-be-deprecated functions. It doesn't look like they're deprecated yet but have documentation warnings that they will be in future versions. I expect at that time Apple will provide alternate ways for that functionality though. So..something to look out for in the future.

PR #635, opened 2025-04-24
feature: Remove SQLite.swift

BobaFetters:
Rewriting the SQLiteDotSwiftDatabase class to ApolloSQLiteDatabase and now interacting directly with the SQLite C API. The goal of this rewrite is to remove the third party dependency on SQLite.swift and preserve all existing functionality as is, not add any new functionality.

Closes apollographql/apollo-ios#3548


AnthonyMDev (approved):
This looks really great to me! Thanks so much @BobaFetters.

Tests/ApolloTests/SQLiteDotSwiftDatabaseBehaviorTests.swift
```diff
@@ -17,10 +17,21 @@ class SQLiteDotSwiftDatabaseBehaviorTests: XCTestCase {
     XCTAssertEqual(rows.count, 1)
 
     // Use SQLite directly to manipulate the database (cannot be done with SQLiteDotSwiftDatabase)
```
calvincestari, line 19:
```suggestion
    // Use SQLite directly to manipulate the database (cannot be done with ApolloSQLiteDatabase)
```

Tests/ApolloTests/SQLiteDotSwiftDatabaseBehaviorTests.swift
```diff
@@ -1,13 +1,13 @@
 import XCTest
 @testable import ApolloSQLite
 import ApolloInternalTestHelpers
-import SQLite
+import SQLite3
 
 class SQLiteDotSwiftDatabaseBehaviorTests: XCTestCase {
```
calvincestari, line 6:
```suggestion
class ApolloSQLiteDatabaseBehaviorTests: XCTestCase {
```

calvincestari: Two suggested changes in this file but I'm not completely sold on the value this test is providing, maybe we don't need it?

calvincestari (approved):

Tests/ApolloTests/SQLiteDotSwiftDatabaseBehaviorTests.swift
BobaFetters: Good catch on both of these, thanks!

Tests/ApolloTests/SQLiteDotSwiftDatabaseBehaviorTests.swift
BobaFetters: I think the point of the test is meant to verify errors get thrown if someone directly modifies the database, which they could do if they know the file url to connect to and use the sqlite api, so I think this is still a valid test to have.

PR #636, opened 2025-04-27
Fix potential data race

tahirmt:
Similar to #599 this PR attempts to fix a data race in the `subscriptions` property within `WebSocketTransport`.  It has the same issue as `subscribers` property.

If you receive subscription message at the same time as a subscription getting started / stopped there is a potential of reading the array while it is being modified resulting in a crash.


BobaFetters (approved):
This LGTM thanks @tahirmt

AnthonyMDev (approved):
Thanks for this fix!

PR #637, opened 2025-04-29
fix: Cache read null list item

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3527

Null list items would previously generate a `wrongType` error if stored and read from the cache. This refactors the execution logic to correctly handle values from cache references in lists.

### DIFF
--- a/apollo-ios/Sources/Apollo/ExecutionSources/CacheDataExecutionSource.swift
+++ b/apollo-ios/Sources/Apollo/ExecutionSources/CacheDataExecutionSource.swift
@@ -41,11 +41,15 @@ struct CacheDataExecutionSource: GraphQLExecutionSource {
       case let reference as CacheReference:
         return deferredResolve(reference: reference).map { $0 as AnyHashable }
 
-      case let referenceList as [CacheReference]:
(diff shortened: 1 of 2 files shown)


apollo-ios/Sources/Apollo/ExecutionSources/CacheDataExecutionSource.swift
```diff
@@ -55,7 +59,7 @@ struct CacheDataExecutionSource: GraphQLExecutionSource {
                 } else {
                   return error
                 }
-              }
+              }.map { $0 as AnyHashable }
           }.map { $0._asAnyHashable }
```
AnthonyMDev, line 63: It's not ideal that we are having to map this and cast is at `AnyHashable` twice here, but it might be unavoidable. Are you certain we need these both?

calvincestari: I believe it's only the result type from `mapError` that actually needs to be cast; it's a build error without it. So there shouldn't actually be any casting unless there was an error no?

PR #647, opened 2025-05-15
[v2] [1/X] Update ApolloStore and associated APIs to Sendable

AnthonyMDev:
This will not compile until GraphQLExecutor async/await support is added. Coming in future stacked branch

### DIFF
--- a/apollo-ios/Sources/Apollo/ApolloStore.swift
+++ b/apollo-ios/Sources/Apollo/ApolloStore.swift
@@ -21,13 +21,21 @@ public protocol ApolloStoreSubscriber: AnyObject, Sendable {
 }
 
 /// The `ApolloStore` class acts as a local cache for normalized GraphQL results.
-#warning("TODO: temp @unchecked Sendable to move forward; is not yet thread safe")
-public class ApolloStore: @unchecked Sendable {
-  private let cache: any NormalizedCache
+#warning("TODO: Docs. ReaderWriter usage; why you should not share a cache with 2 stores, etc.")
+public final class ApolloStore: Sendable {
   private let queue: DispatchQueue
   private let readerWriterLock = ReaderWriter()
 
-  internal var subscribers: [any ApolloStoreSubscriber] = []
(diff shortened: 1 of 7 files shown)


calvincestari:

apollo-ios/Sources/Apollo/ApolloStore.swift
```diff
@@ -375,9 +334,144 @@ public class ApolloStore: @unchecked Sendable {
...
   }
+
+  // MARK: - Deprecations
+
+  /// Clears the instance of the cache.
+  ///
+  /// - Parameters:
+  ///   - callbackQueue: The queue to call the completion block on. Defaults to `DispatchQueue.main`.
+  ///   - completion: [optional] A completion block to be called after records are merged into the cache.
+  @available(*, deprecated, renamed: "clearCache()")
```
BobaFetters, line 350: Do we need all the deprecated functions? Since this is going to be a major release seems like the time for us to clear out deprecated stuff vs adding more, especially since the expectation for the major release would be to have breaking changes.

apollo-ios/Sources/Apollo/DispatchQueue+Optional.swift
```diff
@@ -2,7 +2,11 @@ import Dispatch
 
 extension DispatchQueue {
 
-  static func performAsyncIfNeeded(on callbackQueue: DispatchQueue?, action: @escaping () -> Void) {
+  @available(*, deprecated)
```
BobaFetters, line 5: Same thing with these I would think we just get rid of anything that is getting deprecated vs keeping it around.

apollo-ios/Sources/Apollo/ApolloStore.swift
AnthonyMDev: Agreed. My thought was to include the deprecated functions through the beta releases to make the migration path easier. That way, for the final public release, we can still recommend people with large projects to start on the last beta release, fix any build errors, and _then_ fix deprecations before moving to 2.0.

apollo-bot2:

PR #648, opened 2025-05-15
[v2] GraphQLExecutor async/await

AnthonyMDev:
This makes the GraphQL Executor use async/await.

**Note:** Unit test changes in this PR are incomplete and are not intended for review. All unit tests are in flux for this and all stacked PRs. A final review of all unit-tests will be done prior to release.


calvincestari:

apollo-ios/Sources/Apollo/InterceptorRequestChain.swift
```diff
@@ -232,7 +232,7 @@ final public class InterceptorRequestChain: Cancellable, RequestChain {
     _ error: any Error,
     request: HTTPRequest<Operation>,
     response: HTTPResponse<Operation>?,
-    completion: @escaping (Result<GraphQLResult<Operation.Data>, any Error>) -> Void
+    completion: @escaping @Sendable (Result<GraphQLResult<Operation.Data>, any Error>) -> Void
```
calvincestari, line 235: It's probably easier, and safer, if we define a new type for this sendable closure.

apollo-ios/Sources/Apollo/ApolloStore.swift
```diff
@@ -218,7 +223,7 @@ public final class ApolloStore: Sendable {
     
     final func loadObject(forKey key: CacheKey) -> PossiblyDeferred<Record> {
       self.loader[key].map { record in
-        guard let record = record else { throw JSONDecodingError.missingValue }
+        guard let record = record else { return Record(key: key, [:]) }
```
calvincestari, line 226: Is this correct that we aren't throwing `missingValue` any more?

apollo-ios/Sources/Apollo/InterceptorRequestChain.swift
BobaFetters: Agree with this, I would be in favor of giving explicit names to closures we are using throughout the code.

PR #650, opened 2025-05-15
[v2] [4/x] NetworkTransport APIs

AnthonyMDev:
Changes the `NetworkTransport` protocol and implementations to return an `AsyncThrowingStream` of results instead of using a completion handler.

`ApolloClient` will have new APIs that return the `AsyncThrowingStream` in a later PR in the stack. For now, we are wrapping that stream and returning its results through the completion handlers for `ApolloClient`.

### DIFF
--- a/apollo-ios/Sources/ApolloWebSocket/SplitNetworkTransport.swift
+++ b/apollo-ios/Sources/ApolloWebSocket/SplitNetworkTransport.swift
@@ -4,8 +4,12 @@ import Apollo
 import ApolloAPI
 #endif
 
+#warning("""
+TODO: This is messy. Why is http network transport called "uploadingNetworkTransport"?
+Websocket transport should be typesafe to a protocol that guaruntees it supports web sockets/ subscriptions
+""")
 /// A network transport that sends subscriptions using one `NetworkTransport` and other requests using another `NetworkTransport`. Ideal for sending subscriptions via a web socket but everything else via HTTP.
-public class SplitNetworkTransport {
+public final class SplitNetworkTransport: Sendable {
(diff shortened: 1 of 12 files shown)


calvincestari:

apollo-ios/Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -450,49 +455,52 @@ extension WebSocketTransport: NetworkTransport {
...
+        throw error
       }
+    }
 
-      return WebSocketTask(self, operation) { [weak store, contextIdentifier] result in
-        Task {
-          do {
-            let jsonBody = try result.get()
-            let response = GraphQLResponse(operation: operation, body: jsonBody)
+    #warning("TODO: stream never finishes. WebSocketTask does not report subscription termination.")
```
calvincestari, line 466: I think it has been a request though and now that we have async streams we can signal when that stream is 'done' vs. just never calling a closure again.

apollo-ios/Sources/ApolloWebSocket/SplitNetworkTransport.swift
```diff
@@ -1,13 +1,20 @@
...
 #if !COCOAPODS
-import Apollo
-import ApolloAPI
+  import Apollo
+  import ApolloAPI
 #endif
 
+#warning(
+  """
+  TODO: This is messy. Why is http network transport called "uploadingNetworkTransport"?
```
BobaFetters, line 10: Assume we are fixing this in a different PR?

apollo-bot2:

PR #651, opened 2025-05-15
[v2] [5/X] Client Awareness Refactor

AnthonyMDev:
Refactored Client Awareness features to us a new `ClientAwarenessMetadata` struct.

`clientName` and `clientVersion` have been removed from `NetworkTransport` and are not configured using this struct.


calvincestari:

apollo-ios/Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -337,7 +317,11 @@ public class WebSocketTransport: @unchecked Sendable {
     self.websocket.delegate = nil
   }
 
+<<<<<<< HEAD
```
calvincestari, line 320: Leftover conflict marker here.

apollo-ios/Sources/Apollo/ApolloClient.swift
```diff
@@ -62,33 +60,30 @@ public class ApolloClient: ApolloClientProtocol, @unchecked Sendable {
   ///   key. Client library metadata is the Apollo iOS library name and version. Defaults to `true`.
   public init(
     networkTransport: any NetworkTransport,
-    store: ApolloStore,
-    sendEnhancedClientAwareness: Bool = true
+    store: ApolloStore
```
BobaFetters, line 63: Shouldn't there be some form of initialization for the `ClientAwarenessMetadata` in this defaul `init()` as well? Since there was initialization for the bool value previously and if someone uses this initializer it's not possible to set a value.

PR #655, opened 2025-05-19
Remove double async dispatch queue call on failure

sean-escendant:


calvincestari:
> @calvincestari you see any issue with this?

I'll take a better look at this PR tomorrow but if it is a double call then merging shouldn't be an issue.

sean-escendant:
> @sean-escendant - thanks for the contribution but we cannot merge this PR until you sign the Apollo CLA.

Sorry I don't want to provide my mailing address to sign the CLA. Perhaps a future version of the CLA will allow anonymous contributors.

calvincestari:
@sean-escendant, unfortunately Apollo cannot accept contributions without a signed CLA. That means I'll need to close this PR without merging it.

We can push the change forward still but it will need to be done by a contributor who has signed the CLA, or by one of the Apollo iOS maintainers.

PR #658, opened 2025-05-28
fix: Multiline warning annotation line separator

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3559.

Simple fix - needs the line separator to be explicit rather than the default list-style separator `,\n`.


AnthonyMDev (approved):
Awesome, quick fix! 🚢

PR #659, opened 2025-05-30
fix: Generate local cache mutation referenced fragments as mutable

calvincestari:
Fixes https://github.com/apollographql/apollo-ios/issues/3557.

This causes any fragments referenced within a local cache mutation to be generated as mutable too, including any fragments within those fragments, etc., etc.

This is achieved by way of a local 'override' flag that is used to augment the fragment `isLocalCacheMutation` check and cause it to be treated as a local cache mutation.

I considered adding the local cache mutation directive to the fragment definition instead of the override flag but didn't feel that was truthful to what was actually happening and would be more difficult to debug later.


AnthonyMDev (approved):
Nice work!

PR #665, opened 2025-06-06
New README info architecture and content

bignimbus:
## TODO

Before marking as ready for review:

- [ ] Update other READMEs
- [ ] Verify links


calvincestari (approved):
Looks good, thank you.

_At some time we should change the build status badge too. I think that might be more relevant to apollo-ios-dev though with the nightly build._

PR #677, opened 2025-06-25
[v2] [8/X] Add OperationResponseFormat

AnthonyMDev:
In 2.0, we want `ApolloClient` to have functions with different return values based on if we know there will be a single returned result vs when we expect multiple responses. We can expect multiple responses when the operation includes a `@defer` directive OR when fetching from both the cache and the network (ie. 1.0 `CachePolicy.returnCacheDataAndFetch`).

In order to support these different functions this PR adds `OperationResponseFormat` to `GraphQLOperation`. Now that operations are generated with this `associatedtype`, they can indicate if they use a `SingleResponseFormat` or an `IncrementalDeferredResponseFormat`. Other response formats can be added in the future (for `@stream` possibly).

This allows us to create a variety of functions with different single/multiple return types on `ApolloClient` with constraining `where` clauses.


apollo-ios/Sources/Apollo/ApolloClient.swift
```diff
@@ -1,51 +1,72 @@
...
 
-#warning("TODO: this unsafe is not properly made atomic. Fix this")
-  /// The current default cache policy.
-  nonisolated(unsafe) public static var `default`: CachePolicy = .returnCacheDataElseFetch
+  ///  Always fetch results from the server, and don't store these in the cache.
+  //  case fetchIgnoringCacheCompletely
+
+  #warning("TODO: this unsafe is not properly made atomic. Fix this")
+  //  /// The current default cache policy.
+  //  nonisolated(unsafe) public static var `default`: CachePolicy = .cacheFirst
```
AnthonyMDev, line 29: All changes in `ApolloClient` itself are just for POC and are still in flux.

The plan is to only include those first three cache policies.

apollo-ios/Sources/Apollo/ApolloClient.swift
```diff
@@ -1,51 +1,72 @@
...
-public typealias GraphQLResultHandler<Data: RootSelectionSet> = @Sendable (Result<GraphQLResult<Data>, any Error>) -> Void
+public typealias GraphQLResultHandler<Data: RootSelectionSet> = @Sendable (Result<GraphQLResult<Data>, any Error>) ->
+  Void
+
+struct RequestConfigurationContext {
+  @TaskLocal static var taskLocal: RequestConfigurationContext = .init()
+
+  var requestTimeout: Int32 = 30
+  var writeResultsToCache: Bool = true
+}
```
AnthonyMDev, line 44: This is something I'm toying with, but I'm not sure if it will make the final release. This probably shouldn't even be here right now, but a lot of stuff is under construction at the same time, and I did want to get the primary components of this PR up for discussion. Happy to discuss this here, but it will be finalized (or removed) in a later PR, so let's not let this hold up approving this one.

This adds a new `RequestConfigurationContext` that provides a `TaskLocal` value. This can replace the usage of `RequestContext` for timeout configuration (`RequestContextTimeoutConfigurable`) and other future configuration options.

This configuration context also currently includes a `writeResultsToCache` property. The intention is that `CachePolicy` can be used only for indicating if we should fetch from the cache vs network (and only used for query operations). While configuration of writing results to the cache could be done using this task local value.

Because users can also use their own custom task local values anywhere they want, I'm actually considering if we can remove `RequestContext` completely.

The primary issue I see with this new approach is that I haven't found a great way to allow users to decide what the initial default values are for the configuration properties. If you want to change one of these values for a single operation, it's straightforward:

```swift
var requestConfig = Apollo.RequestConfigurationContext.taskLocal // would probably change the name of this property
requestConfig.writeResultsToCache = false
Apollo.RequestConfigurationContext.$taskLocal.withValue(requestConfig) {
  let result = client.fetch(query: myQuery)
  ...
}
```

But if you wanted to, for instance, change the default `requestTimeout` for _all_ of your operations, I haven't come up with a  great way to do that yet.

I'm thinking that maybe you can initialize `ApolloClient` with a default request configuration and then use the TaskLocal version to override the default when you want to.

apollo-ios/Sources/Apollo/ApolloClientProtocol.swift
```diff
@@ -27,7 +27,7 @@ public protocol ApolloClientProtocol: AnyObject {
   ///   - resultHandler: [optional] A closure that is called when query results are available or when an error occurs.
```
AnthonyMDev, line 27: `ApolloClientProtocol` will be updated once I've finalized the API for `ApolloClient` (or maybe moved into `ApolloTestHelpers`?)

apollo-ios/Sources/Apollo/ApolloClient.swift
calvincestari: We discussed this earlier today. Should be some changes incoming, like #680, from pairing on that.

apollo-ios/Sources/Apollo/ApolloClient.swift
```diff
@@ -1,51 +1,72 @@
...
 public enum CachePolicy: Sendable, Hashable {
   /// Return data from the cache if available, else fetch results from the server.
-  case returnCacheDataElseFetch
-  ///  Always fetch results from the server.
-  case fetchIgnoringCacheData
-  ///  Always fetch results from the server, and don't store these in the cache.
-  case fetchIgnoringCacheCompletely
-  /// Return data from the cache if available, else return an error.
-  case returnCacheDataDontFetch
+  case cacheFirst
```
calvincestari, line 12: I wanted to comment that `cacheFirst` sounds like a different default behaviour than `returnCacheDataElseFetch` but on second thought I don't think the intent of the cache policy has actually changed, the documentation for the case is still the same. That leads me to think that we need to consider another name.
* `returnCacheDataElseFetch` is mutually exclusive, i.e.: one or the other will happen, not both.
* `cacheFirst` sounds like there is a secondary action, i.e.: cache first, then network.

calvincestari: If the case name changes from this then we should take a second look at the rest too.

apollo-ios/Sources/ApolloAPI/GraphQLOperation.swift
```diff
@@ -155,6 +143,25 @@ public extension GraphQLSubscription {
...
+public struct SingleResponseFormat: OperationResponseFormat {}
+
+/// An incremental response format for an operation which uses the `@defer` directive.
+public struct IncrementalDeferredResponseFormat: OperationResponseFormat {
+  public let deferredFragments: [DeferredFragmentIdentifier: any SelectionSet.Type]
+
+  public init(deferredFragments: [DeferredFragmentIdentifier: any SelectionSet.Type]) {
+    self.deferredFragments = deferredFragments
+  }
+}
```
calvincestari, line 163: I think we're missing a response format for HTTP subscriptions? They're closely related to `IncrementalDeferredResponseFormat` but not defined by the use of `@defer`.

apollo-ios/Sources/Apollo/ApolloClient.swift
AnthonyMDev: That's fair. For the work I'm doing right now, I have predefined `FetchBehavior` values called `CacheThenNetwork` and `CacheElseNetwork`. I'm open to changing this to `.cacheElseNetwork.

The reason why I went with `cacheFirst` and `networkFirst` was to align with the [cache policy definitions in ApolloKotlin](https://apollographql.github.io/apollo-kotlin/kdoc/older/4.2.0/apollo-http-cache/com.apollographql.apollo.cache.http/-http-fetch-policy/index.html).

@martinbonnin Do you think that `cacheFirst` has been confusing to users in any way?

apollo-ios/Sources/Apollo/ApolloClient.swift
AnthonyMDev: New CachePolicy API is implemented in #684

apollo-ios/Sources/Apollo/ApolloClient.swift
```diff
@@ -83,62 +106,42 @@ public class ApolloClient: ApolloClientProtocol, @unchecked Sendable {
...
+    query: Query,
+    cachePolicy: CachePolicy? = nil,
+    context: (any RequestContext)? = nil
+  ) throws -> AsyncThrowingStream<GraphQLResult<Query.Data>, any Error>
+  where Query.ResponseFormat == IncrementalDeferredResponseFormat {
+    return try self.networkTransport.send(
+      query: query,
+      cachePolicy: cachePolicy ?? self.defaultCachePolicy,
+      context: context
+    )
```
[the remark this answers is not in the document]
AnthonyMDev: Finalized fetch APIs in #684

apollo-ios/Sources/Apollo/ApolloClient.swift
AnthonyMDev: I decided to go with something more specific and clear for all of the cache policy values. This is now implemented in #684.

apollo-ios/Sources/ApolloAPI/GraphQLOperation.swift
AnthonyMDev: You're right! I'll be getting to that soon in another PR.

apollo-ios/Sources/Apollo/ApolloClient.swift
martinbonnin: @AnthonyMDev we haven't had any issue with [`CacheFirst`](https://apollographql.github.io/apollo-kotlin/kdoc/older/4.2.0/apollo-normalized-cache/com.apollographql.apollo.cache.normalized/-fetch-policy/-cache-first/index.html). This is also what [Apollo Client is using](https://www.apollographql.com/docs/react/data/queries#supported-fetch-policies) and I think quite established today. If you can, I'd love to get those aligned between different platforms. 

Ping @BoD

AnthonyMDev: Yeah, I don't think it's _too_ confusing, but I do agree with Calvin that `.cacheElseNetwork`,`.cacheThenNetwork`, and `.networkElseCache` are clearer.

I prefer the clearer versions, but if we want to weigh parity of the naming across platforms higher than iOS team's preference on naming, I'm okay with that. I just want to make a decision today and move forward.

My feeling is that, while naming parity is nice, it's not something that we have consistently in other areas of the libraries currently. If we were aligned on naming in most of the library, then I'd try to stick with that, but not sure if there is much value in doing that here if we aren't consistent with it.

@bignimbus Do you have any opinion on this?

PR #678, opened 2025-06-25
[v2] [9/X] Interceptor/RequestChain improvements

AnthonyMDev:
## FetchBehavior

`GraphQLRequest` now has a new `FetchBehavior` instead of `CachePolicy`. `CachePolicy` is really only relevant in its current form for queries. In future PR, `ApolloClient` is going to be altered so that relevant cache configuration options are exposed based on the operation type (eg. Mutations should never read from the cache, but may write to it). `ApolloClient` will be responsible for translating per-operation parameters into a `FetchBehavior` to provide to `NetworkTransport` and `GraphQLRequest`.

## Multiple interceptors types

There are now multiple types of interceptors:
- `GraphQLInterceptor`
  - Can inspect and mutate the `GraphQLRequest` and `GraphQLResponse`
  - **Will be renamed from `ApolloInterceptor` to `GraphQLInterceptor` once project is building again**
- `HTTPInterceptor`
  - Can inspect and mutate the `URLRequest`
  - After network response can inspect the `HTTPURLResponse` (readonly) and mutate the actual raw response `Data` **prior to parsing**
- `CacheInterceptor`
  - Handles read/write of cache data
  - Read currently runs before `GraphQLInterceptors` (not sure if that is the desired behavior, we should discuss)
  - Write runs after parsing
- `ResponseParsingInterceptor`
  - Handles the parsing of the response Data into the `GraphQLResponse`

## `GraphQLResponse` and `HTTPResponse` separated

Previously, there was a single `GraphQLResponse` which included the `HTTPResponse` and optionally the `ParsedResult` (if the parsing interceptor had been called already). Now, since different interceptors will be called pre/post parsing, we have separate types for these response objects.

## Interceptors requested as needed

Instead of creating all of the interceptors upfront, the `RequestChain` is now given the `InterceptorProvider`. This allows interceptors to only be requested (and thus instantiated) when they are needed. For example, if the request fails prior to conversion to a `URLRequest`, the `HTTPInterceptors` and the `ParsingInterceptor` will never be instantiated.

This also means that `InterceptorProvider` functions are now passed the `GraphQLRequest` as it exists at the time those interceptors are needed. For example, if the `GraphQLInterceptors` mutate the `GraphQLRequest, the `InterceptorProvider.httpInterceptors(for: request)` will be passed the request  that has been altered by the previous interceptors.

## Removal of `ApolloErrorInterceptor`

With the new `RequestChain` flow, it was untenable to create a single error interceptor type. Because different types of data will be available at different steps in the chain, but an error could be thrown at any time, it would be difficult to understand what data was available when an error was thrown, making the API unwieldy. 

Additionally, the new implementation of traversing the interceptors made it very difficult to capture the request and response objects at the time of an error being thrown, since any interceptor could have mutated them.

Instead of a dedicated `ApolloErrorInterceptor` type, you can now just provide a _regular_ interceptor that handles errors. You can handle errors by providing a`GraphQLInterceptor` that handles them as the first interceptor returned by your provider. If any other interceptors, or the `ApolloURLSession` throw an error, this interceptor may catch it. You can then rethrow it; throw a different error; or trigger a retry by throwing a `RequestChain.Retry` error.

For handling errors in the response, the `InterceptorResultStream` now has a new `mapErrors` function. This allows you to do the same error handling for errors that occur during the response. It also allows you to return a manually created `GraphQLResult` to be emitted as a recovery option.

This implementation does remove the ability to recover with a manually created `GraphQLResult` **when an error is thrown prior to the network response**. But it doesn't really make sense that you should be able to do that to me. However if user feedback provides a use case for it, it is easy to add that functionality back in.

---
### Appendix A: `RequestChain` Order of Operations

For clarity, this is the current flow of data through the `RequestChain` as of this PR. 

- If necessary based on `FetchBehavior`, cache read via `CacheInterceptor`
- `GraphQLInterceptors` receive and may mutate `Request`
- `GraphQLRequest.toURLRequest()` called to obtain `URLRequest`
- `HTTPInterceptors` receive and may mutate `URLRequest`
- `ApolloURLSession` handles networking with `URLRequest`
- `HTTPInterceptors` receive stream of `HTTPResponse` objects for each chunk & may mutate raw chunk `Data` stream
- `ResponseParsingInterceptor` receives `HTTPResponse` and parses data chunks into stream of `GraphQLResponse`
- `GraphQLInterceptors` receive and may mutate `GraphQLResponse` with parsed `GraphQLResult` and (possibly) cache records.
- If necessary, cache write via `CacheInterceptor`
- `GraphQLResponse` emitted out to `NetworkTransport`

When PR #684 is merged, it will modify this so that the `GraphQLInterceptors` receive the `Request` prior to the cache read and the result from the cache will also be sent through the interceptors.


apollo-ios/Sources/Apollo/RequestChain.swift
```diff
@@ -69,7 +60,7 @@ struct RequestChain<Request: GraphQLRequest>: Sendable {
     return doInRetryingAsyncThrowingStream(request: request) { request, continuation in
       let didYieldCacheData = try await handleCacheRead(request: request, continuation: continuation)
```
AnthonyMDev, line 61: ## Discussion Point: Ordering of Cache Read

I'm thinking this might not be the desired behavior, but would like feedback. Right now, the `CacheInterceptor` is called to do a cache read as the first step of the `RequestChain`. This happens before we call the `GraphQLInterceptors`.

Would it make more sense to have the `GraphQLInterceptors` called before the cache read? Then they could alter the `GraphQLRequest` before it is even sent to the cache for the cache read.

Currently, if the cache read is successful, and the `FetchBehavior` isn't requiring a network fetch on a successful cache read, the `GraphQLInterceptors` don't even get called at all. I think that is probably not right.

AnthonyMDev: I've decided to make this change. It is implemented in #684.

apollo-ios/Sources/Apollo/Interceptors/ApolloInterceptor.swift
```diff
@@ -1,38 +1,28 @@
-import Foundation
 import Combine
+import Foundation
+
 #if !COCOAPODS
```
BobaFetters, line 4: Think we talked about this before, but for full 2.0 release we are dropping CocoaPods support yea?

AnthonyMDev: Oh! I think we had talked about that at some point, but I haven't added it to my todo list for 2.0. I'd love to drop it though. Let's talk about this in our next team meeting and make a final decision.

AnthonyMDev: I just posted on the RFC asking for opinions on dropping Cocoapods, if we don't get any objections, I think it's a no brainer. Even if we do, they would have to be very compelling objections to make us keep it around.

PR #680, opened 2025-06-26
[v2] [10/X] Remove RequestContext

AnthonyMDev:
The `RequestContext` is no longer needed due to the introduction of `@TaskLocal` values in Swift. For the beta, we will be deprecating the `RequestContext`, but it will be fully removed in the final release.

`RequestContextTimeoutConfigurable` will be replaced with a new property on `GraphQLRequest` that will be coming in a future PR.


AnthonyMDev:
@danieltiger Would love to get your feedback on this. I know you created the PR to introduce `RequestContext` to solve for some of your own use cases previously. Do you feel that the removal of this feature now would be a blocker for you?

I'm pretty confident that anyone using `RequestContext` would be able to achieve the same thing using `@TaskLocal` values, but I want to verify that you agree with that assessment.

calvincestari (approved):

PR #681, opened 2025-06-26
Create ClientContext for ClientAwareness

AnthonyMDev:
Added `ClientContext` as an internal `@TaskLocal` value that can be consumed by other internal parts of the system.

We are currently using this to provide a `ClientAwarenessMetadata` that is provided by the user during client initialization. This is used by the default implementations of `DefaultRequestBodyCreator` and `GraphQLRequest.createDefaultRequest()`.

### DIFF
--- a/apollo-ios/Sources/Apollo/ApolloClient.swift
+++ b/apollo-ios/Sources/Apollo/ApolloClient.swift
@@ -80,11 +80,13 @@ public final class ApolloClient: ApolloClientProtocol, Sendable {
   public init(
     networkTransport: any NetworkTransport,
     store: ApolloStore,
-    defaultCachePolicy: CachePolicy = .cacheFirst
+    defaultCachePolicy: CachePolicy = .cacheFirst,
+    clientAwarenessMetadata: ClientAwarenessMetadata = ClientAwarenessMetadata()
   ) {
     self.networkTransport = networkTransport
     self.store = store
     self.defaultCachePolicy = defaultCachePolicy
+    self.context = ClientContext(clientAwarenessMetadata: clientAwarenessMetadata)
   }
 
   /// Creates a client with a `RequestChainNetworkTransport` connecting to the specified URL.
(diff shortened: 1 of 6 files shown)


apollo-ios/Sources/Apollo/ApolloClient.swift
```diff
@@ -130,13 +134,21 @@ public final class ApolloClient: ApolloClientProtocol, Sendable {
 
   public func fetch<Query: GraphQLQuery>(
     query: Query,
-    cachePolicy: CachePolicy? = nil
+    cachePolicy: CachePolicy? = nil,
```
calvincestari, line 137: What's meant to be on the next line after the new `,`?

apollo-ios/Sources/Apollo/ApolloClient.swift
```diff
@@ -130,13 +134,21 @@ public final class ApolloClient: ApolloClientProtocol, Sendable {
...
   where Query.ResponseFormat == IncrementalDeferredResponseFormat {
-    return try self.networkTransport.send(
-      query: query,
-      cachePolicy: cachePolicy ?? self.defaultCachePolicy
-    )
+    return try sendQuery(query: query, cachePolicy: cachePolicy, requestConfiguration: requestConfiguration)
+  }
+  private func sendQuery<Query: GraphQLQuery>(
+    query: Query,
+    cachePolicy: CachePolicy?,
```
calvincestari, line 144: > What's meant to be on the next line after the new `,`?

Same here..

apollo-ios/Sources/Apollo/ClientAwarenessMetadata.swift
```diff
@@ -11,12 +11,12 @@ public struct ClientAwarenessMetadata: Sendable {
...
+  public let clientApplicationName: String?
 
   /// The version of the application. This value is sent for the header "apollographql-client-version".
   ///
   /// Defaults to `nil`.
-  public var clientApplicationVersion: String?
+  public let clientApplicationVersion: String?
 
   /// Determines if the Apollo iOS library name and version should be sent with the telemetry data.
   ///
```
calvincestari, line 22: _This comment is not for this line but down on [L28](https://github.com/apollographql/apollo-ios-dev/pull/681/files#diff-37d41f6d8707f6c8ae93505134ae6d66628227b45a1ca58df249eb01a64fb59dR28) instead._

`includeApolloLibraryAwareness` should probably be have `private(set)` added to the definition; set once via the initializer is all that's needed. Alternatively a `let` with no default.

apollo-ios/Sources/Apollo/ApolloClient.swift
AnthonyMDev: This was a mistake, but these APIs are changed in #684, so this isn't an issue.

apollo-ios/Sources/Apollo/ApolloClient.swift
AnthonyMDev: This is fixed in next PR. Disregard here.

apollo-ios/Sources/Apollo/ClientAwarenessMetadata.swift
AnthonyMDev: That's fair. I'll change that in the next PR that I make.

PR #685, opened 2025-06-27
[v2] [13/X] Refactor GraphQLQueryWatcher

AnthonyMDev:
The `GraphQLQueryWatcher` is now an `Actor` to ensure its thread safety. Significant changes to the implementation have been made to use structured concurrency, but the usage stays very similar.

Only significant difference is that I removed the `refetch()` function. The fetch functions are now public and there are overloads for all cache policies and `FetchBehavior`.

### DIFF
--- a/apollo-ios/Sources/Apollo/ApolloStore.swift
+++ b/apollo-ios/Sources/Apollo/ApolloStore.swift
@@ -31,7 +31,7 @@ public final class ApolloStore: Sendable {
 
   /// In order to comply with `Sendable` requirements, this unsafe property should
   /// only be accessed within a `readerWriterLock.write { }` block.
-  nonisolated(unsafe) private(set) var subscribers: [any ApolloStoreSubscriber] = []
+  nonisolated(unsafe) private(set) var subscribers: [SubscriptionToken: any ApolloStoreSubscriber] = [:]
 
   /// Designated initializer
   /// - Parameters:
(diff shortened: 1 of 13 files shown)


apollo-ios/Sources/Apollo/ApolloStore.swift
```diff
@@ -69,23 +69,25 @@ public final class ApolloStore: Sendable {
...
 
   /// Unsubscribes from notifications of ApolloStore content changes
   ///
   /// - Parameters:
-  ///    - subscriber: A subscribe that has previously been added via `subscribe`. To avoid retain cycles,
-  ///                  call `unsubscribe` on all active subscribers before they go out of scope.
-  public func unsubscribe(_ subscriber: any ApolloStoreSubscriber) {
-    Task {
+  ///    - subscriptionToken: An opaque token for the subscriber that was provided via `subscribe(_:)`.
+  ///    To avoid retain cycles, call `unsubscribe` on all active subscribers before they go out of scope.
```
calvincestari, line 86: If there is a way for us to have this done automatically and not rely on the user to unsubscribe, it would be a change for the better.

AnthonyMDev: I thought about this a lot today in regards to the query watcher being retained here. I actually think this is the lesser of two evils. 

Similar to using `Cancellable` in `Combine` (or similar in any FRP library), the user is responsible for disposal of their subscriptions. If they fail to do so, the subscriptions live indefinitely, and that inherently causes a memory leak. This is how it works in every FRP library I've ever seen. This isn't FRP per say, but the paradigm is similar.

If we were to hold a weak reference here, you just get a different problem. If users release their query watcher without cancelling it, then its listener block just silently stops being called. I'm not sure if that's better or worse. It can be confusing for the user and difficult to debug.

I'm not 100% sold on either way, so I'm willing to talk about this more between now and the final release of 2.0.

PR #693, opened 2025-07-01
[v2] Refactor GraphQLResult

AnthonyMDev:
This PR renames `GraphQLResult` to `GraphQLResponse`. "Response" is more accurate to what this actually _is_. 

We used to have an internal type in 1.0 called `GraphQLResponse` which handled parsing the response data into the `GraphQLResult`. 2.0 removes this struct and handles parsing through a `JSONResponseParser` object. This allows us to use the name `GraphQLResponse` where it is more accurate in the public API.

The intermediary struct that is created by the `JSONResponseParser` is now named `ParsedResult`. This struct just holds a `GraphQLResponse` and the cache records to be written to the cache.

---

This PR also changes the new `GraphQLResponse` struct to be generic over the `GraphQLOperation` instead of the operation's `Data` object directly. I think this is more expressive and makes reading the code clearer.


BobaFetters (approved):
LGTM

PR #694, opened 2025-07-01
[v2] Exposed all errors in multipart subscription parser 

AnthonyMDev:
`MultipartResponseSubscriptionParser.ParsingError.irrecoverableError(message: String)` has been changed to `irrecoverableErrors([GraphQLError])`.

This change is due to a user request. When there are multiple errors, they should all be exposed to the user.


BobaFetters (approved):
LGTM

PR #695, opened 2025-07-01
[v2] Return nil on cache miss

AnthonyMDev:
This PR changes the behavior of a cache miss to return `nil` rather than throwing a `JSONDecodingError.missingValue`. A cache miss is an expected possibility, not necessarily an error, and this can cause confusion for users. Now the cache loading mechanisms only throw an error if there is a problem with the cache, such as corrupt or invalid data.

### DIFF
--- a/apollo-ios/Sources/Apollo/RequestChain.swift
+++ b/apollo-ios/Sources/Apollo/RequestChain.swift
@@ -4,19 +4,6 @@ import Foundation
   import ApolloAPI
 #endif
 
-public enum RequestChainError: Swift.Error, LocalizedError {
-  case noResults
-
-  public var errorDescription: String? {
-    switch self {
-    case .noResults:
-      return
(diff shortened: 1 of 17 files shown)


apollo-ios/Sources/Apollo/RequestChain.swift
```diff
@@ -158,16 +145,7 @@ public struct RequestChain<Request: GraphQLRequest>: Sendable {
 
             // Cache miss
 
-          } catch {
-            #warning(
```
AnthonyMDev, line 162: After considering this further, I don't think we should make that change. If the user's cache policy asks for network data on a cache failure, it shouldn't matter if the cache failed do to invalid data or just a cache miss, either way, the correct way to handle that would be to re-fetch the data from the network and re-hydrate the cache with the correct data.

BobaFetters (approved):

PR #698, opened 2025-07-03
[v2] ApolloClientProtocol and Test Deprecation Cleanup

AnthonyMDev:
Removes `ApolloClientProtocol`. This isn't really needed anymore internally and users shouldn't be mocking `ApolloClient` in their tests, they should instead be creating a mock `NetworkTransport`.

Also cleans up tests that were using the deprecated 1.0 fetch functions.


calvincestari:
Looks like some build errors preventing test from running.

PR #699, opened 2025-07-03
[v2] GraphQLOperation.Variable Cleanup

AnthonyMDev:
This removes the conformance of `Optional` to GraphQLOperationVariableValue`.

Any optional operation variables should actually be using `GraphQLNullable` to denote the difference between explictly null and absent values. This prevents us or users from accidentally using `Optional` where we shouldn’t be here.

Also adds back commented out tests for encoding of variable values.


calvincestari:
Same [issue with tests](https://github.com/apollographql/apollo-ios-dev/pull/698#issuecomment-3037046216) as #698 it looks like.

AnthonyMDev:
Tests are fixed in other PRS after this one. Goal of the stack of PRs on top of this is getting CI passing again

PR #537, opened 2024-11-13
feature: @oneOf input object support

BobaFetters:
Adding support for @oneOf Input Objects, more info can be found in the official RFC here: https://github.com/graphql/graphql-spec/pull/825

Closes apollographql/apollo-ios#3430

PR #705, opened 2025-07-10
Add isOneOf to inspection query in ApolloSchemaDownloader

glenngijsberts:
In https://github.com/apollographql/apollo-ios-dev/pull/537 the `@oneOf` directive was introduced. However, I noticed that the directive was not included in my schema when I did an introspection, and so the additional types were not generated. 

I had a look at the source code, and noticed that the `isOneOf` was missing from the inspection query done in `ApolloSchemaDownloader`. I've added it, but I wasn't really able to test this, so I was hoping if somebody can point me in the right direction or help me out with this 🙏

### DIFF
--- a/apollo-ios-codegen/Sources/ApolloCodegenLib/ApolloSchemaDownloader.swift
+++ b/apollo-ios-codegen/Sources/ApolloCodegenLib/ApolloSchemaDownloader.swift
@@ -271,6 +271,7 @@ public struct ApolloSchemaDownloader {
           kind
           name
           description
+          isOneOf
           fields(includeDeprecated: true) {
             name
             description


calvincestari:
Thanks for the PR @glenngijsberts - this seems correct but I'm not sure we can merge it yet since the [`OneOf` spec change PR](https://github.com/graphql/graphql-spec/pull/825) has not been merged yet.

@martinbonnin / @BoD - is this why `isOneOf` is [commented out in the Kotlin introspection query](https://github.com/apollographql/apollo-kotlin/blob/main/libraries/apollo-tooling/src/main/resources/base-introspection.graphql#L59)?

BoD:
Indeed, selecting `isOneOf` on `__Type` will result in an error on servers that don't support it yet - breaking inspection.

To work around this, we implemented "2 step introspection":
- the 1st step is a ["pre-introspection" query](https://github.com/apollographql/apollo-kotlin/blob/main/libraries/apollo-tooling/src/main/graphql/graphql/pre-introspection.graphql) to gather the capabilities of the server
- 2nd step is the actual introspection query, based on info from the 1st step. What we do in Apollo Kotlin is that we use [this query](https://github.com/apollographql/apollo-kotlin/blob/main/libraries/apollo-tooling/src/main/resources/base-introspection.graphql) with certain fields commented, that we programmatically uncomment if the fields are supported.

The logic is [here](https://github.com/apollographql/apollo-kotlin/blob/main/libraries/apollo-tooling/src/main/kotlin/com/apollographql/apollo/tooling/SchemaDownloader.kt#L166).

glenngijsberts:
@calvincestari @BoD makes sense 👍 I'm gonna have another look at this upcoming week, to see if it's possible to do something similar.

calvincestari:
> How could I test this? I'm having a bit trouble understanding how everything works in the project, so any advice would be helpful!

I don't think there are any specific tests required for this - as it is right now. This is because the response is fed into our graphql-js frontend which will validate everything and convert it into SDL. There might be additional tests required for the logic we've identified in the above comments.

PR #711, opened 2025-07-21
Feat expotential backoff

tokizuoh:


apollo-ios/Sources/Apollo/MaxRetryInterceptor.swift
```diff
@@ -51,11 +99,47 @@ public class MaxRetryInterceptor: ApolloInterceptor {
...
+    let retryAttempt = hitCount - 1
+    let exponentialDelay = configuration.baseDelay * pow(configuration.multiplier, Double(retryAttempt))
+    
+    // Apply maximum delay cap
+    let cappedDelay = min(exponentialDelay, configuration.maxDelay)
+    
+    // Apply jitter if enabled to prevent thundering herd problems
+    if configuration.enableJitter {
+      // Full jitter: random value between 0 and calculated delay
+      return TimeInterval.random(in: 0...cappedDelay)
```
BobaFetters, line 140: For the jitter, would it make sense to have it still be relative to the current delay, so if the current current calculated delay is 15s, instead of being from 0s to 15s maybe its like 80-100% of the current delay which would be 12-15s. Just thinking since at any point you have already had retries done in shorter time ranges so maybe there should be a floor relative to the current delay time still for the jitter.

BobaFetters (commented):
@tokizuoh Thanks so much for the contribution, overall I think this looks good, just had one comment on the jitter implementation.

Also want to give time for @AnthonyMDev @calvincestari to provide any input they may have.

apollo-ios/Sources/Apollo/MaxRetryInterceptor.swift
```diff
@@ -51,11 +99,47 @@ public class MaxRetryInterceptor: ApolloInterceptor {
...
-      request: request,
-      response: response,
-      interceptor: self,
-      completion: completion
-    )
+    
+    // Apply exponential backoff delay if enabled and this is a retry (hitCount > 1)
+    if self.configuration.enableExponentialBackoff && self.hitCount > 1 {
+      let delay = calculateExponentialBackoffDelay()
+      DispatchQueue.main.asyncAfter(deadline: .now() + delay) { [weak self] in
```
calvincestari, line 106: I don't remember the queueing strategy for the request chain but we should check that throwing this onto `main` is the right queue to be using.

apollo-ios/Sources/Apollo/MaxRetryInterceptor.swift
```diff
@@ -51,11 +99,47 @@ public class MaxRetryInterceptor: ApolloInterceptor {
...
+        interceptor: self,
+        completion: completion
+      )
+    }
+  }
+  
+  /// Calculates the exponential backoff delay based on current retry attempt.
+  ///
+  /// - Returns: The calculated delay in seconds, including jitter if enabled.
+  private func calculateExponentialBackoffDelay() -> TimeInterval {
```
calvincestari, line 128: Might be a good idea to have tests for the calculated backoff delays since this has the potential to wreck the default interceptor chain.

apollo-ios/Sources/Apollo/MaxRetryInterceptor.swift
tokizuoh: Thanks for the suggestion! I implemented this in 0288e393322f948ae6764c84592ef0707719297a.

I went with 50% as the floor (instead of 80%) based on AWS's equal jitter implementation. The delay now ranges from 50-100% of the calculated delay rather than 0-100%. 

Reference: https://aws.amazon.com/blogs/architecture/exponential-backoff-and-jitter/

apollo-ios/Sources/Apollo/MaxRetryInterceptor.swift
tokizuoh: Thanks for pointing that out. I’m actually unsure which queue fits this pattern best, what do you think?” “I ended up using the main queue, but it’s harder to justify than I expected.

calvincestari: That default `main` queue value on function parameters I believe is only actually used for the error or success callbacks, i.e.: a terminating queue. `kickoff`, `interceptAsync`, `proceedAsync`, or `retry` don't appear to be called on any specific queue. Those in turn are called from a network transport, and ultimately from client app code. So the originating queue is hidden from us.

I agree with @BobaFetters, if we can push this onto the current queue that would be best - I don't like the idea of reaching to Obj-C methods for that but it appears it's all we've got. If that doesn't work then `main` might be the next best choice.

apollo-ios/Sources/Apollo/MaxRetryInterceptor.swift
```diff
@@ -4,7 +4,7 @@ import ApolloAPI
 #endif
 
 /// An interceptor to enforce a maximum number of retries of any `HTTPRequest` with optional exponential backoff support
-public class MaxRetryInterceptor: ApolloInterceptor {
+public class MaxRetryInterceptor: NSObject, ApolloInterceptor {
```
calvincestari, line 7: Adding `NSObject` here makes me question how we can port this change over to 2.0 and it's compatibility with Swift Concurrency?

`NSObject` is not necessarily a problem for `Sendable` because of this:
> [Sendable Classes](https://developer.apple.com/documentation/swift/sendable#Sendable-Classes)
> To satisfy the requirements of the `Sendable` protocol, a class must:
> * Be marked `final`
> * Contain only stored properties that are immutable and sendable
> * Have no superclass or have `NSObject` as the superclass

The bigger problem is that [interceptors in `2.0` are not classes anymore](https://github.com/apollographql/apollo-ios/blob/v2-alpha/Sources/Apollo/Interceptors/MaxRetryInterceptor.swift#L7). We _might_ be saved with all this dancing around though and simply be able to throw this delay logic onto a new subtask.

@AnthonyMDev - thoughts or concerns?

AnthonyMDev: Techincally, interceptors can be any type in 2.0. But `MaxRetryInterceptor` has been converted to be an `actor`.

Regardless, I'm not a fan of using the Obj-C runtime here. It prevents possible Linux compatibility in the future and is not the best for performance. I'd prefer we use a more "Swifty" way of handling this.

Instead of `self.perform(#selector(performRetry), with: nil, afterDelay: delay)`, can we use a `Task`? Check out [this article](https://www.swiftbysundell.com/articles/delaying-an-async-swift-task/) for an example of delaying an async task.

calvincestari: > Why does this need to be performed on the same current queue though? These interceptors are all running asynchronous code, can't they run on whatever queue they decide?

They can but what is a good choice for us here? The only options for doing the work with a `DispatchQueue` is `main` or `global`. When the user spawned the request chain they may have had a specific queue to use but we don't know what that is.

AnthonyMDev: The team has discussed this and have determined that we don't need to worry about what queue the callback is being called on and should just use `DispatchQueue.global(qos: .userInitiated)`. 

Let's make the change back to the original way of handling dispatch here and remove the inheritance from `NSObject`. @tokizuoh would you like to make these changes? I'm happy to go ahead and make the changes myself before merging this in.

Thanks for the contribution!

tokizuoh: @AnthonyMDev
Thank you for the clarification! I've updated the implementation based on the team's decision:

[a8fe846](https://github.com/apollographql/apollo-ios-dev/pull/711/commits/a8fe8462c7f8d6c2c92dafadac45a85bc22fa2be)

The changes are now ready for your review. Please let me know if there's anything else that needs adjustment.

tokizuoh:
@calvincestari Thank you, I'd be glad to! I'll check out the new interceptor pattern in 2.0.0 and work on adapting this feature.

PR #722, opened 2025-07-31
[v2] Int32 as Default Integer Type

AnthonyMDev:
See [#3433](https://github.com/apollographql/apollo-ios/issues/3433) for discussion of this issue.

Changes generated models to use `Int32` instead of `Int` since the GraphQL spec specifically states that Int is a 32 bit signed integer.


calvincestari (approved):
It's a sprawling change but mostly in the generated projects and tests. Looks good - thank you.

PR #735, opened 2025-08-26
feature: @fieldPolicy directive

BobaFetters:
Adds support for the `@fieldPolicy` directive which allows users to configure what cache keys a specific field should use to look for data in the cache before potentially making a network request.

Also adds two new functions to the `SchemaConfiguration` protocol to allow programmatic configuration of field policies.

Note - have a couple of tests I am working on fixing but outside that the code is ready for review

### DIFF
--- a/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/SchemaConfigurationTemplate.swift
+++ b/apollo-ios-codegen/Sources/ApolloCodegenLib/Templates/SchemaConfigurationTemplate.swift
@@ -30,6 +30,7 @@ struct SchemaConfigurationTemplate: TemplateRenderer {
         return nil
       }
     }
+
     """
   }
 }


apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/utilities/fieldPolicyDirective.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  const arg = field.args.find(a => a.name === argName);
+  if (!arg) {
+    throw new GraphQLError(
+      `@fieldPolicy key argument "${keyArg}" does not exist as an input argument of field "${field.name}".`,
+      { nodes: field.astNode }
+    );
+  }
+
+  let curType: GraphQLInputType = base(arg.type);
```
AnthonyMDev, line 194: I'm assuming "cur" is short for current? I'd prefer just using "current" here. I'm fine with using universally commonly used abbreviations, like "arg" for "argument".

apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/utilities/fieldPolicyDirective.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  if (!parts.length || isInputObjectType(curType)) {
+    throw new GraphQLError(
+      `@fieldPolicy key "${keyArg}" must resolve to a leaf input type (scalar/enum), got "${String(curType)}".`,
+      { nodes: field.astNode }
+    );
+  }
+}
+
+function base(type: GraphQLInputType): GraphQLInputType {
```
AnthonyMDev, line 224: This function is confusing. The name `base` tells me nearly nothing about what this does, and even though it's one line, I'm still not clear what the intention is here. Is this just a forced downcast?

apollo-ios-codegen/Sources/IR/IR+Fields.swift
```diff
@@ -12,6 +12,7 @@ public class Field: Equatable, CustomDebugStringConvertible {
   public var responseKey: String { underlyingField.responseKey }
   public var type: GraphQLType { underlyingField.type }
   public var arguments: [CompilationResult.Argument]? { underlyingField.arguments }
+  public var fieldPolicyKeys: [String]? { underlyingField.fieldPolicyKeys }
```
AnthonyMDev, line 15: Can we add a documentation comment here for future reference on what this field is. I think the others are pretty self documenting, but it would be nice to understand specifically that these are the `keyArgs` of a `@fieldPolicy` directive applied to this field.

apollo-ios/Sources/Apollo/ExecutionSources/CacheDataExecutionSource.swift
```diff
@@ -67,6 +68,60 @@ struct CacheDataExecutionSource: GraphQLExecutionSource {
...
+    let key = try info.cacheKeyForField()
+    return object[key]
+  }
+  
+  private func typename(for field: Selection.Field) -> String {
+    switch field.type.namedType {
+    case .object(let selectionSetType):
+      return selectionSetType.__parentType.__typename
+    default:
+      break
```
AnthonyMDev, line 121: Just `return ""` in the default case instead of breaking.

AnthonyMDev: Actually, if the field isn't an object field, we have a serious coding error here, don't we? We shouldn't be continuing on with an empty typename added to the cache key. I think this should be an `assertionFailure` so we catch it early.

apollo-ios/Sources/Apollo/ExecutionSources/CacheDataExecutionSource.swift
```diff
@@ -67,6 +68,60 @@ struct CacheDataExecutionSource: GraphQLExecutionSource {
...
+  
+  private func resolveCacheKey(
+    with info: FieldExecutionInfo,
+    on object: Record
+  ) throws -> AnyHashable? {
+    
+    let fieldTypename = typename(for: info.field)
+    
+    // Programmatic field policy checks
+    switch info.field.type {
```
AnthonyMDev, line 80: This code is messy and hard to read. 

- Programmatic and declarative resolution should be broken out into separate functions. 
- Function calls with multiple params that exceed our standard line length should be properly formatted with each arg on a new line. (We need to set up linter rules eventually... but we do have a `swift-format` config in the project)
- Lines 83-85 are repeated at line 93-95, DRY.
- Formatting of cache key string should be a helper function. It's repeated 6 times here.

apollo-ios/Sources/Apollo/FieldPolicyEvaluator.swift
```diff
@@ -0,0 +1,190 @@
...
+      } else {
+        return ParsedKey(name: key, path: nil)
+      }
+    }
+
+    var fixedParts = [String?](repeating: nil, count: parsed.count)
+    var listIndex: Int? = nil
+    var listValues: [String] = []
+
+    for (i, pk) in parsed.enumerated() {
```
AnthonyMDev, line 43: We try not to use 1 or 2 letter variable names. While reading this code, it took me an addition 15-20 seconds to figure out that `pk` stood for `parsedKey`. Making the code easier for others to quickly understand saves so much time and mental overhead in the long run. Just use `(index, parsedKey)`, or better yet `(index, keyArg)`.

apollo-ios/Sources/Apollo/FieldPolicyEvaluator.swift
```diff
@@ -0,0 +1,190 @@
...
+        return nil
+      }
+
+      guard let resolved = argVal.resolveValue(keyPath: pk.path, variables: variables), !resolved.isEmpty else {
+        return nil
+      }
+      
+      if resolved.count > 1 {
+        listIndex = i
+        listValues = resolved
```
AnthonyMDev, line 54: Add a newline here makes this a lot easier to read to me.
```suggestion
        listValues = resolved
        
```

apollo-ios/Sources/Apollo/FieldPolicyEvaluator.swift
```diff
@@ -0,0 +1,190 @@
...
+      } else {
+        guard let value = resolved.first else {
+          return nil
+        }
+        fixedParts[i] = value
+      }
+    }
+
+    if let idx = listIndex {
```
AnthonyMDev, line 63: It's not entirely clear what the point of this section of the function is unless you already have a lot of context.

Let's rename `idx` to `indexOfListArg` (actually we should rename `listIndex` on line 40 to that as well.
This section should also probably be broken into another function with a descriptive name of what it's doing.

apollo-ios/Sources/Apollo/FieldPolicyEvaluator.swift
```diff
@@ -0,0 +1,190 @@
...
+  
+  func resolveFieldPolicy() -> FieldPolicyResult? {
+    guard let fieldPolicy = field.fieldPolicy,
+          let arguments = field.arguments else {
+      return nil
+    }
+    
+    struct ParsedKey { let name: String; let path: [String]? }
+
+    let parsed: [ParsedKey] = fieldPolicy.keys.map { key in
```
AnthonyMDev, line 29: While this looks fine here, as I'm reading the rest of the code, the mental overhead of translating `parsed` to "this is the list of the keyArgs of the fieldPolicy` is too much. It makes the code hard to read.

```suggestion
    let keyArgs: [ParsedKey] = fieldPolicy.keys.map { key in
```

apollo-ios/Sources/Apollo/FieldPolicyEvaluator.swift
```diff
@@ -0,0 +1,190 @@
...
+      }
+    }
+
+    if let idx = listIndex {
+      var keys: [CacheKeyInfo] = []
+      keys.reserveCapacity(listValues.count)
+      for item in listValues {
+        var parts = [String]()
+        parts.reserveCapacity(parsed.count)
+        for j in 0..<parsed.count {
```
AnthonyMDev, line 69:
```
for index in 0..<keyArgs.count {
   if index == indexOfListArg {
     ...
```

These small naming changes make the code so much easier to read and understand.

apollo-ios/Sources/Apollo/FieldPolicyEvaluator.swift
```diff
@@ -0,0 +1,190 @@
...
+            return nil
+          }
+        }
+        keys.append(CacheKeyInfo(id: parts.joined(separator: "+")))
+      }
+      return .list(keys)
+    } else {
+      let parts = fixedParts.compactMap { $0 }
+      guard parts.count == parsed.count else {
+        return nil
```
AnthonyMDev, line 84: Is there a scenario in which this actually should happen and return `nil`? Or is this actually an error that isn't being thrown?

apollo-ios/Sources/Apollo/FieldPolicyEvaluator.swift
```diff
@@ -0,0 +1,190 @@
...
+    case let floatVal as Float:
+      return String(floatVal)
+    default:
+      return String(describing: self)
+    }
+  }
+}
+
+extension JSONValue {
+  fileprivate func cacheKeyComponentStringValue(keyPath: [String]? = nil) -> [String]? {
```
AnthonyMDev, line 112: We typically have a newline at the end of each switch cases `return`. It makes the code a lot more readable.

apollo-ios/Sources/Apollo/FieldPolicyEvaluator.swift
```diff
@@ -0,0 +1,190 @@
...
+      guard let targetValue = objVal.walk(path: keyPath[...]) else { return nil }
+      return targetValue.cacheKeyComponentStringValue()
+    default:
+      return [String(describing: self)]
+    }
+  }
+}
+
+extension JSONObject {
+  fileprivate func walk(
```
AnthonyMDev, line 139: I don't know what this `walk` function does. The name needs to be more descriptive of the intent here.

apollo-ios/Sources/ApolloAPI/Selection.swift
```diff
@@ -54,16 +57,28 @@ public enum Selection {
       }
     }
   }
+  
+  public struct FieldPolicy: Hashable {
+    public let keys: [String]
```
AnthonyMDev, line 62: I think this should be `keyArgs` not `keys`.

AnthonyMDev (dismissed):

Sources/AnimalKingdomAPI/animalkingdom-graphql/AnimalSchema.graphqls
```diff
@@ -1,5 +1,5 @@
 type Query {
-  allAnimals: [Animal!]!
+  allAnimals(species: String): [Animal!]!
```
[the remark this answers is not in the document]
BobaFetters: nope thought i had removed that already

apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/__tests__/fieldPolicyDirectiveTests.ts
```diff
@@ -0,0 +1,525 @@
...
+
+  type Dog implements Animal {
+    animal(withName: String): Animal!
+
+    id: ID!
+    species: String!
+    owner: String
+  }
+
+  extend interface Animal @fieldPolicy(forField: "animal", keyArgs: "withName")
```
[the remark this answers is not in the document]
BobaFetters: This would throw a validation error (line 189 in the fieldPolicyDirective.ts) because its expected that the field on `Dog` should have the key fields from `Animal` as well

BobaFetters:
Converted this to draft since it should have been that way to begin with, addressed feedback, still need to add a few tests around the `FieldPolicyProvider` and add some more code documentation. Full documentation for doc site will be done in a separate PR

AnthonyMDev (dismissed):

apollo-ios-codegen/Sources/GraphQLCompiler/JavaScript/src/utilities/fieldPolicyDirective.ts
```diff
@@ -0,0 +1,272 @@
...
+
+  if (!parts.length || isInputObjectType(currentType)) {
+    throw new GraphQLError(
+      `@fieldPolicy key "${keyArg}" must resolve to a leaf input type (scalar/enum), got "${String(currentType)}".`,
+      { nodes: field.astNode }
+    );
+  }
+}
+
+function getBaseInputType(type: GraphQLInputType): GraphQLInputType {
```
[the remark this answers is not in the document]
BobaFetters: Verified we do need this, its casting objects to the correct type, so for one of the tests that started failing when this was removed the `currentType` should be `GraphQLInputObjectType` but without this cast it is `GraphQLNonNull` and then our validation using `isInputObject` fails.

apollo-ios/Sources/Apollo/FieldPolicyEvaluator.swift
```diff
@@ -0,0 +1,234 @@
...
+    switch self {
+    case .scalar(let scalar):
+      return [scalar.cacheKeyComponentStringValue]
+    case .variable(let varName):
+      guard let varValue = variables?[varName] else {
+        return nil
+      }
+      return varValue._jsonEncodableValue?._jsonValue.cacheKeyComponentStringValue(keyPath: keyPath)
+    case .list(let list):
+      if list.contains(where: { if case .list = $0 { return true } else { return false } }) {
```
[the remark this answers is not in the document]
BobaFetters: yea that doesn't compile

Tests/ApolloTests/Cache/FieldPolicyTests.swift
```diff
@@ -0,0 +1,1374 @@
+import XCTest
+@testable import Apollo
+import ApolloAPI
+import ApolloInternalTestHelpers
+
+@MainActor
```
AnthonyMDev, line 6: I think in the 1.0 branch this shouldn't be marked `@MainActor`

apollo-ios/Sources/Apollo/FieldPolicyDirectiveEvaluator.swift
```diff
@@ -0,0 +1,234 @@
...
+#if !COCOAPODS
+import ApolloAPI
+#endif
+
+enum FieldPolicyResult {
+  case single(CacheKeyInfo)
+  case list([CacheKeyInfo])
+}
+
+struct FieldPolicyDirectiveEvaluator {
```
AnthonyMDev, line 11: This is some really nice code. Lots of complex logic and edge cases to deal with. Great job here.

AnthonyMDev (approved):

PR #740, opened 2025-09-02
Enable Swift 6 Language Mode

dfed:
Apollo V2 is currently `enableExperimentalFeature` swift settings that are meant for Swift 5.9/5.10. This PR enables Swift 6 Language Mode.

Per [Enabling Complete Concurrency Checking](https://www.swift.org/documentation/concurrency/):

When using Swift 6.0 tools or later, use [SwiftSetting.enableUpcomingFeature](https://developer.apple.com/documentation/packagedescription/swiftsetting/enableupcomingfeature(_:_:)) in the Swift settings for the given target

We _could_ change `enableExperimentalFeature` to `enableUpcomingFeature`, as suggested by the above. But based on local testing, the targets in the `Package.swift` files in this repo already builds in Swift 6 Language Mode.

Feel free to close this out if it's not what you want just yet – I know I'm jumping the gun going straight to a PR here.

### DIFF
--- a/apollo-ios-codegen/Package.swift
+++ b/apollo-ios-codegen/Package.swift
@@ -33,7 +33,7 @@ let package = Package(
         .product(name: "InflectorKit", package: "InflectorKit"),
         .product(name: "OrderedCollections", package: "swift-collections")
       ],
-      swiftSettings: [.enableUpcomingFeature("ExistentialAny")]
+      swiftSettings: [.swiftLanguageMode(.v6)]
     ),
     .target(
       name: "GraphQLCompiler",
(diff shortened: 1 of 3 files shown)


AnthonyMDev:
I actually had done this the other day locally, and then reverted it because I couldn't find any clear documentation of how this affected compatibility with Swift 5 code. Maybe you are more familiar with this than I am.

If we merge this PR and enable Swift 6 language mode, will the library still be able to be linked to user's code if they are using Swift 5? My understanding is that the answer is yes, but I haven't been able to find a definitive answer to that question, so I was hesitant to make the switch.

dfed:
> If we merge this PR and enable Swift 6 language mode, will the library still be able to be linked to user's code if they are using Swift 5? My understanding is that the answer is yes, but I haven't been able to find a definitive answer to that question, so I was hesitant to make the switch.

Yes! Swift 5 can depend on Swift 6 code, and visa versa. The bigger constraint is your `swift-tools-version`: by setting `swift-tools-version:6.1`, you are dropping support for Xcode versions that do not have the Swift 6.1 compiler available (Xcode 16.2 and earlier). My guess is that's intentional though.

PR #748, opened 2025-09-09
Add RootSelectionSet convenience initializer from inline fragment

AnthonyMDev:
[Discussion on the reasoning for this PR is here.](https://community.apollographql.com/t/initializing-selection-set-manually-without-using-datadict-initializer/9395/8) 

Added a convenience initializer to `RootSelectionSet` that allows it to be initialized directly from an `InlineFragment`, simplifying the conversion process between fragments and root entity types.


calvincestari (approved):
LGTM!

PR #750, opened 2025-09-10
Field Policy Documentation

BobaFetters:
Adding documentation for the field policy feature.

### DIFF
--- a/docs/source/caching/cache-key-resolution.mdx
+++ b/docs/source/caching/cache-key-resolution.mdx
@@ -87,3 +87,161 @@ This is possible if:
   A) A type is added to your schema after code generation is executed and that type implements two different interfaces with different type policies
 
   B) Implementation of an interface is added to an existing type which already implements an interface with a different type policy
+
+## The `@fieldPolicy` directive
+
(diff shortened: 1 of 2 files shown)


docs/source/caching/cache-key-resolution.mdx
```diff
@@ -87,3 +87,161 @@ This is possible if:
...
+extend type Query @fieldPolicy(forField: "allAnimals", keyArgs: "habitat species")
+```
+
+#### List Arguments
+
+The `@fieldPolicy` directive does support fields which return a list. There are a few caveats with the usage of lists:
+
+- If a field has a list return type, it must contain exactly one list input parameter.
+- Conversely if a list contains a list input parameter, it must return a list.
+- Fields cannot have more than one list input parameter.
```
calvincestari, line 159: Should this be "List fields cannot..."?

calvincestari: It's also not really necessary since the first point states "...must contain exactly one list input parameter".

AnthonyMDev (dismissed):
Nice work. I want to add a bit more explanation of what a field policy is. Made a few suggestions.

The directive usage explanation and examples are great! Very clear and exhaustive with easy to follow examples.

AnthonyMDev (approved):

BobaFetters:
@AnthonyMDev I will double check, but I think there were multiple AI comments about missing punctuation but the sentence did have a period at the end already lol

BobaFetters:
Nevermine looks like some new AI comments showed up since the last time I checked I see the one that's missing a period.

PR #755, opened 2025-09-16
Improvement: Add result code to SQLiteError

BobaFetters:
Adding the raw sqlite result code as an associated value on enum cases of SQLiteError


calvincestari (approved):
Thank you, looks good!

PR #758, opened 2025-09-18
Make dependentKeys in GraphQLResult public.

chrsmys:
## Summary
Making `dependentKeys` public allows for implementing custom `ApolloStoreSubscribers` similar to `GraphQLQueryWatcher`. Without this it's extremely to know when to react to cache changes for a query.

### DIFF
--- a/apollo-ios/Sources/Apollo/GraphQLResult.swift
+++ b/apollo-ios/Sources/Apollo/GraphQLResult.swift
@@ -20,7 +20,7 @@ public struct GraphQLResult<Data: RootSelectionSet> {
   /// Source of data
   public let source: Source
 
-  let dependentKeys: Set<CacheKey>?
+  public let dependentKeys: Set<CacheKey>?
 
   public init(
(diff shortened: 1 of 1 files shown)


apollo-ios/Sources/Apollo/GraphQLResult.swift
```diff
@@ -20,7 +20,7 @@ public struct GraphQLResult<Data: RootSelectionSet> {
   /// Source of data
   public let source: Source
 
-  let dependentKeys: Set<CacheKey>?
+  public let dependentKeys: Set<CacheKey>?
```
AnthonyMDev, line 23:
```suggestion
  /// The cache keys for the fields that were included in this response. Custom ``ApolloStoreSubscriber``s can use these
  /// keys to understand when changes to the cache would affect the result of a specific response.
  public let dependentKeys: Set<CacheKey>?
```

AnthonyMDev (approved):

PR #760, opened 2025-09-18
Update documentation

ScottPlease:
Not sure which branch is best to use as a base; please advise if v2-staging is not correct. 
Updated watcher docs to read better. 
Updated WebSocketTransport Docs to reflect the expectations outlined in [this comment.](https://github.com/apollographql/apollo-ios/issues/3411#issuecomment-3308589570)

### DIFF
--- a/apollo-ios/Sources/ApolloWebSocket/WebSocketTransport.swift
+++ b/apollo-ios/Sources/ApolloWebSocket/WebSocketTransport.swift
@@ -5,8 +5,9 @@ import ApolloAPI
 public final class WebSocketTransport: SubscriptionNetworkTransport {
 
   public enum Error: Swift.Error {
-    /// WebSocketTransport has not yet been implemented for Apollo iOS 2.0. This will be implemented in a future
-    /// release.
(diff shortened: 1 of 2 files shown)


apollo-ios/Sources/ApolloWebSocket/WebSocketTransport.swift
```diff
@@ -5,8 +5,9 @@ import ApolloAPI
 public final class WebSocketTransport: SubscriptionNetworkTransport {
 
   public enum Error: Swift.Error {
-    /// WebSocketTransport has not yet been implemented for Apollo iOS 2.0. This will be implemented in a future
-    /// release.
+    /// WebSocketTransport has not yet been implemented for Apollo iOS 2.0. This will be implemented
+    /// in a 2.1 update as soon as we can after 2.0 is out. Subscriptions over HTTP will work via the
```
calvincestari, line 9: I don't believe there is any guarantee this will be in a 2.1 release which is why we used the "future release" terminology.

ScottPlease: Okay, I'll revise this to future release.

PR #762, opened 2025-09-19
chore: Adds missing inline documentation

calvincestari:
Adding missing inline documentation, re. https://github.com/apollographql/apollo-ios/issues/3586#issuecomment-3313439457.


calvincestari:
Merging without review since it's just an inline comment for API documentation.

PR #786, opened 2025-10-02
Update API Reference to default theme

AnthonyMDev:
Update Docc Renderer to use the default docc renderer for Swift. Removing our custom renderer and theming.


AnthonyMDev:
This isn't working correctly. I'll try something else eventually.

PR #986, opened 2026-05-19
Fix SelectionSet equality crash on nullable-inner object lists

dfed:
## Summary

`SelectionSet+Equatable.convertElements` traps with `preconditionFailure("Expected list data to contain objects.")` whenever a list field declared as `[Object?]` (non-null outer, nullable inner) contains an actual `null` element. This patch adds a `list as? [DataDict?]` branch that preserves null slots so the equatable walker no longer crashes — and so position-sensitive equality is still correct.

## Mechanism

When the backend returns a partial GraphQL response — e.g. one resolver inside a list throws and GraphQL's null-propagation rules bubble the error up only to the nearest nullable parent (the list element) — the iOS executor materializes the JSON `null` element as `NSNull()` via `DataDictMapper.acceptNullValue`. The list stored at `__data._data[field.responseKey]` is therefore heterogeneous: a mix of `DataDict` values and `NSNull`.

Then during `SelectionSet ==`:

1. `fieldsForEquality()` walks each fragment's fields.
2. `add(field:)` reaches the object-list branch and casts `fieldData as? [DataDict.FieldValue]` — this succeeds because both `DataDict` and `NSNull` are `Hashable & Sendable`.
3. `convertElements(of:to:)` then tries `list as? [DataDict]` — fails the moment any element is `NSNull`.
4. Falls through to `list as? [[FieldValue]]` — fails (not nested arrays).
5. `preconditionFailure("Expected list data to contain objects.")` fires.

Process dies. In our production app this surfaces as repeated crash bursts correlated 1:1 with backend disturbances (e.g. queue purges, transient DB timeouts) on any user that happens to be on a screen that calls `removeDuplicates()` on a fragment containing an `[Object?]` list.

## The fix

Add a third cast that handles the nullable-inner case:

```swift
if let nullableList = list as? [DataDict?] {
  return nullableList.map { dataDict -> DataDict.FieldValue in
    guard let dataDict else { return NSNull() }
    return selectionSetType.init(_dataDict: dataDict)
  }
}
```

`NSNull` bridges to `Optional.none` when cast to `DataDict?`, so this branch catches mixed `DataDict + NSNull` lists. Null slots are preserved as `NSNull()` so two SelectionSets with nulls in the same positions still compare equal, while shifting a null position correctly reports inequality. Filtering nulls would have been simpler but would break position-sensitive equality (`[a, null, b]` would compare equal to `[a, b]`, which is wrong).

The existing `[DataDict]` and `[[FieldValue]]` paths are unchanged so the all-present and nested-list fast paths keep their single-cast performance.

## Tests

`Tests/ApolloTests/SelectionSet_EquatableTests.swift` gains three regression tests next to the existing list-of-objects coverage:

- `test__equatable__listOfNullableObjectsField__allPresent_returns_true` — baseline; passes pre-fix (no null elements ⇒ `[DataDict]` cast succeeds).
- `test__equatable__listOfNullableObjectsField__withNullElement_sameValues_returns_true` — **reproduces the crash**: `Fatal error: Expected list data to contain objects.` at `apollo-ios/Sources/ApolloAPI/SelectionSet+Equatable.swift:153` prior to the fix. Passes post-fix.
- `test__equatable__listOfNullableObjectsField__differentNullPositions_returns_false` — verifies position sensitivity: shifting a null to a different slot reports inequality. Also crashed pre-fix.

To see the crash for yourself, revert the new branch in `convertElements` and run:

```
tuist test ApolloTests --test-targets "ApolloTests/SelectionSet_EquatableTests/test__equatable__listOfNullableObjectsField__withNullElement_sameValues_returns_true"
```

## Production impact

In our production app, this was the cause of Sentry issue APPLE-HGN (and three sibling groupings split by minor stack-frame variance): ~90 events / ~50 users / ~6 weeks of crashes. The triggering field in our schema is a `[ChatGateOption]!` whose inner type is nullable; under any backend disturbance that nulls a single option, the entire iOS process dies during the next `removeDuplicates()` cycle on the user-tree publisher. The same pattern is latent on every `[Object?]` field we ship — this PR closes that latent bug for everyone.

### DIFF
--- a/Tests/ApolloTests/SelectionSet_EquatableTests.swift
+++ b/Tests/ApolloTests/SelectionSet_EquatableTests.swift
@@ -1191,6 +1191,173 @@ class SelectionSet_EquatableTests: XCTestCase {
     expect(selectionSet1.hashValue).to(equal(selectionSet2.hashValue))
   }
 
+  func test__equatable__listOfNullableObjectsField__allPresent_returns_true() {
+    // given
+    final class HumanFragment: MockFragment, @unchecked Sendable {
+      typealias Schema = MockSchemaMetadata
+
+      override class var __parentType: any ParentType { Types.Human }
+      override class var __selections: [Selection] {[
+        .field("items", [Item?].self)
+      ]}
+
+      override class var __fulfilledFragments: [any SelectionSet.Type] {
+        [HumanFragment.self]
+      }
+
+      final class Item: MockSelectionSet, @unchecked Sendable {
+        typealias Schema = MockSchemaMetadata
+
+        override class var __parentType: any ParentType { Types.Item }
+        override class var __selections: [Selection] {[
+          .field("id", Int.self)
+        ]}
+        override class var __fulfilledFragments: [any SelectionSet.Type] {
+          [Item.self]
+        }
+      }
+    }
+
+    let item = DataDict(data: [
+      "__typename": "Item",
+      "id": 1,
+    ], fulfilledFragments: [
+      ObjectIdentifier(HumanFragment.Item.self)
+    ])
+
+    // when
+    let selectionSet1 = HumanFragment(_dataDict: DataDict(
+      data: [
+        "__typename": "Character",
+        "items": [item, item] as [DataDict?]
+      ],
+      fulfilledFragments: HumanFragment.__fulfilledFragmentIds
+    ))
+
+    let selectionSet2 = HumanFragment(_dataDict: DataDict(
+      data: [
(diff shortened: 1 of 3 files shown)


AnthonyMDev:
Thanks for this PR and for bringing the issue to our attention. The solution provided only handled 2D lists. I've modified it so that this can handle lists of any level of depth. I'll be accepting and merging this now. Will go out with the next release.

AnthonyMDev (approved):

PR #1001, opened 2026-05-27
feat(sqlite): row-per-element CRUD against position-keyed schema (PR-009)

AnthonyMDev:
## Goal

Land the amended PR-009 scope per [ADR 0006](apollo-ios/Design/adr/0006-list-storage-strategy.md): row-per-element write + delete CRUD against the position-keyed schema introduced in PR-008b (#1005). The JSON `list_value` path is gone; nested lists recurse via synthetic sub-records at `<parent>.<field>.$[<position>]`.

**Cascading deletion of synthetic sub-records is split into a follow-up PR** with extensive cascade-correctness tests. This PR establishes the foundation; the cascade PR ships the cleanup walks.

### Public API on `SQLiteDatabase`

| Method | Behavior |
|---|---|
| `insertOrUpdate(records: [Record])` | For each field: scalar → one row at `position = -1`; list → N rows at positions `0..N-1`. Each write atomically clears the field's prior rows before inserting the new ones, so list rewrites and scalar↔list transitions never leave in-field orphans. |
| `deleteRecord(forKey:)` | Direct `DELETE WHERE cache_key = ?`. Does not cascade synthetic sub-records (see below). |
| `deleteRecords(matchingKey:)` | `LIKE … COLLATE NOCASE ESCAPE '\\'` with `\`, `%`, `_` in the user-supplied pattern escaped so literal substrings don't act as wildcards. Does not cascade synthetic sub-records. |

### Synthetic sub-records for nested lists

Per ADR 0006 + the design conversation:

- Outer-list rows for a list-of-list field hold `child_key_value = "<parent>.<field>.$[N]"` pointing at a synthetic sub-record.
- The sub-record's element rows live under `field_name = "$"` (sentinel — `$` can't start a GraphQL Name, so no collision with real fields).
- Deeper nesting (`[[[T]]]`) appends `.$[N]` per level: `User:1.matrix3d.$[0].$[0].$[0]` is the level-3 sub-record's key.
- Detection regex: `\.\$\[[0-9]+\]$`.

### Encoder (`SQLiteFieldEncoding`)

Per-value, never per-array. Lists are exploded into rows at the database layer.

| Swift value | Column |
|---|---|
| `NSNumber` (boolean via `CFGetTypeID == CFBooleanGetTypeID`) | `bool_value` |
| `NSNumber` (integer kinds via `CFNumberGetType`) | `int_value` |
| `NSNumber` (other numeric kinds) | `float_value` |
| `String` | `string_value` |
| `CacheReference` | `child_key_value` (stores `.key`) |
| `NSNull` and any other JSON-shaped value | `custom_scalar_value` (`.sortedKeys + .fragmentsAllowed` JSON) |
| `[Record.Value]` | precondition violation — caller must explode it |

The custom-scalar fallback gates on `JSONSerialization.isValidJSONObject([value])` before encoding so an unsupported type surfaces a Swift error rather than an `NSException` abort. The `$reference` dict-wrapper is recognized only when the dict has *exactly one* key, so a legitimate custom scalar that happens to include `$reference` alongside other keys round-trips intact.

### Test-only read path

A non-public `internal selectRecords(forKeys:)` reassembles records from rows and resolves synthetic `child_key_value` pointers to materialize nested lists. Used by the test suite to verify what `insertOrUpdate` wrote. The production read path is a projection-aware API to be introduced in PR-009b–h per ADR 0007 — not on the public protocol.

### Tests

`SQLiteRowPerElementCRUDTests.swift` (new file) — **32 tests, all passing**:

- Per-type round-trip ×6 (string, int, double, bool ×2, cache reference)
- Lists: scalars, cache references, mixed scalars, empty
- Nested lists: 2D, 3D, list-of-list-of-references
- Multi-field record preserves all fields including `writtenAt`
- UPSERT: writing twice overwrites; changing value type clears prior typed column
- Atomic list rewrite shrinks length cleanly (no orphan rows for the field itself)
- Delete by exact key: removes only matching rows; non-existent key is no-op
- Delete by pattern: matches prefixes, empty pattern is no-op, `_` and `%` escaped
- Transactional rollback: scalar encoding failure, list-element encoding failure
- Custom-scalar hardening: NSNull top level, generic dict, `$reference` with extra keys treated as generic dict
- NSNumber bool/int/double routing keeps each value's type identity
- Empty record produces no rows (documented behavior)

## What's deferred to the follow-up cascade PR

Three places where synthetic sub-record orphans can accumulate under this PR's behavior:

1. **`deleteRecord(forKey:)`** — deletes the record's own rows but leaves any synthetic sub-records the record's nested-list fields pointed at. They become unreachable but persistent.
2. **`insertOrUpdate` atomic-rewrite over a nested-list field** — clears the field's own rows but not the synthetic sub-records the prior rows pointed at. They orphan.
3. **`deleteRecords(matchingKey:)` + synthetic sub-records** — the flat `LIKE` filter may or may not match synthetic sub-records depending on the pattern. Behavior is currently ambiguous; the cascade PR pins it down.

The cascade PR will:

- Add `cascadeDeleteFromCacheKey` and `cascadeDeleteFromField` CTEs (and a `LIKE`-pattern variant for `deleteRecords(matchingKey:)`).
- Wire them into all three call sites above.
- Add a test-only helper that **bypasses the synthetic-key filter in `selectRecords`** so orphan detection actually verifies the database state (the previous cascade tests on this PR's branch passed *regardless of cascade behavior* because `selectRecords` filtered synthetic keys before the assertions ran — a real test bug surfaced during review).
- Ship 8+ cascade-correctness tests:
  - depth-1 / depth-2 / depth-3 cascade
  - cascade isolation: deleting record A doesn't disturb record B's synthetic sub-records
  - multiple list-typed fields on one record → each field's synthetic sub-records all cascade
  - mixed list contents (synthetic + real `CacheReference`) — cascade follows synthetic only
  - real `CacheReference` *inside* a synthetic sub-record — that ref's target NOT cascaded
  - atomic-rewrite cascade for nested → scalar transition
  - atomic-rewrite cascade for nested → differently-typed nested
  - `deleteRecords(matchingKey:)` + synthetic interaction (with a pinned-down behavior)

The deferred cascade and its tests don't gate this PR's correctness for the operations it ships — `selectRecords` never returns orphan synthetic sub-records as top-level records, so the read surface is unaffected. The cascade is about storage hygiene + the eventual `clearDatabase` cleanliness contract.

## Acceptance criteria

- [x] Encoder + CRUD + tests compile.
- [x] All 32 new tests pass.
- [x] Full `Apollo-UnitTestPlan` is green.
- [x] Zero navigator errors.
- [x] `selectRecords` is not on the public protocol — only an `internal` helper.
- [x] Doc comments on `insertOrUpdate` and `deleteRecord(forKey:)` clearly call out the deferred cascade.

## Stacks on

`cache-rewrite/phase-1-plan` at `e06a2b859` (the PR-008b merge commit).

## Followup

1. **(next)** Cascading-delete PR — adds the cascade walks + extensive cascade-correctness tests.
2. Per ADR 0007's sub-phase 1A.5: PR-009b (`FieldProjection` types), PR-009c–h.

### DIFF
--- a/Tests/ApolloTests/SQLiteRowPerElementCRUDTests.swift
+++ b/Tests/ApolloTests/SQLiteRowPerElementCRUDTests.swift
@@ -1,4 +1,5 @@
 import XCTest
+import Nimble
 @testable @_spi(Execution) import Apollo
 @testable import ApolloSQLite
 import ApolloInternalTestHelpers
@@ -149,12 +150,57 @@ class SQLiteRowPerElementCRUDTests: XCTestCase {
     try db.insertOrUpdate(records: [original])
 
     let loaded = try db.selectRecords(forKeys: ["User:1"])
-    // An empty list writes zero element rows. With no scalar row and
-    // no list rows, the field is indistinguishable from "never
-    // written" — the documented behavior of the row-per-element
-    // schema. The record itself has no fields and is therefore not
-    // present in the result set.
-    XCTAssertEqual(loaded.count, 0)
+    // An empty list writes a single marker row (`position = -2`) so
+    // it stays distinguishable from a never-written field:
+    // `colors: []` is cached, valid data — not a cache miss.
+    expect(loaded).to(haveCount(1))
+    let reloaded = loaded[0].fields["colors"]?.value as? [Any]
+    expect(reloaded).toNot(beNil())
+    expect(reloaded).to(beEmpty())
+    expect(loaded[0].fields["colors"]?.writtenAt).to(equal(100))
+  }
+
+  func test__insertOrUpdate__overwritingNonEmptyListWithEmptyList__readsBackEmptyList() throws {
+    let db = try makeDatabase()
+    let full: [Record.Value] = ["red", "green"]
+    try db.insertOrUpdate(records: [Record(
+      key: "User:1",
+      fields: ["colors": CachedField(value: full as Record.Value, writtenAt: 100)]
+    )])
+
+    let empty: [Record.Value] = []
+    try db.insertOrUpdate(records: [Record(
+      key: "User:1",
+      fields: ["colors": CachedField(value: empty as Record.Value, writtenAt: 200)]
+    )])
+
+    let loaded = try db.selectRecords(forKeys: ["User:1"])
+    expect(loaded).to(haveCount(1))
+    let reloaded = loaded[0].fields["colors"]?.value as? [Any]
+    expect(reloaded).toNot(beNil())
+    expect(reloaded).to(beEmpty())
+    expect(loaded[0].fields["colors"]?.writtenAt).to(equal(200))
+  }
+
+  func test__insertOrUpdate__overwritingEmptyListWithNonEmptyList__readsBackElements() throws {
+    let db = try makeDatabase()
+    let empty: [Record.Value] = []
+    try db.insertOrUpdate(records: [Record(
+      key: "User:1",
+      fields: ["colors": CachedField(value: empty as Record.Value, writtenAt: 100)]
+    )])
+
+    let full: [Record.Value] = ["red", "green"]
+    try db.insertOrUpdate(records: [Record(
+      key: "User:1",
+      fields: ["colors": CachedField(value: full as Record.Value, writtenAt: 200)]
+    )])
+
+    let loaded = try db.selectRecords(forKeys: ["User:1"])
+    let reloaded = loaded[0].fields["colors"]?.value as? [Any]
+    expect(reloaded).to(haveCount(2))
+    expect(reloaded?[0] as? String).to(equal("red"))
+    expect(reloaded?[1] as? String).to(equal("green"))
   }
 
   // MARK: - Nested lists (2D, 3D)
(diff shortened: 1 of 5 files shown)


apollo-ios/Sources/ApolloSQLite/ApolloSQLiteDatabase.swift
```diff
@@ -308,6 +308,453 @@ public final class ApolloSQLiteDatabase: SQLiteDatabase {
...
+  /// Issues one UPSERT against the records table for a single row.
+  /// `ON CONFLICT (cache_key, field_name, position) DO UPDATE` copies
+  /// every value column from `excluded.*`, so a field changing type
+  /// (e.g. Int → String at the same position) also clears its prior
+  /// column.
+  private func upsertRow(
+    cacheKey: CacheKey,
+    fieldName: String,
+    position: Int64,
+    column: SQLiteFieldEncoding.Column,
```
AnthonyMDev: Calling this parameter column is confusing to me. What about `typedValue`?

apollo-ios/Sources/ApolloSQLite/ApolloSQLiteDatabase.swift
```diff
@@ -308,6 +308,453 @@ public final class ApolloSQLiteDatabase: SQLiteDatabase {
...
+    }
+  }
+
+  // MARK: - Row-per-element read assembly (test-only)
+
+  /// Loads all rows for a batch of cache keys and groups them back
+  /// into `Record` instances. Follows synthetic `child_key_value`
+  /// pointers to materialize nested lists.
+  private func loadRecordBatch(_ keys: Set<CacheKey>) throws -> [Record] {
+    let placeholders = keys.map { _ in "?" }.joined(separator: ", ")
```
AnthonyMDev: Can we use a `repeating` count to do this instead of mapping? Should be faster.

apollo-ios/Sources/ApolloSQLite/ApolloSQLiteDatabase.swift
```diff
@@ -308,6 +308,453 @@ public final class ApolloSQLiteDatabase: SQLiteDatabase {
...
+      guard let cacheKeyPtr = sqlite3_column_text(stmt, 0) else {
+        throw SQLiteError.step(message: "Unexpected NULL cache_key in row-per-element select", resultCode: SQLITE_NOMEM)
+      }
+      guard let fieldNamePtr = sqlite3_column_text(stmt, 1) else {
+        throw SQLiteError.step(message: "Unexpected NULL field_name in row-per-element select", resultCode: SQLITE_NOMEM)
+      }
+      let cacheKey = String(cString: cacheKeyPtr)
+      let fieldName = String(cString: fieldNamePtr)
+      let position = sqlite3_column_int64(stmt, 2)
+      let value = try SQLiteFieldEncoding.decode(
```
AnthonyMDev: Shouldn't we know the expected value type of the record here? Would it be more performant if we did and then we didn't need to try to fetch all of the columns that are nil? We could just fetch the column we expect to have a value.

Tests/ApolloTests/SQLiteRowPerElementCRUDTests.swift
```diff
@@ -0,0 +1,553 @@
...
+  private func totalRowCount(_ db: ApolloSQLiteDatabase) throws -> Int {
+    let dbURL = SQLiteTestCacheProvider.temporarySQLiteFileURL()
+    _ = dbURL // unused; rowCount runs via the internal-test select helper
+    // The cheapest way without exposing more internals is to load
+    // every known key and inspect — but we don't know all keys.
+    // Instead, use clearDatabase as a destructive end-of-test check
+    // where appropriate. For row-count assertions, the tests
+    // explicitly select the keys they wrote and check the returned
+    // structure.
+    fatalError("Use selectRecords to assert structure; this helper is unused")
```
AnthonyMDev: If unused, why is this not deleted?

PR #1011, opened 2026-06-02
Fix WebSocketTransport retain cycle in receive loop

heltoft:
This pull request fixes a retain cycle in `WebSocketTransport` that prevents it from ever being deallocated once a connection is established. The receive loop Task in `startConnectionReceiveLoop()` captures `self` strongly, keeping the actor alive even after all external references are dropped.

The connection needs to be closed manually in deinit because the receive loop Task outlives the transport. Even with `[weak self]`, the Task still holds the connection stream and keeps iterating it. Closing the connection ends the stream and lets the Task exit.

Includes a test that verifies the transport deallocates after a subscription completes and all references are dropped.


AnthonyMDev:
Thanks so much for this fix! I've added a few additional unit tests to this, but it otherwise looks good!

AnthonyMDev (approved):

PR #1024, opened 2026-06-10
Chore: Xcode 27b1 workaround

Brett-Best:
I'm not 100% happy with the changes here but they unblock testing for me.

Some of the challenges I ran into once I got the source + tests to compile was the issue where Xcode 27 in the test bundles would complain that two copies of types came from different sources -- `tuist generate` has warnings around this.

Posting as draft so that other people can potentially use it and perhaps the maintainers can take some of the changes as inspiration for proper fixes.


AnthonyMDev:
Thanks for the contribution! This has been addressed by #1034 which includes a comprehensive fix for Xcode 27 / Swift 6.4 compatibility covering this and several related issues. Closing in favor of that PR.

PR #1037, opened 2026-06-25
Adds opt-in `additionalCapitalizationRules` to codegen so acronyms can be fully capitalized

### LINKED ISSUE #3419: codegen: Add an option to capitalize acronyms when the first character is capitalized
### Use case

We use Nick Lockwood's excellent SwiftFormat tool, which [includes an `acronym` rule](https://github.com/nicklockwood/SwiftFormat/blob/main/Rules.md#acronyms) that we'd like to use in our codebase that states the following: 

```
Capitalize acronyms when the first character is capitalized.

Acronyms to auto-capitalize. Defaults to "ID,URL,UUID"
```

Currently, Apollo's codegen defaults to capitalising the first letter of an acronym, and then lowercasing the rest.

### Describe the solution you'd like

We'd like to be able to configure Apollo's code generation to output acronyms in a format where the entire acronym is uppercased. 

Thanks!

3redrubies:
Implements the API design agreed in #463; closes [apollographql/apollo-ios#3419](https://github.com/apollographql/apollo-ios/issues/3419).

## Design
- New public `CapitalizationRule` (`Term` = .string/.regex, `CaseStrategy` = .upper/.lower).
- Follows SwiftFormat semantics: an acronym is only capitalized when its first
  character is already uppercase (so `apiKey`/`idToken` are preserved; `userId`→`userID`).
- Applies to: field accessors, initializer params, enum cases, input fields, test mocks.

## Intentional scope
- `.upper`/`.lower` only (titlecase/camelcase/pascalcase from #463; keeping it focused).
- Type names are intentionally out of scope (would require threading through the core
  type-renderer; cosmetic-only, better as a follow-up). Happy to add if you'd prefer.

## Tests
- Unit tests for the capitalizer + a rendering-integration layer.
- Opt-in/default-empty, so existing generated output is unchanged and won't effect performance


apollo-ios-codegen/Sources/ApolloCodegenLib/Capitalizer.swift
```diff
@@ -0,0 +1,213 @@
...
+
+  /// Splits a camelCase string into word segments.
+  ///
+  /// Examples:
+  /// - `"userId"` → `["user", "Id"]`
+  /// - `"imageURL"` → `["image", "URL"]`
+  /// - `"userID"` → `["user", "ID"]`
+  /// - `"XMLParser"` → `["XML", "Parser"]`
+  /// - `"id"` → `["id"]`
+  private func splitCamelCase(_ string: String) -> [String] {
```
3redrubies, line 159: @AnthonyMDev this should hopefully alleviate the [concerns](https://github.com/apollographql/apollo-ios-dev/pull/463#discussion_r1722344530) around false matching against the letters `id` inside `isHidden` (Eg. `isHIDden`). This split them into segments by camel case and only capitalise segments after the first index.

AnthonyMDev:
Thanks for bringing this PR back up! This looks like a great start. In order to ship this, we'll need a few things.

### 1. There is a bug when a rule renames an identifier used as a GraphQL default value.
We aren't currently using the `Capitalizer` when rendering default values, so they won't match the casing and will fail to compile. We need to update this in `InputVariableRenderable`, but there may be other places we're missing. Which is why I think we need to also add:

### 2. Comprehensive unit tests
Not just tests for the Capitalizer's behavior, but end-to-end testing of the rendering to ensure that everything matches up.

### 3. Error handling
- Invalid regex silently no-ops (returns false). A typo'd pattern produces silently wrong output instead of a config error.
- I'd like to add the other case strategies at some point, but at the very least, I think we should have the original `replace` strategy that I suggested. Currently, there would be no way to do a capitalization rule for `GraphQL`.

### 4. Documentation updates
Inline docs are good, but we need to add this to the codegen configuration documentation reference for the docs site.

---

I can pretty easily have Claude pull the PR and take a pass at these things, but if you'd like to do it yourself and finish the contribution, that's fine too! Let me know.

3redrubies:
Hey @AnthonyMDev, thank you for the feedback. 

> ### 1. There is a bug when a rule renames an identifier used as a GraphQL default value.
> We aren't currently using the `Capitalizer` when rendering default values, so they won't match the casing and will fail to compile. We need to update this in `InputVariableRenderable`...

Updated `InputVariableRenderable.renderVariableDefaultValue(config:)` and `GraphQLInputObjectType.renderInitializer(values:config:)` to take `ApolloCodegen.ConfigurationContext` so the renderer has context about the capitalization config.

> ### 2. Comprehensive unit tests
> Not just tests for the Capitalizer's behavior, but end-to-end testing of the rendering to ensure that everything matches up.

Added end to end tests for rendering including the ones that render default values with the rules applied (to cover the issue raised above). 

> ### 3. Error handling
> * Invalid regex silently no-ops (returns false). A typo'd pattern produces silently wrong output instead of a config error.

Good call, added a `throw` in `ConfigurationContext` and testing around this so it fails upfront instead of silently.

> * I'd like to add the other case strategies at some point, but at the very least, I think we should have the original `replace` strategy that I suggested. Currently, there would be no way to do a capitalization rule for `GraphQL`.

Added `.replace(String)` so can do custom capitalization like `GraphQL`.

> ### 4. Documentation updates
> Inline docs are good, but we need to add this to the codegen configuration documentation reference for the docs site.

Added `additionalCapitalizationRules` to the docs.

> I can pretty easily have Claude pull the PR and take a pass at these things, but if you'd like to do it yourself and finish the contribution, that's fine too! Let me know.

Totally happy for you to have Claude take it from here for any further tweaks — please flag anything you'd rather I pick up 🙂

AnthonyMDev:
Thanks so much for all the work on this! Merging now. :)

PR #1040, opened 2026-07-03
Fix apollo-ios-cli initialization command in documentation

dmytrolux:
The current documentation specifies --module-type swiftPackage, which causes an error during apollo-ios-cli initialization because the correct value is swiftPackageManager. I have updated the command to reflect the correct parameter value required by the CLI.


dmytrolux:
Closing this PR as I realized the initialization command is actually correct for Apollo iOS v2. My confusion stemmed from the Introduction page of the tutorial, which still states that the tutorial is built for Apollo SDK 1.2.0. Because of that outdated page, I installed an older version where the v2 commands obviously didn't work.

My apologies for the noise! I've left my other PR open, which fixes the GraphQLResponse terminology.

PR #1046, opened 2026-07-09
Fix(doc): Add missing 'await' keywords in bookOrCancel() method

dmytrolux:
Update trip booking logic to use aThis PR fixes a Swift Concurrency compilation error in the `bookOrCancel()` method inside the tutorial code. 

Since both `cancelTrip(with:)` and `bookTrip(with:)` are declared as `async` functions, they must be awaited when called. When using a ternary conditional operator, the `await` keyword needs to be explicitly placed before each asynchronous function call.wait for async calls.


AnthonyMDev (approved):

AnthonyMDev:
Thanks for the fix!

PR #1052, opened 2026-07-14
Resolve deferred response paths against the initial cache records

### LINKED ISSUE #3380: Bug: Cache writes of incremental responses uses incorrect parent key (custom cache keys)

joseph-francis-ih:
Fixes apollographql/apollo-ios#3380

## Problem

When a `@defer` query streams its response, each incremental chunk carries a GraphQL `path` (e.g. `["animal", "friend"]`) telling us where the data belongs. We turn that path into a normalized cache key so the deferred fields merge onto the record the initial chunk already wrote.

Today we build that key by naively joining the root key with the raw path components (`QUERY_ROOT.animal.friend`). But normalized records are keyed by their cache IDs, so the real chain is `QUERY_ROOT -> Animal:<id> -> ...`. The naive key matches nothing, so the deferred fields land on phantom records that nothing references. The real records never receive the deferred data, and a cache-first read comes back missing those fields.

## Fix

Instead of the naive join, walk the initial response's already-normalized `RecordSet` along the deferred `path` to find the object's real cache key, at parse time before the records reach the cache.

- `IncrementalResponseExecutionHandler` now takes the initial records and passes them to `rootCacheReference(for:path:resolvingAgainst:)`.
- `rootCacheReference` walks the records following references and list indices. If the walk completes it returns the real key; if any hop isn't a reference or list (deferred container was null/absent, or records weren't available), it falls back to the old join, so this is no worse than before for those edges.
- `JSONResponseParser.executeIncrementalResponses` seeds the walk with the initial chunk's records and keeps them current as each increment merges in.

Result: deferred fields merge onto the records the query already normalized, and the response is fully readable from cache.

## Why this approach

- Runs at parse time, before `cache.merge`, so records reach the cache already correctly keyed.
- Uses the initial response's own records as an authoritative path-to-key map, so there's no need to re-derive cache keys or know the schema here.
- Small, contained diff in the two response-parsing files.

## Testing

Added `test__intercept__givenIdBasedCacheKey_incrementalRecordsMergeOntoRealRecordNotNaivePath` to `JSONResponseParsingInterceptorTests_IncrementalItems`. It configures `cacheKeyInfo` so the initial chunk normalizes `animal` to an id-based key (`Animal:1`), streams a deferred increment with `path: ["animal"]`, and asserts the deferred field merges onto the real `Animal:1` record. Under the old naive join it lands on a phantom `QUERY_ROOT.animal` record, so this test fails without the fix and passes with it (verified by reverting the source change). The existing `includeCacheRecords: true` tests cover the fallback case and still pass unchanged.

### DIFF
--- a/Tests/ApolloTests/Interceptors/JSONResponseParsingInterceptorTests_IncrementalItems.swift
+++ b/Tests/ApolloTests/Interceptors/JSONResponseParsingInterceptorTests_IncrementalItems.swift
@@ -544,9 +544,6 @@ final class JSONResponseParsingInterceptorTests_IncrementalItems: XCTestCase {
 
   typealias AnimalQuery = MockDeferredAnimalQuery
 
-  /// Like ``MockDeferredAnimalQuery`` but selects `id` on `animal` so it normalizes by an id-based
-  /// cache key when `cacheKeyInfo` is configured. Used to verify deferred paths resolve to the real
-  /// record key rather than the naive `QUERY_ROOT.<path>` join.
   struct AnimalWithIDQuery: GraphQLQuery, @unchecked Sendable {
     static var operationName: String { "AnimalWithIDQuery" }
 
--- a/Tests/ApolloTests/Interceptors/JSONResponseParsingInterceptorTests_IncrementalItems.swift
+++ b/Tests/ApolloTests/Interceptors/JSONResponseParsingInterceptorTests_IncrementalItems.swift
@@ -540,6 +540,325 @@ final class JSONResponseParsingInterceptorTests_IncrementalItems: XCTestCase {
     )
   }
 
+  func test__intercept__givenAliasedPathComponent_incrementalRecordsResolveToRealRecord()
+    async throws
+  {
+    // given
+    // The container is reached via an aliased field (`pet: animal`), so the incremental path is
+    // ["pet"] but the record stores the field under its name `animal`. The resolver must map the
+    // response key back to the field's cache key.
+    await MockSchemaMetadata.stub_cacheKeyInfoForType_Object(IDCacheKeyProvider.resolver)
+
+    let operation = AliasedAnimalQuery()
+    let streamMocker = AsyncStreamMocker<Data>()
+
+    let urlResponse = HTTPURLResponse.deferResponseMock()
+
+    // when
+    var actualResultsIterator = try await subject.parse(
+      response: HTTPResponse(
+        response: urlResponse,
+        chunks: streamMocker.getStream()
+      ),
+      for: JSONRequest.mock(operation: operation, fetchBehavior: .NetworkOnly),
+      includeCacheRecords: true
+    )
+    .getStream()
+    .makeAsyncIterator()
+
+    streamMocker.emit(
+      """
+      {
+        "data": {
+          "pet": {
+            "__typename": "Animal",
+            "id": "1",
+            "species": "Canis Familiaris"
+          }
+        },
+        "hasNext": true
+      }
+      """.data(using: .utf8)!
+    )
+
+    // then
+    let result1 = try await actualResultsIterator.next()
+    expect(result1?.cacheRecords).to(
+      equal(
+        RecordSet(records: [
+          Record(key: "QUERY_ROOT", ["animal": CacheReference("Animal:1")]),
+          Record(key: "Animal:1", ["__typename": "Animal", "id": "1", "species": "Canis Familiaris"]),
+        ])
+      )
+    )
+
+    streamMocker.emit(
+      """
+      {
+        "incremental": [
+          {
+            "label": "deferredGenus",
+            "data": { "genus": "Canis" },
+            "path": ["pet"]
+          }
+        ],
+        "hasNext": false
+      }
+      """.data(using: .utf8)!
+    )
+
+    // then
+    let result2 = try await actualResultsIterator.next()
+    expect(result2?.cacheRecords).to(
+      equal(
+        RecordSet(records: [
+          Record(key: "QUERY_ROOT", ["animal": CacheReference("Animal:1")]),
+          Record(
+            key: "Animal:1",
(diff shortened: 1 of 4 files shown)


joseph-francis-ih:
Hi folks, this PR is ready for review. Details are in the PR description. It'd be awesome if you guys can take a look and possibly allow the rest of the CI checks.

cc: @AnthonyMDev / @calvincestari / @bignimbus / @bobafetters

AnthonyMDev:
I'll take a look at this tomorrow. Thank you!

joseph-francis-ih:
> I'll take a look at this tomorrow. Thank you!

@AnthonyMDev , wanted to check in to see if you got a chance to review. 🙏

AnthonyMDev:
Thanks so much for this @joseph-francis-ih, and for the excellent write-up in the description — the problem statement is exactly right, and the "resolve against the initial records at parse time" approach is a good one. I especially like that the walk can only ever resolve to a real record or fall back to the old join, so it's a strict improvement with no risk of merging deferred data onto the *wrong* record.

I dug into it and I think it's very close. There are a few edge cases I'd want covered before we merge, mostly because the `path` in an incremental payload doesn't line up with our cache keys in every case.

The core thing to know: the incremental `path` is built from GraphQL **response keys**, but records store their fields under `cacheKeyForField()` (see `InputValue+Evaluation.swift`), which is the field *name* — or `name(args)` when there are arguments — never the alias. So `resolveCacheKey` walking `record[pathComponent]` only matches when the response key equals the field name for every hop. Two cases fall out of that:

1. **Aliased fields on the path.** For `pet: animal { ...defer }` the path is `["pet"]`, but the record field is `animal`, so `record["pet"]` misses and we fall back to the naive join. No regression, but the fix silently doesn't help, and aliases are common enough that I'd want this handled or at least explicitly documented.

2. **Fields with arguments on the path.** Same root cause — the record stores `friends(first: 10)` but the path carries `friends`. That means the paginated-list defer case (`animal { friends(first: 10) @defer {...} }`) won't resolve either. This one matters because paginated `@defer` is a real use case.

There's also a smaller one: the list branch casts the stored list with `as? [JSONValue]`, but we store lists as `[JSONValue?]`. That cast happens to work when every element is non-nil, but it fails wholesale if any sibling element is missing, so a valid target index inside a partially-null list falls back. Worth a guard so it degrades per-element rather than all-or-nothing.

On testing — the new test is great for what it covers, but it only exercises the single reference hop (`path: ["animal"]`), and the whole point of the change is walking references *and* list indices across multiple hops. I'd want to add:

- A **list-index test** — a deferred fragment on a list element (`path: ["animal", "friends", 0]`) asserting the increment lands on the real `Friend:<id>` record. This is the highest-value one since the list branch is completely untested today and relies on that non-obvious array cast.
- A **multi-hop / sequential-increment test** where a second increment resolves against a container that a *first* increment introduced, so we actually cover the "keep the records current as each increment merges in" logic.
- Optionally a test that pins the fallback for an aliased or argument-bearing path component, so cases 1 and 2 are captured as intended behavior rather than an accidental gap.

Would you like to take these on yourself, or would you prefer we pick it up from here? Either is totally fine — you've done the hard part already, and I'm happy to push it over the line if you'd rather hand it off. If you want to keep going and have any questions about the alias/argument key handling, post them here and I'll help however I can. Thanks again for the contribution!

joseph-francis-ih:
Thanks @AnthonyMDev for the detailed review, these are great points. I went ahead and took a stab at this. Pushed [changes](https://github.com/apollographql/apollo-ios-dev/pull/1052/changes/edacd521c8bf274376f437ed8e53849a74752f1f) covering all three cases:

Aliased and argument-bearing hops. `resolveCacheKey` now walks the operation's selection set alongside the records, so each field hop maps its response key back to the field's real cache key via `Selection.Field.cacheKey(with:)`. That handles aliases `(pet: animal)` and arguments `(friends(first: 10))`, so paginated `@defer` resolves now. The lookup descends through fragments, inline fragments, deferred fragments, and conditional groups. If a response key matches fields with different cache keys across type branches I treat it as ambiguous and fall back, so it still can only ever resolve to a real record or defer to the naive join.

List cast. Changed to `[JSONValue?]` to match what the normalizer stores.

Tests. Added cases for an aliased hop, an argument-bearing hop, a list-index hop resolving to an `id`-based element record (with a `null` sibling present), and a multi-hop path where a later increment resolves against a container an earlier increment introduced. The aliased and argument tests fail on the previous code and pass now. The list-index and multi-hop tests are new coverage of the index branch and the "keep records current" logic, which were untested before.

For context, this is a priority for our team since the mis-keying is hitting our app not landing in the cache, means we lose cache-first reads on a hot screen, so we are keen to get this out. Feel free to push changes or add/remove commits directly to this PR if that is faster for you/your team, and I am equally happy to keep iterating on it myself. I'm obviously not as well-versed on the SDK as you guys are.

Let me know what you think! ❤️

AnthonyMDev:
Pushed a small follow-up commit on top of your changes (fe985a91f).

The alias, argument, and list-index coverage all looks great now. The one behavior I changed is what happens when cache records are available but the walk *can't* resolve the path: instead of falling back to the naive `QUERY_ROOT.<path>` join, it now throws `IncrementalResponseError.unresolvablePath`.

The reasoning: once we have the records in hand, the only paths the naive join keys correctly are exactly the ones the walk also resolves — so a walk failure with records present is never a case where the fallback happens to be right. It would just silently write the deferred fields onto a phantom record, which is the original bug this resolution exists to fix. For a well-formed `@defer` response the container is always present in the initial chunk, so a failure here means either a malformed response or a selection shape we don't handle yet, and I'd rather surface that than corrupt the cache quietly. The `includeCacheRecords == false` path keeps the naive join, since the root key never reaches the cache there anyway.

I added two tests alongside it — one asserting the throw when records are present and the path is unresolvable, and one pinning the no-throw behavior when records are omitted. Everything's green locally. Thanks again for sticking with this one; it's a solid fix.

PR #1063, opened 2026-07-23
Apply `additionalCapitalizationRules` to generated operation & fragment type names

3redrubies:
## Summary

Follow-on to #1037 which introduced the opt-in `additionalCapitalizationRules` change which applied the rules to property-level names only — field accessors, initializer parameters, enum cases, input-object fields, and test-mock fields. 

This PR extends the same opt-in rules to the generated Swift type names of GraphQL operations and named fragments (the `struct` names).

## Changes

- Adds capitalizer-aware overloads `generatedDefinitionName(capitalizer:)` (operations, fragments, local cache mutations)
- Threads the existing `Capitalizer` alongside `Pluralizer` so every reference to an operation/fragment type name stays consistent
- Updates the `additionalCapitalizationRules` doc comment.
- Only generated Swift identifiers change & fully opt-in so `Capitalizer.apply(to:)` returns its input unchanged when no rules are configured

## Example

```swift
// additionalCapitalizationRules = [ { "term": { "string": "id" }, "strategy": "upper" } ]

// before
public struct SomethingByIdSubscription: GraphQLSubscription {
  public static let operationName: String = "SomethingById"   // unchanged
}

// after
public struct SomethingByIDSubscription: GraphQLSubscription {
  public static let operationName: String = "SomethingById"   // still unchanged
}
```

## File-name re-casing & pruning

This PR re-cases generated file names to match the type names. The pre-existing pruner behavior that made case-only renames unsafe on case-insensitive volumes (default macOS) is fixed here as well:

- `deleteExtraneousGeneratedFiles` skips old paths that differ from a written file's path only by case unless the volume is detected as case-sensitive (`URLResourceKey.volumeSupportsCaseSensitiveNamesKey`, memoized per parent directory; unknown ⇒ skip, so detection failure can never delete fresh output).
- `ApolloFileManager.createFile` renames an existing file whose on-disk name differs from the requested path only by case before writing, so the directory entry adopts the new casing and case-insensitive/case-sensitive checkouts generate identical file names.
- Fragment file names now derive from the same normalization helper as the fragment type names (`asNormalizedFragmentName(capitalizer:)`), mirroring operations — previously rules ran on the raw name for files but the `firstUppercased` name for types, so `.upper` rules and case-sensitive regex terms could make them drift. The reserved-name `_Fragment` suffix never appears in file names, matching schema types whose file names also omit their reserved-name suffixes.

Covered by end-to-end regression tests (skipped on case-sensitive volumes) and deterministic unit tests for both volume modes.

### DIFF
--- a/Tests/ApolloCodegenInternalTestHelpers/MockFileManager.swift
+++ b/Tests/ApolloCodegenInternalTestHelpers/MockFileManager.swift
@@ -10,6 +10,7 @@ public class MockApolloFileManager: ApolloFileManager {
     case removeItem(_ handler: @Sendable (String) throws -> Void)
     case createFile(_ handler: @Sendable (String, Data?, FileAttributes?) -> Bool)
     case createDirectory(_ handler: @Sendable (String, Bool, FileAttributes?) throws -> Void)
+    case moveItem(_ handler: @Sendable (String, String) throws -> Void)
 
     // These are based on the return string from the #function macro. They are used in overriden
     // functions to lookup the provided closure. Be aware that if the function signature changes
@@ -20,6 +21,7 @@ public class MockApolloFileManager: ApolloFileManager {
       case .removeItem(_): return "removeItem(atPath:)"
       case .createFile(_): return "createFile(atPath:contents:attributes:)"
       case .createDirectory(_): return "createDirectory(atPath:withIntermediateDirectories:attributes:)"
+      case .moveItem(_): return "moveItem(atPath:toPath:)"
       }
     }
   }
@@ -58,6 +60,23 @@ public class MockApolloFileManager: ApolloFileManager {
   /// - Parameter closure: The mocked function closure.
   public func mock(closure: Closure) {
     _base.mock(closure: closure)
+  }
+
+  // MARK: Case sensitivity mocking
+
+  /// Mocks the case sensitivity of the volume reported for all paths.
+  ///
+  /// When not mocked, the volume's case sensitivity for the mock's fictitious paths cannot be
+  /// determined, so file name case adoption performs no additional file system calls unless a
+  /// test opts in.
+  public func mock(volumeIsCaseSensitive: Bool) {
+    volumeCaseSensitivityProvider = { _ in volumeIsCaseSensitive }
+  }
+
+  /// Mocks the on-disk casing reported for a path.
+  public func mock(onDiskCasedPath: String, forPath path: String) {
+    let fallback = onDiskCasedPathProvider
+    onDiskCasedPathProvider = { $0 == path ? onDiskCasedPath : fallback($0) }
   }
 
   private func didCall(closure: Closure) {
@@ -209,6 +228,27 @@ public class MockApolloFileManager: ApolloFileManager {
       }
 
       try handler(path, createIntermediates, attributes)
+    }
+
+    public override func moveItem(atPath srcPath: String, toPath dstPath: String) throws {
+      let key = #function
+
+      guard
+        let closure = closures[key],
+        case let .moveItem(handler) = closure else {
+        if strict {
+          XCTFail(missingClosureMessage(key))
+          return
+        } else {
+          return try super.moveItem(atPath: srcPath, toPath: dstPath)
+        }
+      }
+
+      defer {
+        didCall(closure: closure)
+      }
+
+      try handler(srcPath, dstPath)
     }
   }
 }
--- a/Tests/ApolloCodegenTests/ApolloCodegenTests.swift
+++ b/Tests/ApolloCodegenTests/ApolloCodegenTests.swift
@@ -2096,6 +2096,230 @@ class ApolloCodegenTests: XCTestCase {
     await expect { await ApolloFileManager.default.doesFileExist(atPath: testInTestMocksFolderUserFile) }.to(beTrue())
   }
 
+  func test__fileDeletion__givenCapitalizationRuleChangesGeneratedFileNameCase_onCaseInsensitiveVolume_shouldNotDeleteRegeneratedFile() async throws {
+    // given
+    let isCaseSensitive = await ApolloFileManager.default.volumeSupportsCaseSensitiveNames(
+      forPath: directoryURL.path
+    )
+    try XCTSkipIf(isCaseSensitive != false, "This test requires a case-insensitive volume.")
+
+    try await createFile(containing: schemaData, named: "schema.graphqls")
+
+    try await createOperationFile(
+      type: .query,
+      named: "TestOperationById",
+      filename: "TestOperationById.graphql"
+    )
+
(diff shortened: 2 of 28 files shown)


AnthonyMDev:
Thanks for this @3redrubies, and thanks especially for the really thorough writeup of the file-name pruning problem — that saved me a lot of digging.

On the file names: I agree with your diagnosis, and I think this is the thing we most need to get right before merging. It's technically pre-existing pruner behavior, but this change is what makes it likely to actually bite people — turning on a rule re-cases a bunch of file names at once, and on a case-insensitive volume (default macOS) `deleteExtraneousGeneratedFiles` then deletes the very files we just wrote, because `oldGeneratedFilePaths.subtracting(writtenFiles)` is a case-sensitive `Set<String>` comparison. Since `pruneGeneratedFiles` defaults to `true`, that's the default experience, not an edge case. So regardless of how you want to structure things, the pruner needs to become case-insensitivity-aware — detect the filesystem's case sensitivity and treat an old path that case-insensitively matches a written path as "written" so it's excluded from the delete set (and/or have the writer do an explicit `renameFile` when only the case differs, which it already supports).

I'm happy either way on structure: you can split the pruner fix into its own PR and keep this one to the in-code type-name changes, or keep it all together. Whatever's easier for you — the only hard requirement is that we don't ship the file-name re-casing without the pruner fix landing alongside it.

The other thing I want to nail down is the type-name casing. Type names do always need to start with a capital letter, and right now that isn't guaranteed. Property names apply the capitalizer and *then* `firstLowercased`, but the type-name path does `firstUppercased` *before* the capitalizer and nothing after — so `.lower` or `.replace` can undo the leading capital. A `.lower` rule on `id` turns a fragment `IDDetails` into `struct idDetails`, and `.replace("graphQL")` on `GraphqlConfig` gives you `struct graphQLConfig`. It compiles, but a lowercase-leading type is exactly the kind of inconsistency we don't want to hand people. The fix is to apply `firstUppercased` *after* the capitalizer for type names — it's a no-op for `.upper` and corrects the other two.

That gap is currently invisible because the type-name tests only cover `.upper`. Let's get extensive unit tests in for this — `.lower` and `.replace` hitting the leading segment of operation, fragment, and local-cache-mutation type names, plus references to them — so we can be confident every rule strategy still produces a valid, capitalized type name.

If you dig into the pruner side and have any questions, post them here — happy to talk through the case-sensitivity detection if it's fiddly. Thanks again for pushing this forward!

3redrubies:
Thanks for taking a look @AnthonyMDev; will keep it in one PR so its a bit easier.

**Type-name casing** (46102171a)
`firstUppercased` now runs *after* the capitalizer for operation, fragment, and LCM type names, so your examples come out as `IdDetails` / `GraphQLConfig`. While reordering I hit two more bugs in the fragment path and fixed them too: the `_Fragment` escape checked the name *before* the rules ran, so a rule could turn a fragment into a reserved name and slip past it, and `generatedDefinitionName(capitalizer:)` had no escape at all. Declarations and references now go through the same `asFragmentName(capitalizer:)`. With no rules set, the output is unchanged. Added `.lower` and `.replace` tests for declarations, references, file names, and the `Capitalizer` itself.

**Pruner** (f7d4f4a21)
Did both things you suggested, based on whether the volume is case-sensitive (`volumeSupportsCaseSensitiveNamesKey`). The pruner now leaves an old path alone if it only differs by case, unless we know for sure the volume is case-sensitive. If we can't tell, we skip it — a leftover file is harmless, deleting a file we just wrote isn't. And `createFile` renames the existing file first, so case-sensitive and case-insensitive checkouts end up with the same names. The check is injectable, so there are tests for both kinds of volume plus end-to-end ones.

Two things I left alone on purpose, both covered by tests: fragment *file names* and fragment *accessor names* still come from the raw name (`var iDDetails: IdDetails`), same as main.

AnthonyMDev:
This is looking really good — thanks @3redrubies. I pulled it down and ran through it: the type-name fix and the pruner/writer case-insensitivity work are both solid, and the test coverage is genuinely thorough. I especially like that declarations and references now share a single code path so the `_Fragment` escaping stays consistent when a rule re-cases a name into a reserved type name, and that the filesystem case detection is injectable so the pruner behavior is actually unit-testable in both volume modes. Nice work.

One thing I want to talk through before we merge: the fragment file names. Right now a fragment `IDDetails` with a `.lower` rule on `id` generates the file `idDetails.graphql.swift` containing `struct IdDetails` — the type name gets the leading-capital normalization but the file name doesn't, so they don't match and the file ends up lowercase-leading. The operation side doesn't have this — operation file names go through `generatedDefinitionName(capitalizer:)`, so the file name always matches the type. It's really just the fragments that are asymmetric here.

I can see the argument for keeping the `_Fragment` reserved-word suffix out of file names, since that's a Swift keyword-escaping concern and file names have never carried it. But that's separate from the leading-capital normalization, and I don't think we want to ship a file whose name doesn't start with a capital and doesn't match the `struct` inside it — that's the same kind of oddity we were trying to avoid on the type names, just moved to the file name.

Is there a reason for leaving the fragment file names out of the normalization that we're missing? If there's a case you ran into that made this the right call, I'd like to understand it. Otherwise I'd lean toward having the fragment file name match the type name (deriving both from the same helper so they can't drift), and treating "should file names keep the `_Fragment` suffix" as its own explicit decision rather than a side effect.

Everything else looks merge-ready to me once CI is green. Thanks again for sticking with this one!

3redrubies:
Thanks @AnthonyMDev — fixed in 0d0bafae7. Fragment file names and type names now derive from the same normalization helper. To your question, there was no reason worth keeping: file names already get `firstUppercased` on main, so there was never a rename risk; the only real drift was `.upper` rules and case-sensitive regexes seeing raw casing on the file path but normalized casing on the type path.

For the `_Fragment`, made it explicit so file names do not carry it. That matches the existing schema-type behavior. `render(as: .filename)` never applies the `TypeNamesToSuffix` escape, so an object named `Type` is already `struct Type_Object` inside `Type.graphql.swift`. It keeps keyword-escaping a pure type-name concern.

Updated the three existing file-name pins, added the `.upper` leading-segment divergence case, the reserved-name file-name pin, `generatedFileName` unit coverage, and a fragment flavor of the e2e case-rename/pruner regression test.

AnthonyMDev (approved):
This is great — thanks for working through all the iterations on this @3redrubies. Deriving the fragment file name and type name from a single `asNormalizedFragmentName(capitalizer:)` is exactly what I was hoping for, and keeping `_Fragment` out of file names to match how schema type file names already drop their reserved suffixes is the right call — nice touch pinning it with a test. I pulled it down and everything builds and passes locally, including the lowercase-fragment-plus-upper-rule case that used to drift.

Everything I raised is addressed and CI is green. Approving — thanks again for the thorough work and for sticking with it!

AnthonyMDev (approved):

PR #1066, opened 2026-08-04
Restore requireNonOptionalMockFields with current behavior by default

karim-alweheshy:
## Summary

- Restores `requireNonOptionalMockFields` on `ApolloCodegenConfiguration.OutputOptions`.
- Defaults the option to `true`, preserving the current non-optional, schema-defaulted test-mock initializer behavior.
- Allows users to set the option to `false` to generate optional initializer parameters that default to `nil`.
- Preserves nullable list-element types in both modes.
- Documents both behaviors and adds focused rendering and Codable coverage.

## Context

[#669](https://github.com/apollographql/apollo-ios-dev/pull/669) deliberately made non-optional test mock fields unconditional during review, but its v1.23 changelog entry continued to describe a configuration flag.

The unconditional generated signatures can be inconvenient for tests that pass `nil`, forward optional scalar, enum, or object values, or intentionally construct partial response data. This PR restores the flag without changing the current default behavior.

## Compatibility

- Existing configuration files without the key decode to `true`, preserving current generated output.
- With the default `true`, non-null schema fields use non-optional parameters with schema-derived defaults.
- With `false`, every convenience initializer parameter has an optional outer type and defaults to `nil`.
- Nullable list elements remain nullable, for example `[String?]?`, in both modes.
- Encoding omits the key at its default value and serializes `false` for the opt-out.

## Validation

- `swift build --package-path apollo-ios-codegen --target ApolloCodegenLib`
- Full `ApolloCodegenTests` target on macOS: 1,574 tests passed, 0 failed, 0 skipped


karim-alweheshy:
@apollographql/client-swift Could someone please review this test-mock codegen option fix? The focused generator tests, default/opt-in end-to-end codegen smoke tests, PR metadata, AI style review, and CLA checks are green.

AnthonyMDev:
Thanks for this @karim-alweheshy. One correction first: the flag existed early in #669, but @calvincestari and I deliberately dropped it in review to make the behavior unconditional — the v1.23 changelog line just never got updated after that. So the changelog is the bug, not the code.

That said, the compatibility problems you're describing (explicit `nil`, forwarding optionals, partial mocks) are real, so I do want the flag back. Can you flip it around: keep `requireNonOptionalMockFields`, but default it to `true` instead of `false`? `true` preserves current behavior, `false` becomes the opt-out. That means updating `Default.requireNonOptionalMockFields`, the docs/changelog, and the tests that currently assert `false` as the baseline.

Separately, unrelated to this PR: `DefaultMockValueProviding` can still reference a pruned mock type when `reduceGeneratedSchemaTypes` is on — worth a follow-up, doesn't need to block this.

karim-alweheshy:
Updated this to follow your requested direction:

- `requireNonOptionalMockFields` now defaults to `true`, preserving the current generated output.
- Setting it to `false` is the opt-out for optional parameters that default to `nil`.
- Codable coverage verifies a missing key defaults to `true`, the default is omitted when encoding, and explicit `false` round-trips.
- Documentation, changelog, and renderer baselines now describe the default-true behavior.
- I removed the pruned-mock regression from this PR and will keep that issue separate.

Validation: the full `ApolloCodegenTests` target passes with 1,574 tests, and `ApolloCodegenLib` builds successfully.

PR #1067, opened 2026-08-05
Improve SQLiteDatabase extensibility by exposing DatabaseRow properties

ErShubhShankar:
This PR brings the same extensibility improvement that was previously introduced in(#1056 ) Apollo iOS v1 to the v2 in  implementation.
In v2, DatabaseRow is exposed through the public SQLiteDatabase protocol but its stored properties are not publicly accessible, which prevents wrapper/decorator implementations (e.g. encryption, compression, logging) from reusing ApolloSQLiteDatabase through composition.
This change makes the required DatabaseRow members public, allowing custom SQLiteDatabase implementations to inspect and transform row contents while continuing to leverage Apollo's existing SQLite implementation.

Reference: https://github.com/apollographql/apollo-ios-dev/pull/1056

### DIFF
--- a/apollo-ios/Sources/ApolloSQLite/SQLiteDatabase.swift
+++ b/apollo-ios/Sources/ApolloSQLite/SQLiteDatabase.swift
@@ -2,8 +2,8 @@ import Foundation
 import Apollo
 
 public struct DatabaseRow {
-  let cacheKey: CacheKey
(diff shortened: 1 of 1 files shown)


AnthonyMDev:
Sorry I didn't get to this earlier! Merging now. I'll get a patch version out with this soon!

PR #1069, opened 2026-08-19
Fix flaky cancellation test: replace polling with async wait

calvincestari:
## Summary

Fixes `test__cancellingTask__propogatesTaskCancellationToInterceptors` in `RequestChainNetworkTransportTests`, which was flaky in overnight CI runs.

- **Root cause 1 — polling race**: `toEventually(beTrue())` polls every ~10ms for up to 1 second. Cancellation propagates across 4–5 cooperative-scheduler hops (`task.cancel()` → outer task wakes → stream `onTermination` → inner `executingInAsyncTask` task cancelled → `Task.checkCancellation()` throws → flag written). Under CI load this chain can exceed the timeout.
- **Root cause 2 — data race**: `nonisolated(unsafe) var hasBeenCancelled` bypasses Swift's concurrency checker but provides no synchronization — the write from the inner task and the read from Nimble's polling task are a real data race.
- **Root cause 3 — unguarded cancel**: `task.cancel()` was called with no preceding `await`, so the outer task might not have started before cancellation was requested, making the propagation path non-deterministic.

## Changes

**`CancellationTestingInterceptor`**
- Replace `nonisolated(unsafe) var` with `NSLock`-protected state
- Add `waitForCancellation() async` that suspends via `CheckedContinuation` and resumes the instant cancellation is detected (no polling, no timeout)

**`RequestChainNetworkTransportTests`**
- Add `await Task.yield()` before `task.cancel()` to ensure the inner stream task has started
- Replace `toEventually(beTrue())` with `await waitForCancellation()` + synchronous `to(beTrue())`

## Test plan

- [x] Run `test__cancellingTask__propogatesTaskCancellationToInterceptors` locally to confirm it passes
- [ ] Confirm no regressions in `ApolloTests` suite
- [ ] Watch overnight CI for absence of the flaky failure

### DIFF
--- a/Tests/ApolloTests/CancellationHandlingInterceptor.swift
+++ b/Tests/ApolloTests/CancellationHandlingInterceptor.swift
@@ -2,8 +2,14 @@ import Apollo
 import ApolloAPI
 import Foundation
 
-final class CancellationTestingInterceptor: GraphQLInterceptor {
-  private(set) nonisolated(unsafe) var hasBeenCancelled = false
+final class CancellationTestingInterceptor: GraphQLInterceptor, @unchecked Sendable {
+  private let lock = NSLock()
+  private var _hasBeenCancelled = false
+  private var waiters: [CheckedContinuation<Void, Never>] = []
+
+  var hasBeenCancelled: Bool {
+    lock.withLock { _hasBeenCancelled }
+  }
 
   func intercept<Request: GraphQLRequest>(
     request: Request,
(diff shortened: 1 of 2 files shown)


calvincestari:
@AnthonyMDev here's a Claude-suggested change for the consistently flaky cancellation test. Let's see if the change resolves the issue for this single test in the next week or two of overnight CI builds and if it does then I'll generalize the fix as there are a few other tests that have the same pattern (although they're not flaky as yet).

calvincestari:
Merging this PR and will keep an eye on the flakiness of request chain cancellation tests.

PR #1074, opened 2026-09-08
Improve handling of errors when GraphQLInterceptors do not call next

### LINKED ISSUE #3654: RequestChain force-unwraps finalRequest when a GraphQLInterceptor does not call next
### Summary

`RequestChain` documents that a `GraphQLInterceptor` must call `next`, but enforces it with a force-unwrap rather than a diagnostic. An interceptor that returns its own result stream without calling `next` crashes inside Apollo with a nil-unwrap, giving no indication of which interceptor is at fault.

### Detail

In `Sources/Apollo/RequestChain/RequestChain.swift` at `2.3.0`:

```swift
185    nonisolated(unsafe) var finalRequest: Request!
186    var next: @Sendable (Request) async -> InterceptorResultStream<Request> = { request in
188      finalRequest = request
190      return execute(request: request)
191    }
...
216      try await writeToCacheIfNecessary(response: response, for: finalRequest)
```

`finalRequest` is assigned only inside the innermost `next` closure. If an interceptor returns a stream that yields a value without invoking `next`, that assignment never runs and line 216 dereferences a nil implicitly-unwrapped optional.

The two ways of violating the contract behave very differently:

- **Throwing** from `intercept` is safe. The `catch` at lines 200–205 wraps the error into a throwing stream, so it propagates out of the `for try await` at line 213 before reaching 216.
- **Returning a yielding stream** traps, because nothing checks whether `next` ran before the result is used.

So the safe-looking case and the crashing case are both "did not call `next`", which makes the failure hard to attribute.

### Reproduction

Any `GraphQLInterceptor` shaped like this, installed on a client, then `fetch` any query:

```swift
struct CannedResultInterceptor<Operation: GraphQLOperation>: GraphQLInterceptor {
  func intercept<Request: GraphQLRequest>(
    request: Request,
    next: NextInterceptorFunction<Request>
  ) async throws -> InterceptorResultStream<Request> {
    // Never calls `next`; yields a canned response instead.
    InterceptorResultStream(stream: .init(unfolding: { cannedResponse(for: request) }))
  }
}
```

Result: a nil-unwrap crash in `RequestChain.kickoffRequestInterceptors`, not an error surfaced to the caller.

This shape is a natural thing to write for test fixtures — substituting a canned result for the network is the whole point — which is how we hit it.

### Impact

Test doubles are the common case, so this mostly costs debugging time rather than production stability. The cost is that the crash names Apollo internals instead of the offending interceptor, and the workaround is non-obvious: call `next` once on a copy of the request with `fetchBehavior = .CacheOnly` (which carries `networkFetch: .never`) purely to satisfy the bookkeeping, then discard the stream.

### Suggested fix

Any of these would remove the trap:

1. Make `finalRequest` a genuine optional and skip `writeToCacheIfNecessary` when it is nil — an interceptor that never reached the transport has no request worth caching against.
2. Throw a descriptive error instead, naming the contract (e.g. an `Error` case meaning "an interceptor returned a result without calling `next`").
3. Track the offending interceptor's type when the chain is assembled so the diagnostic can name it.

Option 1 seems closest to the existing behavior, since a short-circuited chain arguably should not write to the cache at all.

### Environment

- apollo-ios `2.3.0` (exact pin)
- Swift 6, iOS

AnthonyMDev:
Fixes [#3654](https://github.com/apollographql/apollo-ios/issues/3654).

## The bug

`RequestChain` captured the request used for the post-flight cache write only inside the innermost `next` closure, and force unwrapped it when writing to the cache:

```swift
nonisolated(unsafe) var finalRequest: Request!
var next: ... = { request in
  finalRequest = request                 // only assignment
  return execute(request: request)
}
...
try await writeToCacheIfNecessary(response: response, for: finalRequest)   // crash
```

A `GraphQLInterceptor` that supplied its own results without calling `next` left the value unset and crashed the chain. Nothing in the type system or the protocol enforced calling `next`; the docs said "must", but the enforcement was a force unwrap.

Two details worth calling out:

- **The crash was unconditional.** The IUO was forced at the call site, before `writeToCacheIfNecessary`'s `guard request.writeResultsToCache` could bail out. So it crashed even for mutations, subscriptions, and `.fetchIgnoringCacheCompletely` — anything that would never touch the cache.
- **There was no test coverage** for an interceptor that skips `next` anywhere in the suite.

## The fix

Skipping `next` is treated as the contract violation it is. The chain now detects it and fails the request with a `RequestChain.Error.interceptorDidNotCallNext` that names the offending interceptor — the issue's own suggested fixes #2 and #3.

`finalRequest` becomes a genuine optional, recorded only by the innermost step as it was before. A `nil` value once a result has been emitted means the chain never reached its end:

```swift
guard let finalRequest else {
  throw Error.interceptorDidNotCallNext(
    interceptor: deepestInterceptor,
    operationName: Request.Operation.operationName
  )
}
```

The check is at the point of use rather than immediately after descent, so an interceptor that legitimately calls `next` lazily from inside its returned stream is not falsely accused. `deepestInterceptor` is diagnostics-only and records the last interceptor the request descended into, which is exactly the one that skipped `next`.

### Why not allow short-circuiting

An earlier revision of this PR made short-circuiting legal, on the theory that it is the natural shape for a test double. That was the wrong call. Skipping `next` skips *every* later step with nothing able to observe that it happened:

- `fetchBehavior` silently becomes a no-op — no cache read, no network fetch.
- Later interceptors never run, so `MaxRetryInterceptor`, `AutomaticPersistedQueryInterceptor`, auth refresh, and anything else downstream are silently disabled. Which ones depends on list order, and nothing surfaces it.
- The emitted `ParsedResult` never went through response parsing, so hand-built `cacheRecords` could be written straight into the normalized cache.

That is a bad failure to debug, and there is no use case that needs it. Every legitimate motivation has a seam intended for it:

| Want | Use |
|---|---|
| Canned response, no network | Stub `ApolloURLSession` (single requirement; `URLSession` already conforms) |
| A result only when the fetch fails | Call `next`, then `mapErrors` — its transform returns `Element?`, so it can substitute a `ParsedResult` |
| Serve results from the cache | `fetchBehavior`, or a custom `CacheInterceptor` |
| Abort the request | `throw` from `intercept` — this path was always safe |

Note that plain `map` does *not* cover the stubbing case: it only runs on results that already came back, so the fetch has to succeed first. `mapErrors` is the one that closes the gap, and it keeps the whole chain on its real code path.

The only shape that genuinely wants short-circuiting is in-flight request deduplication — returning an already-running identical request's stream. That is not supported today, and it could not be built on `NonCopyableAsyncThrowingStream` anyway, which is deliberately single-consumer.

**Compatibility:** short-circuiting has never worked. Any short-circuit that emitted a result hit the force unwrap; one that emitted none threw `ApolloClient.Error.noResults`. So no working code can depend on it, and requiring `next` removes no capability that ever functioned — it only turns a crash into a diagnosable error.

## Also in this PR

**Docs.** `RequestChain` step 1 and `GraphQLInterceptor`'s `next` parameter now state that calling `next` is required and point at the error. Step 9 gets simpler than before: the cache write is unambiguously the request as it was after every interceptor ran. The detailed guidance lives in the error's `recoverySuggestion` rather than in the protocol docs.

**Tests** — `Tests/ApolloTests/RequestChainNetworkTransportTests.swift`, three cases using a `RecordingCacheInterceptor` that captures the request each write was made with:

- skips `next` with cache records, asserting the error names the interceptor and that the fabricated records do *not* reach the cache
- skips `next` after an earlier interceptor called it, asserting the error names the second interceptor and not the first
- recovers via `mapErrors` after a failed fetch, asserting the result is emitted and the cache write still gets the fully mutated request — the sanctioned alternative, and not covered by the existing `mapErrors` tests, which only rethrow

Both violation cases assert the error case, that its associated `interceptor` is the offending one, that `errorDescription` names it, and that no cache write occurred.

**No new test-support API.** An earlier revision shipped a session mock in `ApolloTestSupport`, on the premise that the docs point users at stubbing the `ApolloURLSession` and nothing shipped let them. That premise was wrong: `ApolloURLSession` has a single requirement, `AsyncChunkSequence` is public, and `URLSession` already conforms, so a consumer can stub it in a few lines or use `URLProtocol` as any iOS test would. The dev repo's `MockURLSession`, `MockResponseProvider` and `MockURLProtocol` stay internal and unchanged.

No CHANGELOG entry, since this repo writes those at release time rather than per PR.

## Review notes

- **`RequestChain.Error` is nested in a generic type**, so `RequestChain<A>.Error` and `RequestChain<B>.Error` are distinct types and a consumer cannot write one `catch` covering all requests. `RequestChain.Retry` already has this property, so this is consistent with existing precedent. Naming it `Error` also shadows `Swift.Error` inside the `RequestChain` scope — the two `any Error` references in `RequestChain.swift` are now spelled `any Swift.Error`, matching what `ApolloClient` does with its own nested `Error`.
- **The error's `interceptor` is optional.** It is provably non-nil whenever the error is thrown (with zero interceptors the innermost step is called directly), but relying on that invariant would mean a force unwrap inside the fix for a force-unwrap crash.
- **`nonisolated(unsafe) var finalRequest` remains unsynchronized**, as before this PR. It is written only on descent, on one task, before the consuming loop reads it. The write surface is back to a single site, so this PR does not widen it. The pre-existing hazard is an interceptor that calls `next` from inside a `mapErrors` recovery, which writes during stream consumption while the loop reads; deliberately left out of scope. Note that `Synchronization.Mutex` is *not* available as a remedy — it requires iOS 18/macOS 15 and this package targets iOS 15/macOS 12 — so closing it would need an `NSLock`-boxed value or availability gating.

## Verification

Built and tested locally against Xcode's Swift 6.3 toolchain:

- `swift build` on the `apollo-ios` package — clean.
- Full `Apollo-UnitTestPlan` on the `ApolloTests` scheme, `platform=macOS` — **997 tests, 0 failures**.
- The regression coverage was verified to be meaningful: reverting only `RequestChain.swift` to `main` makes every violation case fail, while the control case passes either way.


apollo-ios/Sources/Apollo/RequestChain/Interceptors/GraphQLInterceptor.swift
```diff
@@ -22,7 +22,26 @@ public typealias InterceptorResultStream<Request: GraphQLRequest> =
 /// function called in sequential order prior to fetching the request.
 ///
 /// The interceptor may inspect or modify the provided `request`, which must then be passed into the `next` closure to
-/// continue through the ``RequestChain``
+/// continue through the ``RequestChain``.
```
AnthonyMDev, line 25: We don't need this comment. Too verbose. Remove it

apollo-ios/Sources/Apollo/RequestChain/Interceptors/GraphQLInterceptor.swift
```diff
@@ -98,11 +117,55 @@ public protocol GraphQLInterceptor: Sendable {
...
   ) async throws -> InterceptorResultStream<Request>
 
 }
+
+/// An error indicating that a ``GraphQLInterceptor`` emitted a result without calling its `next` closure, skipping
+/// the remaining steps of the ``RequestChain``.
+///
+/// See [Calling `next` Is Required](<doc:GraphQLInterceptor#Calling-next-Is-Required>) for the supported ways to
+/// supply results without performing a network fetch.
+public struct GraphQLInterceptorDidNotCallNextError: Swift.Error, LocalizedError, CustomStringConvertible {
```
AnthonyMDev: Lets change this to `RequestChain.Error` (in a new file, but namespaced inside an extension on `RequestChain`). Make it an enum and make this a case of that enum with associated values.

apollo-ios/Sources/Apollo/RequestChain/Interceptors/GraphQLInterceptor.swift
```diff
@@ -98,11 +117,55 @@ public protocol GraphQLInterceptor: Sendable {
...
+    self.interceptor = interceptor
+    self.operationName = operationName
+  }
+
+  public var description: String {
+    let interceptorName = interceptor.map { "\(type(of: $0))" } ?? "A GraphQLInterceptor"
+
+    return """
+      \(interceptorName) emitted a result for operation "\(operationName)" without calling the `next` closure. \
+      Every GraphQLInterceptor must call `next`, which is what runs the rest of the RequestChain — the remaining \
```
AnthonyMDev: This comment is way too long. It should just be a description of what went wrong, not an entire explanation of the issue. You can add a recoverySuggestion that points to documentation on this instead.

Tests/ApolloTests/RequestChainNetworkTransportTests.swift
```diff
@@ -817,4 +817,328 @@ class RequestChainNetworkTransportTests: XCTestCase, MockResponseProvider {
...
+      requestConfiguration: RequestConfiguration(writeResultsToCache: true)
+    ).getAllValues()
+
+    expect(results).to(haveCount(1))
+    // Because `next` was called, the chain was fully traversed and the cache write still gets the mutated request.
+    expect(cacheWrites.headers).to(haveCount(1))
+    expect(cacheWrites.headers.first?["X-Mutated"]).to(equal("true"))
+  }
+
+  func test__interceptorChain__givenChainRunsToCompletion__shouldWriteCacheWithFullyMutatedRequest()
```
AnthonyMDev: Is this test actually testing any new behavior that isn't already tested? Testing the new positive behavior (when next isn't called) needs to happen. But for the negative tests (next is always called), if they aren't testing any new behavior they are likely duplicates. Please audit all the tests here and ensure they aren't duplicates.

apollo-ios/Sources/Apollo/RequestChain/Interceptors/GraphQLInterceptor.swift
AnthonyMDev: Removed — the whole `## Calling `next` Is Required` section is gone (was lines 27-45). Kept the existing one-sentence contract statement above it, and retargeted the DocC link in `RequestChain.swift` step 1 at the new error case since the anchor it pointed to no longer exists. The detailed guidance now lives in `recoverySuggestion` instead.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apollo-ios/Sources/Apollo/RequestChain/Interceptors/GraphQLInterceptor.swift
AnthonyMDev: Done — now `RequestChain.Error` in a new `RequestChain+Error.swift`, an enum inside `extension RequestChain`, with `interceptorDidNotCallNext(interceptor:operationName:)` as a case.

Two consequences worth knowing: naming it `Error` shadows `Swift.Error` inside the `RequestChain` scope (`ResultStream`s `any Error` silently resolved to `RequestChain<Request>.Error` and cascaded into 8 compile errors), so the two references there are now spelled `any Swift.Error` — same as `ApolloClient` does with its own nested `Error`. And because it is nested in a generic, `RequestChain<A>.Error` and `RequestChain<B>.Error` are distinct types, so a consumer cannot write one `catch` covering all requests; `RequestChain.Retry` already behaves that way, so this is consistent with existing precedent. Happy to move it out of `RequestChain` if you would rather have a single catchable type.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

apollo-ios/Sources/Apollo/RequestChain/Interceptors/GraphQLInterceptor.swift
AnthonyMDev: Shortened — `errorDescription` is now a single sentence: `"<Interceptor> emitted a result for operation \"<name>\" without calling the `next` closure."` The explanation moved to `recoverySuggestion`, which states the requirement and points at the `GraphQLInterceptor` documentation.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

Tests/ApolloTests/RequestChainNetworkTransportTests.swift
AnthonyMDev: Audited; removed two duplicates, 6 tests down to 4.

Removed:
- `givenChainRunsToCompletion__shouldWriteCacheWithFullyMutatedRequest` (this one) — you are right, it asserts nothing new. "A full chain writes to cache" is already covered by `test__fetch__givenNetworkOnly__shouldWriteResultToCache`, and "the write receives the fully mutated request" by the `mapErrors` test below. It only existed to guard an earlier revision of this PR that seeded `finalRequest` with the initial request; now that only the innermost step records it — the pre-existing behavior — there is no new path to guard.
- `givenWriteResultsToCacheTrue` — duplicated `givenCacheRecords`. The error is now thrown before any cache consideration, so `cacheRecords: nil` vs. set exercises the identical code path.

Kept, each for a distinct reason:
- `givenCacheRecords__shouldThrowErrorAndNotWriteToCache` — the core case, and asserts fabricated records do not reach the cache.
- `givenWriteResultsToCacheFalse` — pins that the failure is not gated on cache eligibility. This is the exact shape of the original bug: the force unwrap ran before the `writeResultsToCache` guard, so it crashed even for operations that would never touch the cache.
- `givenEarlierInterceptorCalledNext__shouldNameTheInterceptorThatDidNot` — covers the `deepestInterceptor` logic; the error must name the second interceptor, not the first.
- `test__interceptorRecoveringWithMapErrors__givenFailedFetch` — not a duplicate: the existing `mapErrors` tests in `GraphQLInterceptor_ErrorHandling_Tests.swift` only rethrow or change fetch behavior, none recover by returning a `ParsedResult`. This covers the sanctioned alternative to skipping `next`, and that the cache write still happens.

If you would rather drop `givenWriteResultsToCacheFalse` too, it is the weakest of the four — strictly it now shares a code path with `givenCacheRecords`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

PR #1082, opened 2026-09-14
[triage] apollo-ios#3656: Is there a risk of fetch(fetchBehavior:) overload becoming private/internal to apollo?

AnthonyMDev:
# Triage: apollo-ios#3656

**Upstream issue:** https://github.com/apollographql/apollo-ios/issues/3656
**Category:** question  
**Confidence:** medium  
**Trigger:** new
_The factual half of the answer is fully verifiable in the code on main (the overload is public, not deprecated, not SPI-gated, and FetchBehavior is a public NetworkTransport protocol requirement), but the reporter is really asking about future API intent, which is a maintainer roadmap call I cannot verify. The phrase 'a warning from code to use callback APIs instead' is also ambiguous - I could not tell whether they saw a real compiler deprecation warning or the doc-comment recommendation._

## Summary

This concerns the apollo-ios subtree. The reporter migrated off the 1.x callback fetch APIs to ApolloClient.fetch(query:fetchBehavior:requestConfiguration:) on Apollo 2.x, because that single overload handles every cache policy and response format uniformly, and wants to know whether that overload is at risk of being made internal/private. Checking main (currently the 2.4.0 line), the overload at apollo-ios/Sources/Apollo/ApolloClient.swift:264 is plain public with no @​available(*, deprecated) and no @_spi; the only thing steering users away is a doc-comment recommendation at line 253 saying the cachePolicy overloads are preferred because they return a single value for operations known to produce one response. FetchBehavior itself is a public struct and appears in the public NetworkTransport.send(query:fetchBehavior:requestConfiguration:) and SubscriptionNetworkTransport.send(subscription:fetchBehavior:requestConfiguration:) protocol requirements, so it cannot become internal without breaking custom transports. Worth noting FetchBehavior's memberwise init is internal, so callers outside the module can only use the five predefined constants. No code change was made; a draft reply covering the current state is provided, but the forward-looking commitment needs a maintainer.

**Affected area:** `apollo-ios/Sources/Apollo/ApolloClient.swift`

## Related

- apollographql/apollo-ios#3637

## Questions for the maintainer

- Do you want the reply to include a forward-looking commitment that fetch(query:fetchBehavior:requestConfiguration:) will stay public through 2.x? I left that out because I cannot verify team intent from the code, and the draft only states the current state of the API.
- The reporter says the warning told them 'to use callback APIs instead', which reads like a typo for the cachePolicy overloads. Should we ask them to paste the exact compiler warning in case they hit a real deprecation on a different symbol?
- The doc comment on the fetchBehavior overload reads as a soft steer away from a fully supported API and is what prompted this issue. Is it worth rewording it to make clear it is a recommendation and not a deprecation?

## Draft reply (not posted)

@​ferrufino Thanks for writing this up, and for explaining how you're using the API.

`ApolloClient.fetch(query:fetchBehavior:requestConfiguration:)` is `public`, is not marked deprecated, and is not behind an `@_spi` attribute in 2.x — you can see it in [`Sources/Apollo/ApolloClient.swift`](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/ApolloClient.swift). The note in its documentation recommending the `fetch` functions that take a `cachePolicy` is guidance, not a deprecation. Those overloads exist so that operations known to return a single response hand you a single value instead of an `AsyncThrowingStream` you have to unwrap. If you'd rather have one entry point that always returns a stream, the `fetchBehavior` overload is a supported way to get that.

`FetchBehavior` is also part of the public [`NetworkTransport`](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/NetworkTransport.swift) and `SubscriptionNetworkTransport` protocol requirements — `send(query:fetchBehavior:requestConfiguration:)` — so it needs to remain public for anyone implementing a custom transport.

One thing to be aware of: [`FetchBehavior`](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/FetchBehavior.swift)'s memberwise initializer is internal, so from outside the module you can only use the five predefined values (`.CacheFirst`, `.NetworkFirst`, `.CacheAndNetwork`, `.CacheOnly`, `.NetworkOnly`).

If the warning you saw was an actual compiler deprecation rather than the documentation note, could you paste the exact text? That would tell us whether a different symbol is involved.

## Next steps

This draft PR is a tracking record opened by automated triage; it contains no file changes. Reply with a comment mentioning `@claude`, for example:

- `@claude post the draft reply on the upstream issue and close this PR`
- `@claude the reporter is on 2.3.0 and uses ApolloSQLite; implement the fix on this branch`
- `@claude this is expected behavior; reply upstream explaining why, then close this PR`

When a fix lands on this branch, Claude marks the PR ready for review.

---
_Workflow run: https://github.com/apollographql/apollo-ios-dev/actions/runs/34879039512_


AnthonyMDev:
Re-triaged by automation; description updated. Run: https://github.com/apollographql/apollo-ios-dev/actions/runs/34882031648

AnthonyMDev:
Re-triaged by automation; description updated. Run: https://github.com/apollographql/apollo-ios-dev/actions/runs/34884901240

AnthonyMDev:
Re-triaged by automation; description updated. Run: https://github.com/apollographql/apollo-ios-dev/actions/runs/34887586762

AnthonyMDev:
Re-triaged by automation; description updated. Run: https://github.com/apollographql/apollo-ios-dev/actions/runs/34891099625

AnthonyMDev:
Re-triaged by automation; description updated. Run: https://github.com/apollographql/apollo-ios-dev/actions/runs/34893658504

AnthonyMDev:
@claude post draft reply and close this PR

PR #1083, opened 2026-09-14
[triage] apollo-ios#3657: Apollo 2 watcher missing cancellation

AnthonyMDev:
# Triage: apollo-ios#3657

**Upstream issue:** https://github.com/apollographql/apollo-ios/issues/3657
**Category:** bug  
**Confidence:** medium  
**Trigger:** new
_The reporter is correct and I located the exact code, wrote failing tests, and produced a working fix that passes the full Apollo unit test plan. I dropped to medium because the fix makes two existing ApolloPagination tests fail, and making them pass would require changing their expectations, which is a maintainer decision rather than something triage should do._

## Summary

This is an apollo-ios (Apollo module) issue. The reporter is right on both counts, and the behavior is unchanged between 2.1.2 and main (GraphQLQueryWatcher.swift has not been touched since the 2.0 merge). In GraphQLQueryWatcher.swift the `cancelBlock` built in `init` only unsubscribes from the store and asynchronously sets `cancelled = true`; it never calls `lastFetch?.task.cancel()`, so the in-flight fetch task and the network work underneath it keep running after `cancel()`. Only `didReceiveResult`/`didReceiveError` check `cancelled`, so the results are dropped while the request still completes. This contradicts `cancel()`'s own doc comment ("Cancel any in progress fetching operations and unsubscribe from the store"). `fetch(fetchBehavior:requestConfiguration:)` also does not check `cancelled`, so a fetch after cancellation does start a new request. A third, related defect: `cancelBlock` returns early under `guard let client`, so if the client has been deallocated neither the unsubscribe nor the `cancelled` flag happens at all, which makes the `self.cancel()` call inside `fetch` (the "client has been deinitialized" path) a no-op. I prototyped a three-line fix (cancel and clear `lastFetch` inside the cancel task, move the unsubscribe to `client?.store.unsubscribe(...)` so cancellation is not gated on the client, and `guard !self.cancelled` at the top of `fetch`), added two tests to Tests/ApolloTests/Cache/WatchQueryTests.swift that fail before and pass after, and confirmed the whole Apollo-UnitTestPlan still passes. However, ApolloPaginationTests `GraphQLQueryPagerCoordinatorTests.test__reset__loadingState` and `test_actor_canResetMidflight` then fail: both call `pager.reset()` (which cancels the page watchers) 10ms into a 150ms mock-server delay and still wait for the server request expectation to be fulfilled, which only holds while cancellation does not reach the network. I reverted everything and left the tree clean rather than rewrite those expectations.

**Affected area:** `apollo-ios/Sources/Apollo/GraphQLQueryWatcher.swift`

## Questions for the maintainer

- Should `cancel()` propagate cancellation to the in-flight fetch task (matching its doc comment and Apollo 1), accepting that ApolloPagination's `reset()` will now actually abort mid-flight network requests and that GraphQLQueryPagerCoordinatorTests.test__reset__loadingState and test_actor_canResetMidflight need to be reworked to use a different sync point?
- Should `fetch(...)` become a no-op once the watcher is cancelled, or should calling `fetch` after `cancel` remain undefined behavior?
- `cancelled` is set from a detached Task, so there is still a window where a `fetch` that is already enqueued on the actor runs after `cancel()`. Closing that fully means making `cancelled` a nonisolated atomic, which changes the isolation of the public `cancelled` property. Is that acceptable, or should the ordering window stay?

## Draft reply (not posted)

@​ferrufino Thanks for the detailed report — you've read this correctly on both counts.

In [`GraphQLQueryWatcher`](https://github.com/apollographql/apollo-ios/blob/main/Sources/Apollo/GraphQLQueryWatcher.swift), `cancel()` unsubscribes the watcher from the store and marks it as cancelled, but it does not cancel the task stored in `lastFetch`. Because only the result-delivery paths check `cancelled`, an in-flight fetch runs to completion and its results are discarded, rather than the network work being stopped. That does not match the documented behavior of `cancel()` ("Cancel any in progress fetching operations…"), so we're treating it as a bug. You're also right that `fetch(fetchBehavior:requestConfiguration:)` does not check `cancelled`, so a fetch issued around or after cancellation can still start a new request.

There is no watcher API today that stops the underlying request — `fetch` starts its own unstructured task, so cancelling the task you call `fetch` from will not reach it either. The workaround is to own the network fetch yourself and use the watcher only for cache observation:

```swift
let fetchTask = Task {
  for try await response in try client.fetch(query: query, cachePolicy: .cacheAndNetwork) {
    // handle response
  }
}
await watcher.fetch(fetchBehavior: .CacheOnly)

// on teardown
fetchTask.cancel()
watcher.cancel()
```

Cancellation does propagate through `ApolloClient.fetch` when the task consuming the stream is cancelled, so this stops the in-flight request while the watcher still delivers cache updates. We'll follow up here when there's a fix to share.

## Next steps

This draft PR is a tracking record opened by automated triage; it contains no file changes. Reply with a comment mentioning `@claude`, for example:

- `@claude post the draft reply on the upstream issue and close this PR`
- `@claude the reporter is on 2.3.0 and uses ApolloSQLite; implement the fix on this branch`
- `@claude this is expected behavior; reply upstream explaining why, then close this PR`

When a fix lands on this branch, Claude marks the PR ready for review.

---
_Workflow run: https://github.com/apollographql/apollo-ios-dev/actions/runs/34879039512_


AnthonyMDev:
Re-triaged by automation; description updated. Run: https://github.com/apollographql/apollo-ios-dev/actions/runs/34882031648

AnthonyMDev:
Re-triaged by automation; description updated. Run: https://github.com/apollographql/apollo-ios-dev/actions/runs/34884901240

AnthonyMDev:
Re-triaged by automation; description updated. Run: https://github.com/apollographql/apollo-ios-dev/actions/runs/34887586762

